PR #225, opened 2022-09-25
:sparkles: Add streaming zarr read/write and h5ad read

Koncopd:
It is not possible to do streaming write h5ad and `fsspec`, but it should not be a big problem


Koncopd:
Ok, i have added zarr streaming read and write and h5ad streaming read. zarr is surprisingly slow, maybe something is incorrect, but don't see what immediately. I will show proper benchmarks soon.

PR #233, opened 2022-09-28
🎨 Get db metadata as a dictionary

fredericenard:
This PR will contains all the modifications needed to have a very basic version of lamindb-rest.

### DIFF
--- a/lamindb/schema/__init__.py
+++ b/lamindb/schema/__init__.py
@@ -44,7 +44,7 @@ wetlab.__doc__ = f"""Generic wetlab.
 See `lnschema-wetlab <https://lamin.ai/docs/lnschema-wetlab/api>`__.
 """
 
-from ._core import draw, get_db_metadata_as_dict, list_entities
+from ._core import draw, list_entities
 
 try:
     import lnschema_retro as retro
--- a/lamindb/schema/__init__.py
+++ b/lamindb/schema/__init__.py
@@ -44,7 +44,7 @@ wetlab.__doc__ = f"""Generic wetlab.
 See `lnschema-wetlab <https://lamin.ai/docs/lnschema-wetlab/api>`__.
 """
 
-from ._core import draw, list_entities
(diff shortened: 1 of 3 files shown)


lamindb/schema/_core.py
```diff
@@ -23,8 +22,45 @@ def draw(view=True):
...
     """Return all entities."""
-    metadata = sql.MetaData()
-    engine = settings.instance.db_engine()
-    metadata.reflect(bind=engine)
+    metadata = get_db_metadata()
     table_names = [table.name for table in metadata.sorted_tables]
     return table_names
+
+
+def get_db_metadata():
```
falexwolf, line 30: It's completely fine to add this here, and you can go ahead with merging the PR!

However, I think this can be moved to `lndb_setup` in the next refactor. ☺️ I'll be happy to do that if you agree! No need to act on this right now.

lamindb/schema/__init__.py
```diff
@@ -44,7 +44,7 @@
 See `lnschema-wetlab <https://lamin.ai/docs/lnschema-wetlab/api>`__.
 """
 
-from ._core import draw, list_entities
+from ._core import draw, get_db_metadata_as_dict, list_entities
```
falexwolf, line 47: I'd prefer to no re-export `get_db_metadata_as_dict` for the typical user. To me, that's a dev function that is either accessed through `schema._core` or through `.dev`.

falexwolf: It's important that the auto-lookup for `ln.schema.*` displays a concise list of user-centered functionality.

fredericenard: Yes totally agree

fredericenard:
@falexwolf just to be sure, is that ok to merge it although the 2 codecov checks failed ?

falexwolf:
Yes, that's OK, Fred! Don't worry too much about codecov!

PR #274, opened 2022-10-11
:sparkles: Compute checksum during ingest

fredericenard:


falexwolf:
Great! Just: can we have a test?

Koncopd:
It is important to have checksum only optional, because I am working on streaming writes directly to the cloud, in this case checksum is not directly accessible on ingestion.

fredericenard:
> It is important to have checksum only optional, because I am working on streaming writes directly to the cloud, in this case checksum is not directly accessible on ingestion.

Ok, I made it optional in dobject schema

tests/test_file.py
```diff
@@ -16,3 +17,17 @@ def test_get_name_suffix_from_filepath():
     for path, name, suffix in dataset:
         filepath = Path(path)
         assert name, suffix == get_name_suffix_from_filepath(filepath)
+
+
+def test_compute_checksum():
+    # based on https://stackoverflow.com/questions/31890341/clean-way-to-get-the-true-stem-of-a-path-object  # noqa
+    dataset = [
```
falexwolf, line 24: Is this going to work for any file type or just text files?

You mentioned that the hash value differed for h5ad back then. If we are unsure, it'd be good test a few common file formats.

falexwolf:
> It is important to have checksum only optional, because I am working on streaming writes directly to the cloud, in this case checksum is not directly accessible on ingestion.

Interesting!

tests/test_file.py
falexwolf: I'll test by re-running CI and then checking whether the h5ad was ingested with the checksum value that's displayed: 429aea8f43b6df5f7b2a8f95a1b6998b.

PR #342, opened 2022-11-10
✅ Add tests for zarr ingest and load

Koncopd:
subj

### DIFF
--- a/docs/faq/index.md
+++ b/docs/faq/index.md
@@ -12,4 +12,5 @@ linked-select
 ingest-bfx
 link-entities
 postgres
+zarr-ingest-load
 ```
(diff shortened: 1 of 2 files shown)


Koncopd:
@falexwolf 
```
/home/runner/work/lamindb/lamindb/.nox/build-3-9/lib/python3.9/site-packages/lnschema_core/_core.py:docstring of lamindb.schema.DTransform.pipeline_run_id:1:py:class reference target not found: lnschema_core.PipelineRun
nox > Command lndocs  failed with exit code 2
```
There is a problem with the docs of `lamindb.schema.DTransform`, but i can't find the reason.

falexwolf:
This is just in the docs... I did a rename `PipelineRun` -> `Run` and the docs are pointing to that most recent version. Hm.

falexwolf:
This can only happen for cross-referencing docs, which we only have for very fundamental things like the core schema module.

I'll think about a longterm solution but right now will just quickly migrate lamindb.

Koncopd:
@falexwolf thx for updating, i was on the way home, so couldn't do myself.

Issue #360, opened 2022-11-19
`use_fsspec` ignored in `ln.add` now


Koncopd:
Not sure how to fix properly, as it is only a dobject related option. However, we will replace `cloudpathlib` with `fsspec` soon anyways, so it won't be a problem anymore.

falexwolf:
This would need to be accounted for here: https://lamin.ai/docs/db/lamindb.record

Let me know if we still need it before you do the overall migration!

PR #361, opened 2022-11-20
🩹 Restore default `fsspec` for upload

Koncopd:
subj, and adds optional arguments to `ln.add` to pass `use_fsspec` if needed.
https://github.com/laminlabs/lamindb/issues/360


falexwolf:
👍

PR #372, opened 2022-11-23
🏗️ Persist the session

falexwolf:
- https://github.com/laminlabs/lndb-setup/pull/99

### DIFF
--- a/lamindb/dev/db/_core.py
+++ b/lamindb/dev/db/_core.py
@@ -1,11 +1,5 @@
-import os
-
 import sqlmodel as sqm
 from lndb_setup import settings
-
-_session = None
-if "LAMIN_SKIP_MIGRATION" not in os.environ:
-    _session = sqm.Session(settings.instance.db_engine(), expire_on_commit=False)
 
 
 def session() -> sqm.Session:
@@ -13,8 +7,4 @@ def session() -> sqm.Session:
 
     Returns a `sqlmodel.Session` object.
     """
-    if _session is not None:
-        assert _session
-        return _session
-    else:
-        return sqm.Session(settings.instance.db_engine(), expire_on_commit=False)
+    return settings.instance.session()
--- a/lamindb/dev/db/_core.py
+++ b/lamindb/dev/db/_core.py
@@ -1,5 +1,11 @@
+import os
+
 import sqlmodel as sqm
 from lndb_setup import settings
+
+_session = None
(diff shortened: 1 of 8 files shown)


lamindb/dev/db/_core.py
```diff
@@ -1,10 +1,20 @@
+import os
+
 import sqlmodel as sqm
 from lndb_setup import settings
 
+_session = None
+if "LAMIN_SKIP_MIGRATION" not in os.environ:
```
falexwolf, line 7: @fredericenard, I'm using your env for detecting the rest.

Koncopd:
Every call of `settings.instance.db_engine()` synchronizes the local sqlite file with cloud. If the session is stored, then this synchronization doesn't happen anymore.

falexwolf:
Ahhh, thanks for the hint! I moved this logic over to lndb-setup and I'll think what to do about this!

lamindb/dev/db/_core.py
falexwolf: This is now in lndb-setup!

falexwolf:
My current thought is that the main thing this means is that we need to mutual exclusion check for the sqlite file. And we might need more control for influencing synching for the user.

Koncopd:
The problem with storing the session is that the db file is synchronized only once on session initialization. Actually, this change might break a lot of things. For example, if the cached db file is deleted, it won't be synchronized anymore as far as i understand.

falexwolf:
Yes, I see. It’s an experimental change that’s deeply grounded in how users should experience lazyily loading sqlalchemy objects.

falexwolf:
I’ll walk you through it. And if there is another solution, then that’d be great.

falexwolf:
To note: In the current implementation, the old behavior (function-scoped sessions) is just an environment variable away. So, we can just switch it back anytime.

Koncopd:
What happens with the stored session when the cached db file is deleted? Or replaced? Maybe it is bettter to initiate sync of yhe db file on every session() call.

falexwolf:
We can do that, with some caveats; let’s try scoping both the session and the cache check with the `InstanceSession` lifetime (which is essentially until call of `lndb_setup.init`) for one day or two, and then talk about mutual exclusion and updates in the sqlite case in depth

Koncopd:
Mutual exclusion is not meaningful in this setting, because we don't get any changes from the cloud until new init call where new session is received. 
I mean this change seems to break the sync logic totally.

PR #387, opened 2022-11-30
✨ Check duplication before inserting records

sunnyosun:


lamindb/_delete.py
```diff
@@ -16,41 +36,58 @@ def delete(record: sqm.SQLModel):
 
     >>> # Delete metadata records
     >>> experiment = ln.select(Experiment, id=experiment_id)
-    >>> db.delete(experiment)
+    >>> ln.delete(experiment)
     >>> # Delete data objects
     >>> dobject = ln.select(DObject, id=dobject_id)
-    >>> db.delete(dobject)
+    >>> ln.delete(dobject)
+    >>> # Delete without selecting a record
```
falexwolf, line 43: How about `# Delete by field`

lamindb/_delete.py
```diff
@@ -16,41 +36,58 @@ def delete(record: sqm.SQLModel):
 
     >>> # Delete metadata records
     >>> experiment = ln.select(Experiment, id=experiment_id)
```
falexwolf, line 38: `Delete by record`

lamindb/_delete.py
sunnyosun: ok!

lamindb/dev/db/_add.py
```diff
@@ -38,22 +48,34 @@ def add(  # type: ignore  # no support of different naming of args across overlo
...
     >>> # insert a new record
-    >>> db.add(wetlab.experiment(name="My test", biometa_id=test_id))
+    >>> ln.add(wetlab.Experiment(name="My test", biometa_id=test_id))
     >>> # update an existing record
-    >>> experiment = ln.select(wetlab.experiment, id=experiment_id).one()
+    >>> experiment = ln.select(wetlab.Experiment, id=experiment_id).one()
     >>> experiment.name = "New name"
-    >>> db.add(experiment)
+    >>> ln.add(experiment)
+    >>> # check existence before inserting a new record
```
falexwolf, line 56: How about `# add a record for fields if not yet exists`?

sunnyosun: ok

lamindb/dev/db/_add.py
```diff
@@ -38,22 +48,34 @@ def add(  # type: ignore  # no support of different naming of args across overlo
     Example:
 
     >>> # insert a new record
```
falexwolf, line 50: How about, `# add a record (by passing a record)`

sunnyosun: ok!

falexwolf: I think it'd be more readable to write it in two lines, making an experiment record, and passing it.

lamindb/dev/db/_add.py
```diff
@@ -38,22 +48,34 @@ def add(  # type: ignore  # no support of different naming of args across overlo
...
-    >>> db.add(wetlab.experiment(name="My test", biometa_id=test_id))
+    >>> ln.add(wetlab.Experiment(name="My test", biometa_id=test_id))
     >>> # update an existing record
-    >>> experiment = ln.select(wetlab.experiment, id=experiment_id).one()
+    >>> experiment = ln.select(wetlab.Experiment, id=experiment_id).one()
```
falexwolf, line 53: Then this line can be skipped.

lamindb/dev/db/_add.py
sunnyosun: by fields?

lamindb/dev/db/_add.py
sunnyosun: why?

lamindb/dev/db/_add.py
```diff
@@ -38,22 +48,34 @@ def add(  # type: ignore  # no support of different naming of args across overlo
...
+        use_fsspec: Whether to use fsspec.
     """
     if isinstance(record, list):
         records = record
-    else:
+    elif isinstance(record, sqm.SQLModel):
         records = [record]
+    else:
+        model = record
+        results = select(model, **fields).all()
```
falexwolf, line 69: Could we do `.one()`, here?

falexwolf: For this reason: If think you can only "add by field" if you have one unique record in mind.

sunnyosun: ok!

falexwolf: This is why our pattern was: 
```
result = select().one_or_none()
if result is None:
    # add
```

sunnyosun: Yes, but I thought it doesn’t hurt to return all results? [](edited-with-Axolo)

falexwolf: If the latter is relevant, maybe that's great!

But the downside is that one has an unexpected return type.

sunnyosun: Yes, that’s what I did [](edited-with-Axolo)

falexwolf: So, one can't use it as an assert statement, but will always need to check that it's in fact just one record that's returned.

falexwolf: Hence, I'd rather throw an error than returning a list if there a more return results in the query.

sunnyosun: but I think the point is to check whether any records with the field exists [](edited-with-Axolo)

sunnyosun: without caring much about the uniqueness [](edited-with-Axolo)

falexwolf: With these fundamental additions to the API, I think the more minimalistic and expectable the behavior, the better.

lamindb/dev/db/_add.py
sunnyosun: ok! [](edited-with-Axolo)

lamindb/dev/db/_add.py
sunnyosun: I'm not sure; merging now, we can discuss!

PR #430, opened 2023-01-11
🚸 More encapsulated connections and ACID test 1

falexwolf:
This is the first of at least 2 PRs making improvements. Several issues are not yet fixed with this.


falexwolf:
The above build actually passed up to building the docs.

Hence, scoping transactions within add & delete is compliant with the current guide.

Let's now think about select.

PR #445, opened 2023-01-20
🚸 Introduce data validation on the ORM level

bpenteado:
Through ⬆️ `lnschema-core` to 0.25.0.

### DIFF
--- a/tests/test_db.py
+++ b/tests/test_db.py
@@ -17,7 +17,7 @@ def test_create_to_load():
 
     notebook = ln.schema.Notebook(id="83jf", v="1", name="test")
     ln.add(notebook)
-    run = ln.schema.Run(notebook_id=notebook.id, notebook_v=notebook)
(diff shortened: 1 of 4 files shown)


tests/test_db.py
```diff
@@ -17,7 +17,7 @@ def test_create_to_load():
 
     notebook = ln.schema.Notebook(id="83jf", v="1", name="test")
     ln.add(notebook)
-    run = ln.schema.Run(notebook_id=notebook.id, notebook_v=notebook)
+    run = ln.schema.Run(notebook_id=notebook.id, notebook_v=notebook.v)
```
falexwolf, line 20: This is just nice that these things are now errors!

bpenteado: Yes, that catch felt great :)

falexwolf (approved):
👌

PR #456, opened 2023-01-26
📝 Prettier section headings in the docs

sunnyosun:


sunnyosun:
@falexwolf I think this looks much better now! But I'm not sure how to get rid of the main landing page as it seems the navbar is dependent on that toctree.

falexwolf:
I agree! It looks a lot better! I made a dummy root-level page for now. We can deal with this question when integrating docs across packages!

PR #463, opened 2023-01-30
🐛 Fix fallback for notebook name

falexwolf:


Koncopd:
The can be a situation when the notebook filepath inference fails and a user wants to pass it manually, but this change will result in lamindb `header` failing anyways because `notebook_path` will fail again. I propose to import the internal variable `_filepath` from [here](https://github.com/laminlabs/nbproject/blob/f3835a94378f9bbebb0968582663d8501036d614/nbproject/_header.py#L12), it will hold an inferred or manually passed filepath after nbproject `header()` execution.

falexwolf:
Perfect, Sergei! Could you please make this change? The semantics of `_filepath` was exactly what I had been looking for!

PR #471, opened 2023-02-02
✨ Allow ingesting existing data in the cloud

sunnyosun:
Together with: https://github.com/laminlabs/lnschema-core/pull/96

### DIFF
--- a/lamindb/_record.py
+++ b/lamindb/_record.py
@@ -38,14 +38,16 @@ def serialize(
 ) -> Tuple[Any, Union[Path, CloudPath], str, str]:
     """Serialize a data object that's provided as file or in memory."""
     # Convert str to either Path or CloudPath
-    if isinstance(data, str):
+    if isinstance(data, (str, Path, CloudPath)):
         try:
-            data = CloudPath(data)
+            filepath = CloudPath(data)
+            if setup_settings.instance.storage.root not in filepath.parents:
+                raise ValueError(
+                    "Can only track objects in configured cloud storage locations."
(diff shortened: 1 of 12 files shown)


lamindb/_record.py
```diff
@@ -32,24 +34,33 @@
...
-    data: Union[Path, str, pd.DataFrame, ad.AnnData], name, format
-) -> Tuple[Any, Path, str, str]:
+    data: Union[Path, CloudPath, str, pd.DataFrame, ad.AnnData], name, format
+) -> Tuple[Any, Union[Path, CloudPath], str, str]:
     """Serialize a data object that's provided as file or in memory."""
     memory_rep = None
-    if isinstance(data, (Path, str)):
-        local_filepath = Path(data)
-        name, suffix = get_name_suffix_from_filepath(local_filepath)
+    if isinstance(data, str):
```
falexwolf, line 41: Can you add a comment: # convert str to either Path or CloudPath?

sunnyosun: ok!

lamindb/_record.py
```diff
@@ -32,24 +34,33 @@
...
+        try:
+            data = CloudPath(data)
```
falexwolf, line 43: I think this should error here if the returned cloudpath object doesn't equate the one in settings

Hence, e.g., the following code block
```
from lndb_setup import settings
if data != settings.instance.storage.root:
    raise ValueError(
        "Can only track objects in configured cloud storage locations. Please call `lndb_setup.set_storage("<bucket_name>")`.")
```

sunnyosun: Done!

PR #472, opened 2023-02-02
📝 Add notebook on ORM lazy loading behavior to FAQ

bpenteado:


falexwolf:
You could use the existing db from the bio guide! "lndb load testuser1/mydata" has the data you're interested in!

bpenteado:
> You could use the existing db from the bio guide! "lndb load testuser1/mydata" has the data you're interested in!

😆 thank you! I realized I don't fully grasp how nbproject works and I want to understand why this is not passing, so decided to take my time with this one over the weekend.

Let me know if you want this merged quickly and I'll use the db from the guide.

falexwolf:
You'll need to update the query for the featureset by something like this:
```
features = ln.select(lns.Features, id=dobject.features[0].id).one()
```
Or you retrieve the dobject within a session and then lazily load the associated features
```
with ln.Session() as ss:
    dobject = ln.select()... # query from the testuser1/mydata instance
    features = dobject.features[0]
```

falexwolf:
Oh, no hurry here! Just thought I'll add this as I noticed it yesterday but was rushing and didn't drop a comment!

PR #476, opened 2023-02-03
🚸  Make DObject upload ACID

bpenteado:


lamindb/dev/db/_add.py
```diff
@@ -81,28 +81,84 @@ def add(  # type: ignore
...
         settings.instance._cloud_sqlite_locker.lock()
         close = True
     else:
         close = False
+
+    added_records = []
+    raise_error = False
     for record in records:
-        session.add(record)
-    session.commit()
```
falexwolf, line 95: Most calls of `ln.add()` will affect ORMs that are non-DObject, and hence, to not have a corresponding upload of a (often large) file associate to them

We should retain the performance and safety benefits from committing all records in a list of records at the same time.

Say, we add 30k gene records: We'd not want to run 30k db transactions for these very light-weight operations

Hence, we need clean logic that separates ingesting a list of dobjects from a list of non-dobjects

lamindb/dev/db/_add.py
```diff
@@ -81,28 +81,84 @@ def add(  # type: ignore
...
-    for record in records:
+        # commit metadata to database
+        prepare_filekey_metadata(record)
+        try:
+            session.add(record)
+            session.commit()
+            session.refresh(record)
+        except Exception:
+            raise_error = True
```
falexwolf, line 102: Can you track the errors (or error messages), here?

So, you do something like
```
except Exception as e:
    raised_errors.append(e)
 ```

Comment (falexwolf) [reply lamindb/dev/db/_add.py:102]: It's going to very hard to understand for the user what went wrong otherwise.

File: lamindb/dev/db/_add.py [hunk 3]
```diff
@@ -81,28 +81,84 @@ def add(  # type: ignore
...
+    raise_error = False
     for record in records:
-        session.add(record)
-    session.commit()
-    for record in records:
+        # commit metadata to database
+        prepare_filekey_metadata(record)
+        try:
+            session.add(record)
+            session.commit()
```
falexwolf, line 99: This is the only thing that should be contained in the try except, I think!

lamindb/dev/db/_add.py
```diff
@@ -81,28 +81,84 @@ def add(  # type: ignore
...
+            session.refresh(record)
```
falexwolf, line 100: This is going to be called in line 117. If it triggers an exception here, it would also trigger an Exception in line 117.

Hence, it shouldn't be within a try-except.

lamindb/dev/db/_add.py
```diff
@@ -81,28 +81,84 @@ def add(  # type: ignore
...
+    added_records = []
+    raise_error = False
     for record in records:
-        session.add(record)
-    session.commit()
-    for record in records:
+        # commit metadata to database
+        prepare_filekey_metadata(record)
+        try:
+            session.add(record)
```
falexwolf, line 98: This will always go well and if it doesn't, we want a proper traceback as we'll deal with a fundamental problem - the "expected risk" that we want to mitigate by manually adding try except lies in the db transaction (say, unstable internet connection, overloaded network, etc.)

lamindb/dev/db/_add.py
falexwolf: Consider dropping the `refresh()`, here. I don't see why you need it, but I may miss something.

falexwolf:
Thanks, Bernardo!

This is all great up to the remarks above and the test case I shared on Slack.

Could you refactor and add the test case?

Please feel free to make more helper functions if sensible.

(I also see how the increased complexity here means we should move the whole logic into `lamindb.Session` to reduce some of the complexity with the session management, but that's another PR.)

lamindb/dev/db/_add.py
bpenteado: Understood. I've updated the logic and implemented single-DB-transaction for all metadata ingestion cases. In case of failure at upload, I simply create a transaction to delete the relevant metadata record.

lamindb/dev/db/_add.py
bpenteado: Agreed and implemented. Errors with `session.add()` will be raised unhandled with the full traceback. Nothing will have been commited to the DB or uploaded to storage at that point.

lamindb/dev/db/_add.py
bpenteado: Implemented. This is no longer relevant with the single DB transaction implementation.

lamindb/dev/db/_add.py
bpenteado: Implemented. See [here](https://github.com/laminlabs/lamindb/pull/476#discussion_r1097623303).

lamindb/dev/db/_add.py
```diff
@@ -81,28 +81,93 @@ def add(  # type: ignore
...
+    except Exception as e:
+        raised_errors.append(e)
+
+    # refresh records and upload data objects to storage
+    added_records = []
+    if not raised_errors:
+        for record in records:
+            session.refresh(record)
+            if isinstance(record, DObject) and hasattr(record, "_local_filepath"):
+                print("Inside LOOP!")
```
falexwolf, line 108: Remaining debugging statement :)

lamindb/dev/db/_add.py
```diff
@@ -81,28 +81,93 @@ def add(  # type: ignore
...
+                try:
+                    upload_data_object(record, use_fsspec=use_fsspec)
+                except Exception as e:
+                    # clean up metadata committed to the database
+                    print("CAUGHT ERRROR!")
```
falexwolf, line 113: Same here!

lamindb/dev/db/_add.py
```diff
@@ -81,28 +81,93 @@ def add(  # type: ignore
...
     for record in records:
+        prepare_filekey_metadata(record)
         session.add(record)
-    session.commit()
-    for record in records:
-        session.refresh(record)
+    try:
+        session.commit()
+    except Exception as e:
+        raised_errors.append(e)
```
falexwolf, line 100: I'm not entirely sure why you'd want to continue with the execution of the function in this case - let's simply discuss this in our meeting later today!

If there is an error during `session.commit()` in line 98, I'd simply close the session and exit the function rather than going into the "refresh & upload section".

bpenteado: Agreed - in fact, I initially implemented it as such!

Your previous comment on appending error messages made me think you did not want to exit the function after a first error.

falexwolf:
@Koncopd, is this OK from your side?

Koncopd:
@falexwolf I haven't checked yet, i will check soon.

falexwolf:
Nice! :D

PR #489, opened 2023-02-13
🚑 Fix tracking of added records during `ln.add()`

bpenteado:
`added_records` was not being assigned in the case of a DB error.


falexwolf (approved):
Perfect!

PR #519, opened 2023-02-24
🍱 `DFolder.get_dobject`

sunnyosun:
- Write _objectkey to .dobjects from DFolder
- Implemented `DFolder.get_dobject(relpath)`

### DIFF
--- a/lamindb/dev/db/_add.py
+++ b/lamindb/dev/db/_add.py
@@ -1,5 +1,5 @@
 from functools import partial
-from pathlib import Path
+from pathlib import Path, PurePath
 from typing import List, Tuple, Union, overload  # noqa
 
 import sqlmodel as sqm
@@ -194,7 +194,7 @@ def write_objectkey(record: sqm.SQLModel) -> None:
         root, root_str = get_storage_root_and_root_str()
 
         if isinstance(root, UPath):
-            relpath = Path(filepath.as_posix().replace(root_str, ""))
+            relpath = PurePath(filepath.as_posix().replace(root_str, ""))
         else:
             relpath = filepath.relative_to(root_str)
 
--- a/lamindb/dev/db/_add.py
+++ b/lamindb/dev/db/_add.py
@@ -203,7 +203,7 @@ def write_objectkey(record: sqm.SQLModel) -> None:
             relpath.parent / record.name if isinstance(record, DObject) else relpath
         )
 
-        set_attribute(record, "_objectkey", _objectkey.to_posix())
+        set_attribute(record, "_objectkey", _objectkey.as_posix())
 
     # _local_filepath private attribute is only added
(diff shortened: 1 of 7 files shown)


lamindb/dev/db/_add.py
```diff
@@ -176,32 +176,35 @@ def local_instance_storage_matches_local_parent(dobject: DObject):
...
+        # for DObject, _objectkey is relative path to the storage root without suffix
         _objectkey = (
-            filepath_str.replace(root_str, "").replace(record.suffix, "")
-            if isinstance(record, DObject)
-            else filepath_str.replace(root_str, "")
+            relpath.parent / record.name if isinstance(record, DObject) else relpath
         )
 
-        set_attribute(record, "_objectkey", _objectkey)
+        set_attribute(record, "_objectkey", str(_objectkey))
```
Koncopd, line 207: I think it is better to use `_objectkey.as_posix()` instead of `str(_objectkey)` when writing to the db, because on Windows and Linux\Mac `str(_objectkey)` gives different results (\\ vs /).

lamindb/dev/db/_add.py
```diff
@@ -176,32 +176,35 @@ def local_instance_storage_matches_local_parent(dobject: DObject):
...
-            root_str += "/"
-
-        if not isinstance(record, (DObject, DFolder)):
-            raise NotImplementedError
-        # for DObject, _objectkey is relative path without suffix
+    def set_objectkey(record: Union[DObject, DFolder], filepath: Union[Path, UPath]):
+        root_str = get_storage_root_str()
+
+        try:  # is local filepath
+            relpath = filepath.relative_to(root_str)
```
Koncopd, line 197: I think it is better to test for `UPath` class instead of relying on `try except` to figure out if it is cloud storage.

lamindb/dev/db/_add.py
```diff
@@ -176,32 +176,35 @@ def local_instance_storage_matches_local_parent(dobject: DObject):
...
+        except ValueError:  # is cloud filepath
+            # remove the server prefix
+            relpath = Path(filepath.as_posix().replace(root_str, ""))
```
Koncopd, line 200: I would propose to use `PurePath` here instead of `Path` for safety.

Koncopd:
Thanks you, left some minor comments.

falexwolf:
Great work, Sunny! Great comments, Sergei! I'll do the small fixes as suggested by Sergei so that this can be merged.

lamindb/dev/db/_add.py
falexwolf: Done!

lamindb/dev/db/_add.py
falexwolf: Fixed, and fixed another occurance of str(path)

PR #528, opened 2023-03-01
🚑 Fix for listing cloud dir

sunnyosun:


Koncopd:
I am not sure that this unique constraint `core.dobject.storage_id, core.dobject._objectkey, core.dobject.suffix` is needed.

Koncopd:
Copying here also:

CI fails due to a failed past CI run of the the existing-data notebook, Species.csv was ingested but not deleted at the end, just deleting the record from the db (without deleting storage) should solve the problem.

upd:
@falexwolf fixed it.

PR #530, opened 2023-03-01
✅ Towards a single API

falexwolf:
Depends on
- https://github.com/laminlabs/lndb/pull/312


Koncopd:
CI fails due to a failed CI run of the the `existing-data` notebook, Species.csv was ingested but not deleted at the end, just deleting the record from the db (without deleting storegae) should solve the problem.

PR #531, opened 2023-03-01
🔥 Remove `lamin` dependency

falexwolf:


Koncopd:
`from lamindb.setup import import lazy` in the `stream` notebook, invalid syntax.
in addition, `lazy` is not (or should not) be in `lamindb.setup`, it is in `lamindb`.

falexwolf:
Thank you!

Koncopd:
Actually i am not sure this is a good approach to the syntax errors in `nbproject`, but also i don't know how to treat them better.

falexwolf:
It's very much an edge case that's not important right now, I'd say!

I just generated them doing bulk replacement of strings.

Koncopd:
I am also confused a bit now, what is the CLI command name then?

Koncopd:
Now that is an interesting CI fail.
On the first `import lamindb` in the `lazy-loading` notebook it tries to connect to the posgres db from the previous notebooks due to `_check_migrate` in the `__init__.py` and fails.

falexwolf:
The CLI command name is `lamin`.

Koncopd:
But it is still lndb https://github.com/laminlabs/lndb/blob/1f7d4ef90111926f3cff178f40d3a72247ba0561/pyproject.toml#L60 ...

falexwolf:
> Now that is an interesting CI fail. On the first `import lamindb` in the `lazy-loading` notebook it tries to connect to the postgres db from the previous notebooks due to `_check_migrate` in the `__init__.py` and fails. This is the problem we discussed during the meeting today.

Hmmmmm. OK. I ran into this problem also when trying to blindly replace `lndb` with `lamindb.setup` in the `lndb` repository.

falexwolf:
> But it is still lndb https://github.com/laminlabs/lndb/blob/1f7d4ef90111926f3cff178f40d3a72247ba0561/pyproject.toml#L60 ...

What you show is for backward compat. The new CLI is defined within `lamindb` (see this PR).
image

falexwolf:
> Now that is an interesting CI fail. On the first `import lamindb` in the `lazy-loading` notebook it tries to connect to the postgres db from the previous notebook due to `_check_migrate` in the `__init__.py` and fails. This is the problem we discussed during the meeting today.

Unfortunately, I don't fully understand why this happens - it's a bit late, but if you think you understand what's going on here, I'd appreciate a brief Slack huddle. :)

Koncopd:
And we can't even avoid this problem by importing `setup` from `lamindb.setup` at first like before with `lndb`, because `_check_migrate` in the main `__init__.py` will be triggered anyways.

falexwolf:
It shouldn't trigger `_check_migrate` if the instance is not setup!

Koncopd:
Lets talk.

falexwolf:
Calling you on Slack!

This is what's currently running in __init__.py:
```
if (
    not _current_instance_settings_file().exists()
    or _setup_settings.instance.storage.root is None
):
    _logger.warning(
        "You haven't yet setup an instance using the CLI. Please call"
        " `lamindb.setup.init()` or `lamindb.setup.load()`."
    )
    _instance_setup = False
else:
    _instance_setup = True
    _check_migrate(usettings=_setup_settings.user, isettings=_setup_settings.instance)
```

PR #537, opened 2023-03-03
📝 Prettify guide landing page

sunnyosun:


falexwolf:
I moved the "quickstart"-like content into the README.md of the repo, as we discussed earlier today.

From there, we can replicate it on the lamin.ai/docs landing page. I don't think the guide itself should have another quick start then. Do you think so?

PR #538, opened 2023-03-05
🔥 Remove lnhub-rest dependency

fredericenard:
Would be much better if only lndb was dependent of lnhub-rest.


falexwolf:
As discussed: I think we need to keep pinning lnhub_rest here as we can't pin it in lndb because of cyclicity between lndb and lnhub_rest.

Not pinning it at all still seems dangerous to me.

PR #561, opened 2023-03-12
🏗️ Introduce `lamindb.context` and enable `ln.Run` to create contexts

falexwolf:
Instead of all the specificity for notebooks, notebooks and pipelines are now treated on an equal footing.

Along with

- https://github.com/laminlabs/lnschema-core/pull/133

Below is all gone and got replaced with `ln.Run(global_context, load_latest=True)`.

Everything is 100% backward compatible.

image

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,6 +21,7 @@ jobs:
       - name: Checkout main
         uses: actions/checkout@v3
         with:
+          submodules: recursive
           fetch-depth: 0
       - name: Checkout lndocs
         uses: actions/checkout@v3
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "lnschema-core"]
+	path = lnschema-core
+	url = https://github.com/laminlabs/lnschema-core.git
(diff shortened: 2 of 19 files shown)


sunnyosun:
Could we simplify to have the tracking centered around `Run`, rather than introducing `.context`? To me, context isn't intuitive to use.

Also, I don't feel entirely intuitive with `is_run_input`, and I always forget to set it.

Can we do the following?
```
# initiate a run from a notebook
run = ln.Run()

# track run inputs
input = ln.select(ln.DObject, name=...).one()

# initiate a run from a pipeline
run = ln.Run(pipeline_name=..., inputs=...)
# or
run.inputs.append(input)

# any new dobjects are tracked as outputs of this run
dobject = ln.DObject(data=...)
```

In a script:
```
@Run(pipeline_name=...)
def run_pipeline():
   ln.load(...) # track inputs
   ln.DObject() # track outputs
```

PS: run.inputs should also accept DFolder?

falexwolf:
```
run = ln.Run(global=True)
```

falexwolf:
Now it seems to all work with the completely run-centered design. Still I need to pass two cryptic args:
```
ln.Run(load_latest=True, global_context=True)
```

And then this things returns the global run record, even though I don't even want to have it around. 🤔

falexwolf:
And then, what if wanted to have an open session for the global run record? 🤔

PR #571, opened 2023-03-15
🐛 Actually fix instance load

sunnyosun:


Koncopd:
image
It looks like the Run is not added to the db.

PR #591, opened 2023-03-28
🚚 No longer import bionty

sunnyosun:


pyproject.toml
```diff
@@ -36,7 +36,7 @@ dev = [
     "laminci>=0.2.5",
 ]
 bionty = [
-    "lnschema_bionty==0.13.1",
+    "lnschema_bionty==0.14.0",
     "bionty>=0.11.2",
```
falexwolf, line 40: I think we could even drop the bionty dependency here entirely, right? @sunnyosun

falexwolf: Then it's even cleaner! :D

PR #605, opened 2023-03-30
Increase pinned bionty version

Zethson:
We fixed the bug where the `.current.yaml` or `.lndb.yaml` could get out of sync for entities that were recently added.

Forgot to name the branch properly from the GH web GUI offff


falexwolf:
Should one also add it here? https://github.com/laminlabs/lamindb/blob/main/lamindb/_check_versions.py

PR #614, opened 2023-04-04
🚸 Filename in `File.name`, new `File.key` and `Folder.key`, robustness overhaul

falexwolf:
- https://laminlabs.slack.com/archives/C04FPE8V01W/p1680264717331509
- https://github.com/laminlabs/lnschema-core/pull/150
- https://github.com/laminlabs/lndb/pull/349
- https://github.com/laminlabs/lndb/pull/351
- https://github.com/laminlabs/lndb/pull/348
- https://github.com/laminlabs/lndb/pull/352


Koncopd:
Does it add a suffix to a name?

falexwolf:
No, it won't add it. But it will stop stripping it.

falexwolf:
Hence, when you do
```
file = ln.File(adata, name="My cool dataset")
ln.add(file)
```
This will be queryable by `name="my cool dataset"` without a `.zarr` or a `.h5ad` suffix.

However, when you do
```
filename = "my_cool_dataset.h5ad"
file = ln.File(filename)
ln.add(file)
```
This will be queryable by `name="my_cool_dataset.h5ad"`.

Koncopd:
But then
```
file.replace("my_cool_dataset.zarr")
ln.add(file)
```
file will have `"my_cool_dataset.h5ad"` as a `name` still.

falexwolf:
> When we probably need to check this in replace, it would be strange to have a name with an old suffix after replacement with a file with a different suffix.

Yes, let us please construct unit tests for all combinations that can occur here.

In my mind, this should just issue a warning if the user tries to replace the object underlying a record while changing suffixes. But this problem should already occur independent of whether name has a suffix in it or not.

falexwolf:
Re: https://github.com/laminlabs/lamindb/pull/614#issuecomment-1495555856

Yes, this should raise a warning, and ask for "updating the name" or something similar. And I think that's legit here, because if your passing a filepath, you might actually want to query for the new `"my_cool_dataset.zarr"` rather than the old `"my_cool_dataset.h5ad"`.

If you're using auto-serialization, you'll not be interested in what's going on under the hood, and you just query for `"My cool dataset"`.

Koncopd:
But i don't understand why we need to have suffix in name? We store the file as id.suffix, not as name.

falexwolf:
> But i don't understand why we need to have suffix in name. We store a file as id.suffix, not as name. I think this just adds confusion.

See the linked Slack discussion at the top! Fred & Sunny felt it was confusing that the name column did *not* contain the suffix. Given it's literally a filename (`File.name`), I'm meanwhile also inclined to mimic `Path.name` rather than have some logic on top that strips a suffix.

So, I do think that this updated behavior is less confusing.

Finally, we meanwhile believe that default storage locations should rather lean towards semantic storage keys, rather than the primary key ones, at least when ingesting files. We fear that having object storage full of undecipherable object keys makes users shy away from LaminDB.

We'll have a thorough discussion about how to go about this, because auto-serialization should continue to write non-semantic storage keys, I'm pretty convinced.

There is a whole catalog of reasons and complexities that needs to be written down.

For this particular question though, the believe in the Slack thread was we should just mimic S3.

Koncopd:
@falexwolf 
The problem now comes from here https://github.com/laminlabs/lamindb/blob/a7af72e01f1d3c5968fb2351aff3d2b290ebd2e7/lamindb/dev/db/_add.py#L245
`_objectkey` (used for manual adding) now could contain suffix. And on `load` suffix is added again https://github.com/laminlabs/lnschema-core/blob/bdb94f5dfba0deb145fca06ff653207ded943ec1/lnschema_core/dev/_storage.py#L18

falexwolf:
Yes, am aware! Will fix this properly. I had started to work on the migrations testing problem at the same time and therefore lost focus here. This here should be straightforward and I should be able to finish it soon!

tests/test_file.py
```diff
@@ -0,0 +1,41 @@
...
+        assert suffix == "".join(filepath.suffixes)
+
+
+def test_storage_root_upath_equivalence():
+    storage_root = UPath("s3://lamindb-ci")
+    filepath = UPath("s3://lamindb-ci/test-data/Species.csv")
+    assert filepath.parents[-1] == storage_root
+
+
+def test_get_relative_path_to_root():
```
falexwolf, line 29: This one.

PR #621, opened 2023-04-09
🐛 Storage-related fixes

falexwolf:
- https://github.com/laminlabs/lamindb/issues/620

### DIFF
--- a/lamindb/_file.py
+++ b/lamindb/_file.py
@@ -155,15 +155,14 @@ def get_path_size_hash(
 
 def get_check_path_in_storage(filepath: Union[Path, UPath]) -> bool:
     assert isinstance(filepath, Path)
+    storage_root = lndb.settings.storage.root
     # the following comparisons can fail if types aren't comparable
-    if isinstance(filepath, UPath) and isinstance(lndb.settings.storage.root, UPath):
+    if isinstance(filepath, UPath) and isinstance(storage_root, UPath):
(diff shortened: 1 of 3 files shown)


lamindb/_file.py
```diff
@@ -65,8 +68,14 @@ def serialize(
...
         memory_rep = data
         suffix = infer_suffix(data, format)
-        # this is always local
-        filepath = lndb.settings.storage.cache_dir / name
+        # the following filepath is always local
+        if lndb.settings.storage.cache_dir is not None:
+            filepath = lndb.settings.storage.cache_dir / name
+        else:
+            # this should likely be added to lndb.settings.storage
+            cache_dir = Path(DIRS.user_cache_dir)
```
Koncopd, line 76: Should this be cached / written directly into the instance storage, as it is only for local instances anyways.

falexwolf: Yes, that would be reasonable. It could happen along similar lines to this: https://github.com/laminlabs/lndb-storage/blob/618a6b514c3937b5754be4596dbc47c4c74891f0/lndb_storage/object/_anndata.py#L16-L19

I just wasn't sure how the copy to storage part would be affected, because that's no longer going to be needed then.

I just wanted to move it out of the cwd for now.

falexwolf:
Tests are all passing here. There is some weird CI breakage related to artifacts or something else happening downstream of the tests.

PR #626, opened 2023-04-13
🐛 Fix track for copied / renamed notebooks

Koncopd:
subj

@falexwolf this is a quick fix. But `track` needs to be restructured completely, it becomes unmanageable.

### DIFF
--- a/lamindb/_context.py
+++ b/lamindb/_context.py
@@ -162,6 +162,8 @@ class context:
                 when automatic inference fails.
         """
         cls.instance = settings.instance
+
+        metadata = None
         # original location of this code was _nb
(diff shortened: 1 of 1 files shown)


falexwolf:
Thank you, Sergei! I agree on that a restructure is needed.

All the complexity comes from calls to nbproject, I guess. If people pass a `Transform` object via `ln.track(transform=transform)`, which they have to in case nbproject fails, the whole thing boils down to a few lines of code.

I didn't expect this to become so complicated and therefore didn't think much about structure.

Let's see whether everything works now, and then do a single final restructure in case it does, and ideally never touch it again (unless for small maintenance questions). It's so hard to test this interactive stuff.

PR #628, opened 2023-04-14
✨ Add Google Colab integration

falexwolf:


Koncopd:
I believe it should be done in `nbproject`, not here.

falexwolf:
Merging for now just to keep iterating! As mentioned on Slack, I'm happy if this code moves into nbproject! Let's discuss! It's not going to be a big refactor.

PR #629, opened 2023-04-15
🐛 Use correct _save_notebook in track

Koncopd:
@falexwolf incorrect function was used, it was saving only for jupyter lab.


Koncopd:
@falexwolf , also why i think adding save here is in general not a good solution and why i was reluctant to add this yesterday is that save here makes track significantly slower, because during save we also call `sleep(1)`.

falexwolf:
Thank you, Sergei! Let's discuss more in a meeting!

PR #653, opened 2023-04-22
🚸 Add `File.subset`

falexwolf:
- https://github.com/laminlabs/lnschema-core/pull/158


Koncopd:
We will probably need to refactor load and stage in the same vein then.

falexwolf:
Yes, we can! It's not a big deal though as there is no typing problem for `load` and `stage`. I just thought it's simplest if this is directly in lamindb, and then we don't need to worry about moving the lazy selector between lndb-storage and lnschema-core

falexwolf:
I will add a test for this in another PR! Merging for now!

PR #560, opened 2023-03-12
♻️ Move storage related code to `lndb_storage`

Koncopd:
- https://github.com/laminlabs/lndb-storage/pull/4

PR #700, opened 2023-06-01
🚚 Move lndb-storage back into lamindb

falexwolf:
The code was in the following repository for a couple of months:

- https://github.com/laminlabs/lndb-storage

It had been moved out there here:

- https://github.com/laminlabs/lamindb/pull/560


Koncopd:
The problem with `add-replace-stage.ipynb` is that cache and actual stored filepath are the same on a local instance, that means that at this cell there is already no "cache" for csv, it was deleted by `replace`.
image

Koncopd:
Yep, it was intended for testing on a remote.

PR #711, opened 2023-06-04
➖ Move nbproject to extra dependencies

Koncopd:
subj
This requires releases of https://github.com/laminlabs/lnschema-core and https://github.com/laminlabs/readfcs


falexwolf:
Thank you! This is great!

PR #724, opened 2023-06-07
🔊 `folder.tree` can only be used with existing folders in storage

sunnyosun:
when folder is not in storage

### DIFF
--- a/lamindb/_folder.py
+++ b/lamindb/_folder.py
@@ -106,10 +106,16 @@ def tree(
                 yield prefix + pointer + path.name
                 files += 1
 
-    print(dir_path.name)
-    iterator = inner(dir_path, level=level)
-    for line in islice(iterator, length_limit):
-        print(line)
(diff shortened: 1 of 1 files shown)


lamindb/_folder.py
```diff
@@ -106,10 +106,16 @@ def inner(dir_path: Path, prefix: str = "", level=-1):
...
+        iterator = inner(dir_path, level=level)
+        for line in islice(iterator, length_limit):
+            folder_tree += f"\n{line}"
+        if next(iterator, None):
+            folder_tree += f"... length_limit, {length_limit}, reached, counted:"
+        print(folder_tree)
+        print(f"\n{directories} directories" + (f", {files} files" if files else ""))
+    except FileNotFoundError:
+        raise NotImplementedError(
+            "Tree display is only supported for existing folders in the storage!"
```
falexwolf, line 120: "Only works for real folders: folders that exist in storage."

PR #752, opened 2023-06-11
✨ Add to_adata method to AnnDataAccessor

Koncopd:
This allows to load subsets of AnnData into memory as actual AnnData objects.
This allows `File.backed()` to supersede `File.stream()`.

### DIFF
--- a/docs/guide/04-stream.ipynb
+++ b/docs/guide/04-stream.ipynb
@@ -1,7 +1,6 @@
 {
  "cells": [
   {
-   "attachments": {},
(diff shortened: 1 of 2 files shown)


falexwolf:
Thanks, @Koncopd!

Would you like me to take a crack at refactoring the streaming notebook, moving some of its content into unit tests & removing `.stream()` and `.subsetter()` from `File`?

Koncopd:
Maybe we should leave stream without subsetting? Just to load an anndata object into memory without caching?

PR #796, opened 2023-06-22
🚚 Move default storage location from `lndb/` to `.lamindb/`

falexwolf:


lamindb/_file_methods.py
```diff
@@ -198,11 +198,15 @@ def inner(dir_path: Union[Path, UPath], prefix: str = "", level=-1):
         nonlocal files, directories
         if not level:
             return  # 0, stop iterating
+        stripped_dir_path = dir_path.as_posix().rstrip("/")
+        # do not iterate through zarr directories
+        if stripped_dir_path.endswith(".zarr"):
```
Koncopd, line 203: Also needed for `.zrad`.

falexwolf: Right, thanks for spotting this!

PR #803, opened 2023-06-26
♻️ Replaced `ln.parse` with `ORM.from_values`

sunnyosun:
with https://github.com/laminlabs/lnschema-bionty/pull/138


lamindb/_from_iter.py
```diff
@@ -0,0 +1,304 @@
+from typing import Any, Dict, Iterable, List, Optional, TypeVar, Union
+
+import deprecation
```
falexwolf, line 3: We've so far used this one, which does not require to add a new dependency: https://github.com/laminlabs/lamindb-setup/blob/d8490c0117f8b267bc3d3d3ee28b373f3cd844ce/lamindb_setup/_set.py#L61

PR #811, opened 2023-06-29
🐛 Fix the variable bug in File.delete

Koncopd:
image


lamindb/_file_methods.py
```diff
@@ -142,10 +142,10 @@ def delete(file, storage: Optional[bool] = None) -> None:
     """
     if storage is None:
         response = input(f"Are you sure you want to delete {file} from storage? (y/n)")
-        if response == "y":
-            delete_in_storage = True
+        delete_in_storage = response == "y"
```
falexwolf, line 145: Thanks!

PR #834, opened 2023-07-03
🚸 Better error behavior when no notebook title set in `ln.track()`

Koncopd:
subj

### DIFF
--- a/lamindb/_context.py
+++ b/lamindb/_context.py
@@ -21,7 +21,15 @@ msg_path_failed = (
 )
 
 
-class NonInteractiveEditorError(Exception):
+class InitWithNonInteractiveEditorError(Exception):
+    pass
+
+
+class NotebookNotSavedError(Exception):
+    pass
+
+
+class NoTitleError(Exception):
     pass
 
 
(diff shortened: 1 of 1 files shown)


lamindb/_context.py
```diff
@@ -179,9 +179,14 @@ def _track(
                         )
                     elif isinstance(e, NonInteractiveEditorError):
                         raise e
+                    elif isinstance(e, IndexError):
```
falexwolf, line 182: Are there other possibilities for creating IndexError?

Can we wrap `nbproject.meta.live.title` and throw a `NotebookNoTitleError` or something like this?

We don't want notebooks without titles to be tracked.

I think this will only ever be raised during init phase, because if we raise an error it will guarantee that there is a `name` in the Transform table and  a title in the notebook.

If we don't raise an error, I fear we might corrupt notebook & database. [Corrrupt means: either no title in notebook or no name in Transform table].

Don't you think?

falexwolf: Context: https://laminlabs.slack.com/archives/C04FPE8V01W/p1688380731268789?thread_ts=1688379577.407719&cid=C04FPE8V01W

PR #878, opened 2023-07-10
✨ Detect interactive environments and other fixes for ln.track()

Koncopd:
subj
also includes fixes for `reinitialize_notebook` (`transform.id` was passed before instead of `transform.stem_id`).

### DIFF
--- a/lamindb/_context.py
+++ b/lamindb/_context.py
@@ -1,6 +1,7 @@
 import builtins
 import os
 import re
+from datetime import datetime
 from pathlib import Path, PurePath
 from typing import Dict, List, Optional, Tuple, Union
 
@@ -20,6 +21,13 @@ msg_path_failed = (
     " `notebook_path` to ln.track()."
 )
 
+msg_manual_init = (
+    "\n(1) Save your notebook!"
+    "\n(2) Attach metadata to the notebook by running the CLI:\n"
+    "lamin track {notebook_path}"
+    "\n(3) Reload or re-open your notebook"
+)
+
 
 class UpdateNbWithNonInteractiveEditorError(Exception):
     pass
@@ -31,6 +39,10 @@ class NotebookNotSavedError(Exception):
 
 class NoTitleError(Exception):
     pass
+
+
+def _seconds_modified(filepath):
+    return datetime.now().timestamp() - Path(filepath).stat().st_mtime
 
 
 def _write_notebook_meta(metadata):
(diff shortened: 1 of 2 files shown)


lamindb/_context.py
```diff
@@ -88,9 +101,10 @@ def reinitialize_notebook(
         new_version = "0"
     metadata["version"] = new_version
 
-    transform = Transform(
-        stem_id=new_id, version=new_version, type=TransformType.notebook
-    )
+    if transform.stem_id != new_id or transform.version != new_version:
```
falexwolf, line 104: Could you add an extensive comment to this line? I don't fully understand what's going on!

lamindb/_context.py
```diff
@@ -20,6 +21,13 @@
     " `notebook_path` to ln.track()."
 )
 
+msg_manual_init = (
+    "\n(1) Save your notebook!"
+    "\n(2) Attach metadata to the notebook by running the CLI:\n"
+    "lamin track {notebook_path}"
```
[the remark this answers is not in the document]
falexwolf: Saw below! All good!

lamindb/_context.py
```diff
@@ -336,20 +351,24 @@ def _track_notebook(
...
             # save the notebook in case that title was updated
             # but notebook not saved
             nbproject.dev._frontend_commands._save_notebook(_env)
+            # check here if interactivity really works
 
         if metadata is not None:
+            # this only executed if nbproject is used
+            # check here if interactivity really works
+            # timestamp should be recent due to save
+            is_interactive = _seconds_modified(_filepath) < 1.5  # should be ~1 sec
```
falexwolf, line 367: I think this is a fine way of doing it, but wouldn't it be even more robust to compare the timestamp prior to saving with after saving and making sure it differs?

This here could fail in some cases. But it's also fine for me to keep it with a comment that we might want to make it more robust in the future.

lamindb/_context.py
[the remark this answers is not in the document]
Koncopd: Pls check further, i use `msg.format(notebook_path=...)` to insert the path.

lamindb/_context.py
Koncopd: Hm, i am not really sure it is more reliable, because autosave can happen, for example.

falexwolf: You're right that this should be fine! After all, we just want to make sure that the notebook just got saved.

PR #918, opened 2023-07-23
🐛 filepath_from_file now returns file._local_filepath if it is present

Koncopd:
also `file.load()` will return `file._memory_rep` if it is present.

### DIFF
--- a/tests/test_file.py
+++ b/tests/test_file.py
@@ -181,6 +181,21 @@ def test_create_from_local_filepath(get_test_filepaths, key, name):
     # file.delete(storage=True)
 
 
+def test_local_path_load():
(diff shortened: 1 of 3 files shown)


tests/test_file.py
```diff
@@ -181,6 +181,21 @@ def test_create_from_local_filepath(get_test_filepaths, key, name):
     # file.delete(storage=True)
 
 
+def test_local_path_load():
+    local_filepath = Path("tests/test-files/pbmc68k.h5ad").resolve()
+
+    file = ln.File(local_filepath)
+    assert local_filepath == file._local_filepath
+    assert local_filepath == file.path()
+    assert local_filepath == file.stage()
```
falexwolf, line 190: Nice! :D

PR #926, opened 2023-07-24
🎉 Stage 0.48 (5)

falexwolf:


lamindb/_queryset.py
```diff
@@ -121,7 +121,11 @@ def df(self, include: Optional[List[str]] = None):
                     if field.field.model != ORM
                     else field.field.related_model
                 )
-                if field.field.model == related_ORM:
+                # print(ORM)
+                # print(field.field.model)
```
Zethson, line 125: I guess all of these can be killed :)

falexwolf: Grrrr

falexwolf: Yes, fo course!

PR #928, opened 2023-07-24
♻️ Remove unnecessary check in `from_dir`

Zethson:


tests/test_file.py
```diff
@@ -176,16 +176,11 @@ def test_local_path_load():
 
 
 def test_init_from_directory(get_test_filepaths):
-    isin_default_storage = get_test_filepaths[0]
```
falexwolf, line 179: I'll need to add this back and define what the storage root for this path should be. Otherwise this operation isn't well-defined.

PR #1010, opened 2023-08-07
🎉 Stage 0.50.1

Koncopd:


lamindb/__init__.py
```diff
@@ -85,7 +85,7 @@
 
 """
 
-__version__ = "0.50.0"  # denote a release candidate for 0.1.0 with 0.1rc1
+__version__ = "0.50.1"  # denote a release candidate for 0.1.0 with 0.1rc1
```
falexwolf, line 88: @Koncopd - this can only be bumped to a release when we make a release. The bump here happens at the same time as the PyPI upload if all integration tests pass.

If you'd like to change this number during the dev cycle, you can only change it to a pre-release like 0.50.1a1 or something similar.

But we haven't indicated pre-releases on staging in the past months - would you prefer to do it going forward?

lamindb/dev/datasets/__init__.py
```diff
@@ -22,13 +22,15 @@
    anndata_pbmc3k_processed
    anndata_with_obs
    anndata_suo22_Visium10X
+   anndata_file_pbmc68k_test
```
falexwolf, line 25: Thanks for this! Could you list it next to `anndata_pbmc68k_reduced` and indicate the difference between these two assets? I'm still a bit confused why we need both in our "test dataset ontology" :D

lamindb/dev/datasets/_core.py
```diff
@@ -132,6 +132,17 @@ def anndata_pbmc68k_reduced() -> ad.AnnData:
     return ad.read(filepath)
 
 
+def anndata_file_pbmc68k_test() -> Path:
+    """Modified from scanpy.datasets.pbmc68k_reduced().
+
+    Additional slots were added for testing purposes.
```
falexwolf, line 138: So far, for each test asset, we've added the exact steps for exact reproducibility of the dataset. Could you please just add the lines of code that were being run similar to how it's done for the other datasets?

tests/test_storage.py
```diff
@@ -29,7 +28,7 @@ def callback(*args, **kwargs):
 
 @pytest.mark.parametrize("adata_format", ["h5ad", "zarr"])
```
falexwolf, line 29: Why don't we adopt the `.zrad` ending that we agreed on, here? 🤔

lamindb/__init__.py
Koncopd: sorry, will do like you say.

lamindb/dev/datasets/__init__.py
Koncopd: Yep, will do.

lamindb/dev/datasets/_core.py
Koncopd: will do.

tests/test_storage.py
Koncopd: It is ".zarr" to test also that even in this case the function understand that this is actually anndata.

falexwolf: Make sense!

lamindb/dev/datasets/_core.py
falexwolf: Thanks!

lamindb/__init__.py
falexwolf: No worries!

PR #1069, opened 2023-08-22
♻️ Use principled conversion to paths with options inheritance

Koncopd:
fixes problems with anon access.


falexwolf:
This is already much better!

PR #1102, opened 2023-08-27
♻️ Refactor to use UPath everywhere

Koncopd:
corresponds to https://github.com/laminlabs/lamindb-setup/pull/490
in progress

### DIFF
--- a/lamindb/_file.py
+++ b/lamindb/_file.py
@@ -63,7 +63,8 @@ def process_pathlike(
                 raise FileNotFoundError(filepath)
         except PermissionError:
             pass
-    filepath = filepath.resolve()  # works for UPath also
+    if isinstance(filepath, LocalPathClasses):
+        filepath = filepath.resolve()
     # check whether the path is in default storage
     default_storage = lamindb_setup.settings.storage.record
     if check_path_is_child_of_root(filepath, default_storage.root_as_path()):
@@ -146,14 +147,14 @@ def process_data(
 
 
 def get_hash(
-    filepath,
+    filepath: UPath,
     suffix,
     filepath_stat=None,
     check_hash: bool = True,
 ) -> Union[Tuple[Optional[str], Optional[str]], File]:
     if suffix in {".zarr", ".zrad"}:
         return None
-    if isinstance(filepath, UPath):
+    if not isinstance(filepath, LocalPathClasses):
         stat = filepath_stat
         if stat is not None and "ETag" in stat:
             # small files
--- a/lamindb/_file.py
+++ b/lamindb/_file.py
@@ -250,8 +250,8 @@ def check_path_is_child_of_root(
     if root is None:
         root = lamindb_setup.settings.storage.root
 
-    filepath = UPath(filepath)
-    root = UPath(root)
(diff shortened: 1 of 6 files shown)


lamindb/_file.py
```diff
@@ -55,16 +56,15 @@
...
 ) -> Tuple[Storage, bool]:
     if not skip_existence_check:
         try:  # check if file exists
             if not filepath.exists():
                 raise FileNotFoundError(filepath)
         except PermissionError:
             pass
-    if not isinstance(filepath, UPath):
-        filepath = filepath.resolve()
+    filepath = filepath.resolve()  # works for UPath also
```
falexwolf, line 67: This isn't needed for a CloudPath object as by definition, the cloudpath is fully specified including the bucket.

Hence: The check for a local path makes sense. Only if it's a local path, we'd like to resolve it. Calling it on CloudPath, too, causes an API request, I fear, and hence unnecessarily reduces performance.

I've made several experiments ingesting high numbers of files and found that reducing the number of API requests is the only factor that determines performance if files aren't big.

lamindb/_file.py
```diff
@@ -229,16 +229,8 @@ def get_path_size_hash(
...
-                try:
-                    size = filepath_stat["size"]  # type: ignore
-                # here trying to fix access issue with new s3 buckets
-                except Exception as e:
-                    if filepath._url.scheme == "s3":
-                        filepath = UPath(filepath, cache_regions=True)
-                        size = filepath_stat["size"]  # type: ignore
-                    else:
-                        raise e
+            if not isinstance(filepath, LocalPathClasses):
```
falexwolf, line 232: Great!

lamindb/_file.py
Koncopd: Agree, will remove it.

Koncopd:
`_file.py` is still a total mess, maybe need to enforce UPath everywhere instead of mix due to usage of `Path.cwd()` etc.

falexwolf:
But yeah, we don't want a "total mess" anymore. 😅

lamindb/_file.py
```diff
@@ -387,7 +386,7 @@ def log_storage_hint(
     if check_path_in_storage:
         display_root = storage.root  # type: ignore
         # check whether path is local
-        if not storage.root.startswith(("s3://", "gs://")):  # type: ignore
+        if fsspec.utils.get_protocol(storage.root) == "file":  # type: ignore
```
falexwolf, line 389: Why don't you use the `isinstance(filepath, LocalPathClasses)` check here?

Koncopd: Here `storage` is a model with `root` being a string.

falexwolf: Yes, you're right! 

To make this more readable (`fsspec.utils.get_protocol(storage.root) == "file"` isn't understandable at all), could we introduce a helper function?

```
def is_local_path(path: [str, UPath, Path]) -> bool:
    if isinstance(path, str):
         return fsspec.utils.get_protocol(storage.root) == "file"
    else:
         return isinstance(path, LocalPathClasses)
```

This could be imported like so:
```
from lamindb.dev.upath import is_local_path, is_cloud_path
```
And then be used anywhere.

(Not sure whether the `_path` is needed.)

PR #1160, opened 2023-09-11
🎨 More validation for `File` creation, fix some aspects of zarr backend

Koncopd:
- Fixes `zarr` or `zrad` objects upload from disk and adds a test for this.
- Adds validation of `key`'s suffix for in-memory objects in `File`.

### DIFF
--- a/lamindb/_file.py
+++ b/lamindb/_file.py
@@ -104,7 +104,8 @@ def process_pathlike(
 def process_data(
     provisional_id: str,
     data: Union[PathLike, DataLike],
-    format,
+    format: Optional[str],
+    key: Optional[str],
     skip_existence_check: bool = False,
 ) -> Tuple[Any, Union[Path, UPath], str, Storage, bool]:
     """Serialize a data object that's provided as file or in memory."""
@@ -119,7 +120,19 @@ def process_data(
     elif isinstance(data, (pd.DataFrame, AnnData)):  # DataLike, spelled out
         storage = lamindb_setup.settings.storage.record
         memory_rep = data
+        if key is not None:
+            key_suffix = PurePosixPath(key).suffix
+            # use suffix as the (adata) format if the format is not provided
+            if isinstance(data, AnnData) and format is None:
+                format = key_suffix[1:]
+        else:
+            key_suffix = None
         suffix = infer_suffix(data, format)
+        if key_suffix is not None and key_suffix != suffix:
+            raise ValueError(
+                f"The suffix '{key_suffix}' of the provided key is incorrect, it should"
+                f" be '{suffix}'."
+            )
         cache_name = f"{provisional_id}{suffix}"
         filepath = lamindb_setup.settings.storage.cache_dir / cache_name
         # Alex: I don't understand the line below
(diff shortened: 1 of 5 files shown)


lamindb/_file.py
```diff
@@ -119,13 +120,25 @@ def process_data(
     elif isinstance(data, (pd.DataFrame, AnnData)):  # DataLike, spelled out
         storage = lamindb_setup.settings.storage.record
         memory_rep = data
+        if key is not None:
```
falexwolf, line 123: Could this call the following functionality instead of implementing new logic?

https://github.com/laminlabs/lamindb/blob/51b220ddb09ff13d87b68b9c331d2cd7b60691c0/lamindb/dev/storage/file.py#L51-L65

Inferring the suffix correctly is a science on its own. :)

lamindb/_file.py
```diff
@@ -119,13 +120,25 @@ def process_data(
...
+            key_suffix = PurePosixPath(key).suffix
+            # use suffix as the (adata) format if the format is not provided
+            if isinstance(data, AnnData) and format is None:
+                format = key_suffix[1:]
+        else:
+            key_suffix = None
         suffix = infer_suffix(data, format)
+        if key_suffix is not None and key_suffix != suffix:
```
falexwolf, line 131: Great!

tests/test_file.py
```diff
@@ -559,14 +560,22 @@ def test_serialize_paths():
     up_str = "s3://lamindb-ci/test-data/test.csv"
     up_upath = UPath(up_str)
 
-    _, filepath, _, _, _ = process_data("id", fp_str, None, skip_existence_check=True)
+    _, filepath, _, _, _ = process_data(
```
falexwolf, line 563: This would be much more easy to read if it used `@pytest.mark.parametrize` - there are many examples for this in `test_file.py`. Could you use one of them?

tests/test_file.py
```diff
@@ -624,3 +633,62 @@ def test_file_zarr():
     file.save()
     file.delete(storage=False)
     UPath("test.zarr").unlink()
+
+
+def test_zarr_folder_upload():
```
falexwolf, line 638: Is this an actual folder, or is it a zarr object (represented as a folder)?

Could you add a comment here?

tests/test_file.py
```diff
@@ -624,3 +633,62 @@ def test_file_zarr():
...
+    file.save()
+
+    assert isinstance(file.path, CloudPath) and file.path.exists()
+
+    file.delete(storage=True)
+    delete_storage(zarr_path)
+    ln.settings.storage = previous_storage
+
+
+def test_df_suffix():
```
falexwolf, line 658: Great!

tests/test_file.py
```diff
@@ -624,3 +633,62 @@ def test_file_zarr():
...
+
+    with pytest.raises(ValueError) as error:
+        file = ln.File(df, key="test_.def")
+    assert (
+        error.exconly().partition(",")[0]
+        == "ValueError: The suffix '.def' of the provided key is incorrect"
+    )
+
+
+def test_adata_suffix():
```
falexwolf, line 670: Can you please use `pytest.mark.parametrize`, here? Otherwise it's hard to read and to assess completeness.

falexwolf:
This is great, Sergei!

I just have small remarks in the comments above.

tests/test_file.py
Koncopd: It is a zarr object represented as an actual folder.

tests/test_file.py
Koncopd: Different combinations can raise different errors or require different asserts, that is why not parametrized. Parametrization here becomes a mess actually in my opinion.

lamindb/_file.py
Koncopd: I will check that.

tests/test_file.py
falexwolf: Ok, then! No parametrization!

tests/test_file.py
Koncopd: I am really not sure. I mean i don't do parametrizations for some tests because it turns into some cross lists with repeating parameters etc.
Here i would need to match different results to different parameters. Maybe it can be done, but why?

falexwolf: It's ok, Sergei. I think it's easier to read but this case here is so simple that it's also OK to read without it. Keep it as is if you don't agree! :)

falexwolf: Maybe I'm off here and it'd in fact be harder to read!

falexwolf:
Good to merge!

lamindb/_file.py
Koncopd: Done.

PR #1168, opened 2023-09-13
🔇 Do not log transform twice when a transform in updated

sunnyosun:


sunnyosun:
@falexwolf Is this ok?

PR #1220, opened 2023-10-03
♻️ Do not use `DatabaseWrapper` directly

Koncopd:
https://github.com/django/django/blob/679212a47ac3e22a6fbb6ee3cd4c09f29aae8b5d/django/utils/connection.py#L56
https://github.com/django/django/blob/679212a47ac3e22a6fbb6ee3cd4c09f29aae8b5d/django/db/utils.py#L191

Using a specific `DatabaseWrapper` seems to be unneeded, this might be important if we want to add support for databases other than sqlite and postgres.

### DIFF
--- a/lamindb/_filter.py
+++ b/lamindb/_filter.py
@@ -3,10 +3,6 @@ from uuid import UUID
 
 import dj_database_url
 from django.db import connections
-from django.db.backends.postgresql.base import (
-    DatabaseWrapper as DatabaseWrapperPostgres,
-)
-from django.db.backends.sqlite3.base import DatabaseWrapper as DatabaseWrapperSQLite
 from lamindb_setup._init_instance import InstanceSettings
 from lamindb_setup._load_instance import get_owner_name_from_identifier
 from lamindb_setup.dev._hub_core import load_instance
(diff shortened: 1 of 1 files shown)


Koncopd:
Shouldn't `add_db_connection` be in `lamindb-setup` actually?

falexwolf:
@fredericenard said this is needed at some point! I'm not sure why.

falexwolf:
https://laminlabs.slack.com/archives/C04T2TQC7LL/p1695315236292569

Koncopd:
@falexwolf this lets django do exactly the thing that Fred suggests to do manually but in a more correct way, i.e. django infers database wrapper automatically. You can check the code in the links, it does exactly that.

Koncopd:
@fredericenard https://github.com/laminlabs/lamindb/blob/main/docs/transfer.ipynb
this notebook

setting `connections.settings ` instead of `connections` allows to provide a db config. The links to the relevant code in django are in the post.

falexwolf:
> Shouldn't add_db_connection be in lamindb-setup actually?

Yes, let's discuss this! But that will be another PR.

PR #1227, opened 2023-10-10
⬆️ Make backed compatible with anndata==0.10.1

Koncopd:
subj


Koncopd:
Lower coverage is due to tests run on only one version of `anndata`.

falexwolf (approved):
This is great, Sergei!

PR #1294, opened 2023-11-10
🚑  Fix visibility choices, updated intro

sunnyosun:


lamindb/_dataset.py
```diff
@@ -210,9 +210,10 @@ def from_df(
     reference_type: Optional[str] = None,
     version: Optional[str] = None,
     is_new_version_of: Optional["File"] = None,
+    **kwargs,
```
falexwolf, line 213: Why would you like to introduce `**kwargs`?

falexwolf:
The simplification is awesome! 😄

lamindb/_dataset.py
sunnyosun: Because I want to pass organism=

PR #1298, opened 2023-11-14
🚸 Ensure saved notebooks aren't rerun with the same version

falexwolf:
https://laminlabs.slack.com/archives/C04FPE8V01W/p1699443688434429

### DIFF
--- a/lamindb/dev/_run_context.py
+++ b/lamindb/dev/_run_context.py
@@ -32,15 +32,6 @@ msg_manual_init = (
 )
 
 
-def save_or_load(msg):
-    if logger._verbosity > 2:
-        logger.success(msg)
-    else:
-        logger.important(
-            msg.replace("loaded: ", "").replace("saved: ", "").replace("updated: ", "")
-        )
-
-
 class UpdateNbWithNonInteractiveEditorError(Exception):
     pass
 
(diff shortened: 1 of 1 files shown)


lamindb/dev/_run_context.py
```diff
@@ -468,6 +468,26 @@ def _track_notebook(
...
+            if (
+                transform.source_file_id is not None
+                or transform.latest_report_id is not None
+            ):
+                response = input(
+                    "You already saved a source file and a report for this transform"
+                    " version! Do you want to increase the version number? (y/n)"
+                )
+                if response == "y":
+                    if _env in ("lab", "notebook") and is_interactive:
```
falexwolf, line 480: Hey @Koncopd, do we need `_env in ("lab", "notebook")`?

The only thing that matters should be `is_interactive`, right?

Koncopd: This is potentially different. Also this is used before `is_interactive` is set. I would really avoid changing something in the logic here because we don't really test the interactive features, it is pretty easy to break something subtle without realizing. It need a refactor, but very careful and not as a part of a PR with another purpose.

Koncopd: At this exact place it looks you are right, but see above...

falexwolf: > It needs a refactor, but very careful and not as a part of a PR with another purpose.
Agreed!

lamindb/dev/_run_context.py
```diff
@@ -32,15 +32,6 @@
 )
 
 
-def save_or_load(msg):
-    if logger._verbosity > 2:
-        logger.success(msg)
-    else:
-        logger.important(
-            msg.replace("loaded: ", "").replace("saved: ", "").replace("updated: ", "")
```
falexwolf, line 40: @sunnyosun, removing this as, over the past months of this having been live, I felt it's really irritating to not know whether a transform or run are loaded or freshly created

PR #1307, opened 2023-11-18
✨ Initial implementation of IndexedDataset

Koncopd:
subj
in progress


lamindb/dev/dataloader/dataset.py
```diff
@@ -0,0 +1,182 @@
...
+from typing import List, Optional, Union
+
+import numpy as np
+from lamindb_setup.dev.upath import UPath
+
+from ..storage._backed_access import ArrayTypes, GroupTypes, StorageType, registry
+
+
+# this is based on sCimilarity (https://github.com/Genentech/scimilarity) Datasets
+class ListDataset:
```
falexwolf, line 12: How about calling this `IndexedDataset`?

lamindb/_dataset.py
```diff
@@ -329,6 +332,23 @@ def from_files(files: Iterable[File]) -> Tuple[str, Dict[str, str]]:
     return hash, feature_sets_union
 
 
+def filelist_dataset(
```
falexwolf, line 335: How about `indexed()`?

falexwolf: Or `to_indexed()`?

falexwolf:
I'm just realizing that torch calls this Map-style dataset. But I'm almost tempted to think that that `IndexedDataset` will be clearer to users than `MappedDataset`. Or what do you think, @Koncopd?
image

Koncopd:
Yes, you are right. Not sure, we don't expose the class anyways. I am fine with both.

PR #1350, opened 2023-12-12
🐛 Fix size and hash for local folders in Artifact

Koncopd:
`glob("*")` only gives the first level files and folders.


falexwolf:
The strange thing is that this hash-compute differs across my local setup and the GA action worker (I've added tests in the next PR):
image

PR #1358, opened 2023-12-13
👷 Use ruff and clean up pre-commit

Zethson:


.pre-commit-config.yaml
```diff
@@ -1,77 +1,52 @@
...
+minimum_pre_commit_version: 2.16.0
 repos:
-  - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v3.2.0
-    hooks:
-      - id: trailing-whitespace
-      - id: end-of-file-fixer
-        exclude: |
-          (?x)(
-              .github/workflows/latest-changes.jinja2
```
falexwolf, line 9: Are you sure we don't need to exclude it?

.github/workflows/latest-changes.jinja2
```diff
@@ -1,2 +1 @@
 - {{pr.title}} [PR{{pr.number}}]({{pr.html_url}}) [@{{pr.user.login}}]({{pr.user.html_url}})
```
falexwolf, line 1: This will break things AFAIK remember

.pre-commit-config.yaml
falexwolf: I think you need to add it back so that things don't break, I messed with this early on

falexwolf:
Looks great except for the latest-changes.jinja, which I  believe requires the white space!

Otherwise good to merge!

falexwolf (approved):

PR #1369, opened 2023-12-20
✨ Enable filtering by latest version and version family

bpenteado:
https://laminlabs.slack.com/archives/C04FPE8V01W/p1703160270853689


falexwolf:
Great, Bernardo! Could you simply replicate the test here that you already wrote in `laminapp-rest`?

tests/test_queryset.py
```diff
@@ -128,3 +130,39 @@ def test_validate():
...
+        index=["measurement1", "measurement2", "measurement3"],
+    )
+    artifact_2 = ln.Artifact(
+        df_2,
+        description="Versioned test artifact",
+        is_new_version_of=artifact_1,
+    )
+    artifact_2.save()
+
+    # Filter artifact by latest version
```
falexwolf, line 156: How do we pass information about the version family here? 🤔 

Without characterizing the family, the "latest" query won't meet most users needs.

An easy way would be to pass `initial_version`, but the problem with this is that one has to know the initial version, which seems impractical.

falexwolf: I think there are two user stories, here.

One is the one from the UI, or PyPI, or S3, or any other versioned experience where you don't whatever you're looking at be cluttered with all previous versions of the same thing.

The other one is to query for the latest version of a given version family, as in, "what's the latest version of this notebook".

falexwolf: And damn it, the query would again be much simpler if we also labeled the seed element of the family with `initial_version`... it seems hairy and non-idiomatic to construct logic to fulfill user story #2 without it? Does it?

I mean, I think this here is something I can get used to
```
ln.Transform.filter(initial_version=my_reference_notebook.initial_version, version="latest")
```

But is this going to account for the seed element of the version family defined by `my_reference_notebook.initial_version`?

bpenteado: Hmm, I see how this becomes hairy.

If you already have `my_reference_notebook`, couldn't you simply query for it:

```
my_notebook = ln.Transform.filter(id=my_reference_notebook.id).one()
```

And then list the versions via `.versions()`:

```
mynotebook_versions = my_notebook.versions()
```

And possibly also implement a `.latest_version()` property to have it be easily accessible?

I'm going to have to think about the straight versioned query. It's not immediately clear to me...

falexwolf: The problem with this is that it takes two network requests and two lines of code:
image

For something as fundamental and basic, I'd expect a single network request and one line of code.

Could you try the following out, which you can then also use for writing a test for this?
```
ln.Transform.filter(initial_version=my_reference_notebook.initial_version, version="latest")
```

In the worst case, we need to think about labeling the seed element of the version family with its own primary key.

bpenteado: Agreed. Will try the query out and update my thinking based on your Slack arguments.

falexwolf:
@bpenteado, please note the changes to the versioning framework: https://github.com/laminlabs/lamindb/pull/1375

It'd still be great to have your tests & docs. Some small things like the `versions` property won't be needed anymore. I'll mention you in the changelog also for the other PR; much of the thought and the design is due to you working through these different queries.

falexwolf:
With the below, I'm closing this PR, @bpenteado:

- https://github.com/laminlabs/laminapp-rest/pull/229

Let's discuss everything that happened during the holidays, and why I heavily deviated from the original spec we discussed when embarking on this topic in the last week before Christmas!

PR #1380, opened 2024-01-02
✨ Add outer join and categories caching to Collection.mapped

Koncopd:
- Adds virtual outer join
- Improves access speed by caching categories of labels.

### DIFF
--- a/tests/test_collection.py
+++ b/tests/test_collection.py
@@ -300,21 +300,25 @@ def test_collection_mapped():
     assert ls_ds.closed
     del ls_ds
 
-    with collection.mapped(label_keys="feat1", join_vars="inner") as ls_ds:
+    with collection.mapped(
+        label_keys="feat1", join_vars="inner", dtype="float32"
+    ) as ls_ds:
         assert not ls_ds.closed
         assert len(ls_ds) == 4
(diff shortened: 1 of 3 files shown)


tests/test_collection.py
```diff
@@ -292,18 +300,34 @@ def test_collection_mapped():
...
         assert len(ls_ds[0]) == 2 and len(ls_ds[2]) == 2
+        assert str(ls_ds[0][0].dtype) == "float32"
+        assert str(ls_ds[2][0].dtype) == "float32"
     assert ls_ds.closed
 
     ls_ds = collection.mapped(label_keys="feat1", parallel=True)
     assert len(ls_ds[0]) == 2 and len(ls_ds[2]) == 2
 
+    # adata3 goes first here
+    with collection_outer.mapped(label_keys="feat1", join_vars="auto") as ls_ds:
```
falexwolf, line 317: What does `auto` do? Why did you decide against an explicit test of `outer`?

Koncopd: I need probably to expand on this in the corresponding docstring. `join_vars="auto"` checks if it needs and is able to preforms inner join and if not sets `join_vars="outer"`. This piece also tests this logic.

PR #1390, opened 2024-01-07
🎨 Introduce ordered collections and simplify `.mapped()`

falexwolf:
- https://github.com/laminlabs/lnschema-core/pull/333

### DIFF
--- a/tests/test_collection.py
+++ b/tests/test_collection.py
@@ -300,9 +300,7 @@ def test_collection_mapped():
     assert ls_ds.closed
     del ls_ds
 
-    with collection.mapped(
-        label_keys="feat1", join_vars="inner", dtype="float32"
-    ) as ls_ds:
+    with collection.mapped(label_keys="feat1", join="inner", dtype="float32") as ls_ds:
         assert not ls_ds.closed
         assert len(ls_ds) == 4
         assert len(ls_ds[0]) == 2 and len(ls_ds[2]) == 2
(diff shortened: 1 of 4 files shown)


tests/test_collection.py
```diff
@@ -314,8 +312,8 @@ def test_collection_mapped():
     assert len(ls_ds[0]) == 2 and len(ls_ds[2]) == 2
 
     # adata3 goes first here
-    with collection_outer.mapped(label_keys="feat1", join_vars="auto") as ls_ds:
-        assert ls_ds.join_vars == "outer"
+    with collection_outer.mapped(label_keys="feat1", join="inner") as ls_ds:
+        assert ls_ds.join_vars == "inner"
```
Koncopd, line 316: collection_outer has adatas with non-intersecting vars, this collection should be used to test outer join, not inner.

Koncopd:
I don't think removing "auto" is necessary, it is easier to check with "auto" first then to manually check variables. Also it is important to avoid any join if the variables are exactly the same for performance reasons.

Koncopd:
I believe reoving join_vars="auto" doesn't simplify anything, just requires a user to do additional actions manually instead.

tests/test_collection.py
falexwolf: Fixed!

falexwolf:
Pasting our discussion for reference: https://laminlabs.slack.com/archives/C03Q5TXF797/p1704653536322439?thread_ts=1704228794.520039&cid=C03Q5TXF797

tests/test_collection.py
```diff
@@ -314,18 +317,29 @@ def test_collection_mapped():
...
 
     # adata3 goes first here
-    with collection_outer.mapped(label_keys="feat1", join_vars="auto") as ls_ds:
+    with collection_outer.mapped(label_keys="feat1", join="outer") as ls_ds:
         assert ls_ds.join_vars == "outer"
         assert len(ls_ds.var_joint) == 6
         assert len(ls_ds[0]) == 2
         assert len(ls_ds[0][0]) == 6
-        assert np.array_equal(ls_ds[1][0], np.array([4, 5, 8, 0, 0, 0]))
+        print(ls_ds[0][0])
```
Zethson, line 325: Are those print statements a relict from making the tests work?

falexwolf: Yes, and a slight worry that this output might be helpful in the future again. It took me a while to understand the test, and so I thought I'd leave these statements there for a little longer.

PR #1391, opened 2024-01-08
🎨 Not create organism records when calling `.bionty()`

sunnyosun:


falexwolf:
It'd be good if the up-stream PR in lnschema-bionty was linked here!

PR #1405, opened 2024-01-16
🎨 Transfer collection.artifacts

sunnyosun:


lamindb/_registry.py
```diff
@@ -468,6 +469,12 @@ def transfer_to_default_db(
 def save(self, *args, **kwargs) -> None:
     db = self._state.db
     pk_on_db = self.pk
+    artifacts = []
+    if self.__class__.__name__ == "Collection":
+        try:
+            artifacts = self.artifacts.list()
+        except Exception:
+            artifacts = []
```
Koncopd, line 477: There should be a specific exception here, no?

sunnyosun: Yes! Fixed!

Issue #1409, opened 2024-01-25
🐛 Collection `save()` is non-idempotent


falexwolf:
https://github.com/laminlabs/lamindb/blob/bf1c81e7ae8c177a7338d560521cc5b0c1cdc1aa/lamindb/_collection.py#L422-L442

PR #1410, opened 2024-01-25
🎨 Make `Collection.save()` ACID

falexwolf:
- https://github.com/laminlabs/lamindb/issues/1409


falexwolf:
@sunnyosun - I think we also need to make sure that it's ACID and encapsulate it into one atomic transaction.

PR #1416, opened 2024-01-29
✨ Enable to use hosted storage on S3

falexwolf:
- https://github.com/laminlabs/lamindb-setup/pull/630
- https://github.com/laminlabs/lamindb-setup/pull/626


falexwolf:
I don't know. I added it to s3://lamindb-ci. Maybe, before, it was silently ignored. Right now, there is a hard error if one can't infer the region.

But it should work according to how the bucket is configured.

I'll figure this one out another time.

PR #1420, opened 2024-01-30
♻️ Use Click for CLI

falexwolf:
- https://github.com/laminlabs/lamin-cli/pull/22

### DIFF
--- a/docs/bionty-base.md
+++ b/docs/bionty-base.md
@@ -1,3 +1,3 @@
-# Bionty-base
+# `bionty-base`
 
 https://lamin.ai/docs/bionty-base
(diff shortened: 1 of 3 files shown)


falexwolf:
Running tests against the rebased lamin-cli!

falexwolf:
@chaichontat - I think everything works with the exception that you're not treating return codes correctly

image

The CLI should return an error code when it fails. Right now it returns 0.

chaichontat:
> @chaichontat - I think everything works with the exception that you're not treating return codes correctly
> 
> image
> The CLI should return an error code when it fails. Right now it returns 0.

```python
import os
from subprocess import run

env = os.environ
env["LAMIN_TESTING"] = "true"
print(run("lamin save x.ipynb", env=env, shell=True).returncode)
```
ValueError: Notebook cannot be empty. It must have at least a title cell.

Return code: 1

falexwolf:
Why is the return code in the pytest session than `0`?

image

PR #1430, opened 2024-02-04
🐛 Fix upload

falexwolf:


Koncopd:
Hm, strange that the tests didn't fail before.

Koncopd:
Ah, ok, i see why it still worked fine
image

Koncopd:
image

falexwolf:
Yes, I was also confused that the tests didn't fail. We only just found out in the web platform.

Could you add a test that would have made it fail?

Koncopd:
I can add, but not now. Maybe tmr.

falexwolf:
Not urgent!

PR #1443, opened 2024-02-19
✨ Add unknown_label and use dict for output in Collection.mapped

Koncopd:
Subj. Also allows to select label keys for encoding.


jkobject (commented):
I would rename "get_label_weights" to "get_label_freqs" or something. Because the weight one would want might often not be "1/count" this seems too much for most applications..

Koncopd:
https://github.com/laminlabs/lamindb/pull/1460

Koncopd:
But i think calling this frequencies is more misleading.

PR #1492, opened 2024-03-12
🩹 Handle OSError during delete

sunnyosun:


sunnyosun:
@Koncopd @falexwolf Do you understand why the CI is failing here?

PR #1502, opened 2024-03-15
adding the ability to output the dataset

jkobject:
Hello Sergey, Alex,

I just realized I had made one more update to .mapped() in which I am adding the ability to output the dataset. I need it to remove batch effects in my model like scGPT does.

Let me know if that is ok. Once this is updated I can use this version :)

### DIFF
--- a/lamindb/core/_mapped_collection.py
+++ b/lamindb/core/_mapped_collection.py
@@ -79,6 +79,7 @@ class MappedCollection:
         unknown_label: Optional[Union[str, Dict[str, str]]] = None,
         cache_categories: bool = True,
         parallel: bool = False,
+        add_dataset_id: bool = False,
         dtype: Optional[str] = None,
     ):
         assert join in {None, "inner", "outer"}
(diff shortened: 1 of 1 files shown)


falexwolf:
Thanks @jkobject! @Koncopd could you take a look at this so that it gets into the new release?

lamindb/core/_mapped_collection.py
```diff
@@ -79,6 +79,7 @@ def __init__(
         unknown_label: Optional[Union[str, Dict[str, str]]] = None,
         cache_categories: bool = True,
         parallel: bool = False,
+        add_dataset_id: bool = False,
```
falexwolf, line 82: @Koncopd - I don't fully understand what this does, but likely it deserves a better name.

Koncopd:
Hm, maybe return this by default without any additional arguments? As we return a dict now, it is not a problem if we add some additinal keys.

falexwolf:
Sounds like the right thing, Sergei!

Koncopd:
@jkobject are you fine with the suggestion?

jkobject:
Works with me!

Koncopd:
done here https://github.com/laminlabs/lamindb/pull/1504

PR #1521, opened 2024-03-25
✨ Collection stage

Koncopd:


lamindb/_collection.py
```diff
@@ -467,7 +477,6 @@ def restore(self) -> None:
 @doc_args(Collection.artifacts.__doc__)
 def artifacts(self) -> QuerySet:
     """{}."""
-    _track_run_input(self)
```
Koncopd, line 470: @falexwolf i don't think track should be here because

- just querying for an artifact is not tracked anywhere else and `collection.artifacts` is equivalent to querying.
- there is no way to pass `is_run_input`.
- `_track_run_input` is duplicated in all methods that access `collection.artifacts` (such as `mapped` etc).

falexwolf: I agree this is problematic

I had added it at a time when we didn't have mapped because the assumption was that this is how you iterate through the collection 

I agree it should now be removed

Koncopd: Ok, great.

PR #1541, opened 2024-03-31
💄 Polish validate & annotate flow

falexwolf:
I'd make reference to the slots in the constructor and introduce `add_new_columns()` here

image

Here I'd not talk about "features" and "labels" but about "records in ln.ULabel for perturbation" and "records in ln.Feature for column names" or "columns":

image


sunnyosun:
Continued in https://github.com/laminlabs/lamindb/pull/1542

PR #1550, opened 2024-04-04
🚸 Error if the exact same artifact is returned from trash

falexwolf:
FYI: @timtreis @chaichontat


falexwolf:
@sunnyosun do you also think this is better?

Background: https://laminlabs.slack.com/archives/C03P6D8U1PC/p1712156528130079

chaichontat:
Can we use this exception? https://docs.python.org/3/library/exceptions.html#FileExistsError

falexwolf:
Done!

PR #1554, opened 2024-04-09
✨ Support MuData

sunnyosun:


lamindb/_annotate.py
```diff
@@ -410,6 +420,199 @@ def save_artifact(self, description: str, **kwargs) -> Artifact:
...
+            ``{"modality_1": bt.Gene.ensembl_gene_id, "modality_2": ln.CellMarker.name}``
+        categoricals: A dictionary mapping ``.obs.columns`` to a registry field.
+            For example:
+            ``{"cell_type_ontology_id": bt.CellType.ontology_id, "donor_id": ln.ULabel.name}``
+        using: A reference LaminDB instance.
+    """
+
+    def __init__(
+        self,
+        mdata: MuData,
```
falexwolf, line 439: I fear this might make it dependent on mudata. Maybe I'm wrong. But you should try in an environment that doesn't have mudata installed.

falexwolf: (I know you have if TYPE_CHECKING up there, but still!)

lamindb/_artifact.py
```diff
@@ -296,7 +305,7 @@ def get_relative_path_to_directory(
 
 def get_artifact_kwargs_from_data(
     *,
-    data: Path | UPath | str | pd.DataFrame | AnnData,
+    data: Path | UPath | str | pd.DataFrame | AnnData | MuData,
```
falexwolf, line 308: Also this might be problematic.

falexwolf (approved):
This looks good! It's a bit complicated, but I fear that's just the way it is.

Please double check that the typing is possible also when mudata isn't installed.

lamindb/_artifact.py
sunnyosun: All CI doesn't have mudata installed except for docs

falexwolf: OK, great!

falexwolf:
It's all tested in the use cases.

falexwolf:
It'd be nice to collect coverage from use cases similar to how we build docs across repositories.

I think this shouldn't be too hard given we built all the infra already for the docs artifacts.

PR #1557, opened 2024-04-10
🐛 Fix passing organism to Annotate

sunnyosun:
Fixes https://github.com/laminlabs/lamin-usecases/issues/115

### DIFF
--- a/lamindb/_annotate.py
+++ b/lamindb/_annotate.py
@@ -550,17 +550,19 @@ class MuDataAnnotator:
 
     def add_validated_from(self, key: str, modality: str | None = None, **kwargs):
         """Add validated categories."""
+        self._kwargs.update(kwargs)
         modality = modality or "obs"
         if modality in self._df_annotators:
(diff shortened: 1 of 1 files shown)


lamindb/_annotate.py
```diff
@@ -550,17 +550,19 @@ def add_new_from_var_index(self, modality: str, **kwargs):
...
     def add_new_from(self, key: str, modality: str | None = None, **kwargs):
         """Add validated & new categories."""
+        self._kwargs.update(kwargs)
         modality = modality or "obs"
         if modality in self._df_annotators:
             df_annotator = self._df_annotators[modality]
-            df_annotator.add_new_from(key=key, **kwargs)
+            df_annotator.add_new_from(key=key, **self._kwargs)
 
     def validate(self, **kwargs) -> bool:
```
falexwolf, line 567: This seems like kwargs is ignored here, is it?

I think we should consider making it a priority to eliminate all kwargs from the API.

PR #1562, opened 2024-04-11
✨ Add `.obsm` and  `.layers` to `MappedCollection`

Koncopd:
Also https://github.com/laminlabs/lnschema-core/pull/367

### DIFF
--- a/lamindb/_collection.py
+++ b/lamindb/_collection.py
@@ -209,7 +209,9 @@ def from_artifacts(artifacts: Iterable[Artifact]) -> tuple[str, dict[str, str]]:
 # docstring handled through attach_func_to_class_method
 def mapped(
     self,
-    label_keys: str | list[str] | None = None,
+    layers_keys: str | list[str] | None = None,
+    obs_keys: str | list[str] | None = None,
+    obsm_keys: str | list[str] | None = None,
     join: Literal["inner", "outer"] | None = "inner",
     encode_labels: bool | list[str] = True,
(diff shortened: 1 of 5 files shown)


lamindb/_collection.py
```diff
@@ -209,7 +209,9 @@ def from_artifacts(artifacts: Iterable[Artifact]) -> tuple[str, dict[str, str]]:
 # docstring handled through attach_func_to_class_method
 def mapped(
     self,
-    label_keys: str | list[str] | None = None,
+    layers_keys: str | list[str] | None = None,
+    obs_keys: str | list[str] | None = None,
```
falexwolf, line 213: I wouldn't make this change without first deprecating `label_keys`, Sergei. It's going to break people's code. I mean, you could simply add one line of logic that feeds `label_keys` into `obs_keys` and prints a warning, and thing things wouldn't break.

I also wouldn't introduce all this anndata jargon in `mapped()`. It'd turn into an ugly method.

In another PR, towards an API that's pretty in the longterm, I suggest
```
collection.to_mapped.via_anndata(...) -> mudata ...
collection.to_mapped.via_df(...) -> parquet ...
```

Koncopd: I am not sure it makes a difference taking into account all other changes like key names changes in `__getitem__`.

falexwolf: It's alright, I'll make a clear note in the changelog! I'm merging!

PR #1572, opened 2024-04-16
🚚 Deprecate `stage()` in favor of `cache()`

falexwolf:
https://www.notion.so/laminlabs/Data-sharing-through-hybrid-local-remote-LaminDB-instances-e01025ed0e3e4f06948c4390894f20a8

- https://github.com/laminlabs/lamindb-setup/pull/717
- https://github.com/laminlabs/lnschema-core/pull/369
- https://github.com/laminlabs/lamin-cli/pull/48


Koncopd:
I really don't like download because it actually synchronizes, stage is much better here.

PR #1577, opened 2024-04-18
🚚 Rename resources in `core.storage` & eliminate `.zrad` from code base

falexwolf:
No chances other than renaming and eliminating `.zrad`.


falexwolf:
What the heck is this fixture? 😆  @Koncopd 

image

falexwolf:
I see, it's an invalid zrad file.
image

PR #1587, opened 2024-04-20
🐛 Fix detecton of AnnData in zarr and h5ad, fix directory upload paths

Koncopd:


Koncopd:
@falexwolf it looks like [your commit](https://github.com/laminlabs/lamindb/commit/2744df75e87efd27e1ee9c7847851b3e624ece91) broke the `test / tutorial`

lamindb/_artifact.py
```diff
@@ -461,23 +473,28 @@ def data_is_mudata(data: MuData | UPathStr):
...
             logger.warning("data is a DataFrame, please use .from_df()")
             accessor = "DataFrame"
-        elif data_is_anndata(data):
-            logger.warning("data is an AnnData, please use .from_anndata()")
+            return accessor
+
+        data_is_path = isinstance(data, (str, Path))
+        if data_is_anndata(data):
+            if not data_is_path:
+                logger.warning("data is an AnnData, please use .from_anndata()")
```
Koncopd, line 491: @falexwolf `.from_anndata` implies only in-memory object or not? I can see from typing and examples, that only in-memory object is allowed, so i add this suppression of the warning for adatas on disk. Is this correct?

Koncopd: Or we also should allow now `Artifact.from_anndata("something.h5ad")`?

falexwolf (approved):
You could add a check for `.anndata.zarr` because we consider it a valid suffix

Koncopd:
> You could add a check for .anndata.zarr because we consider it a valid suffix
Done

lamindb/_artifact.py
falexwolf: Only in memory is allowed here!

Koncopd: Ok, got it.

falexwolf: I think it'd be nice for a few applications if we allowed backed objects, but then typing is going to be a mess again. So, just in memory is easier for now.

PR #1589, opened 2024-04-22
♻️ Better user feedback on folder-like artifacts

falexwolf:


lamindb/_artifact.py
```diff
@@ -935,41 +947,42 @@ def delete(
...
                 delete_in_storage = response == "y"
             else:
                 delete_in_storage = storage
+        if not storage:
+            logger.warning(
+                f"you will retain a dangling store here: {path}, not referenced via an artifact"
+            )
         # we don't yet have logic to bring back the deleted metadata record
         # in case storage deletion fails - this is important for ACID down the road
-        if delete_in_storage is True:
```
falexwolf, line 969: FYI: @sunnyosun - this is an anti-pattern!

PR #1600, opened 2024-04-23
✨ Allow passing path to .from_anndata

sunnyosun:


lamindb/_annotate.py
```diff
@@ -349,14 +350,27 @@ class AnnDataAnnotator(DataFrameAnnotator):
...
         verbosity: str = "hint",
         organism: str | None = None,
     ) -> None:
-        self._adata = adata
+        from lamindb.core.storage import UPath
+
+        from ._artifact import data_is_anndata
+
+        if not data_is_anndata(data):
+            raise TypeError
```
falexwolf, line 365: Can you add some user feedback here?

lamindb/_annotate.py
```diff
@@ -349,14 +350,27 @@ class AnnDataAnnotator(DataFrameAnnotator):
...
+        if isinstance(data, ad.AnnData):
+            self._adata = data
+        else:
+            from lamindb.core.storage._backed_access import backed_access
+
+            self._adata = backed_access(UPath(data))
```
falexwolf, line 371: Can you use `setup.core.upath.create_path()` rather than UPath, please? Only create_path incorporates access management.

lamindb/_annotate.py
```diff
@@ -920,10 +935,11 @@ def validate_categories_in_df(
 
 def save_artifact(
     data: pd.DataFrame | ad.AnnData | MuData,
-    description: str,
     fields: dict[str, FieldAttr] | dict[str, dict[str, FieldAttr]],
     columns_field: FieldAttr | dict[str, FieldAttr],
+    description: str | None = None,
     organism: str | None = None,
+    adata: ad.AnnData | None = None,
```
falexwolf, line 942: Why can we pass `adata` and `data` here? I think both should be removed for clarity.

lamindb/_artifact.py
```diff
@@ -664,7 +664,7 @@ def from_df(
 @doc_args(Artifact.from_anndata.__doc__)
 def from_anndata(
```
falexwolf, line 665: Can you add the data_is_anndata validation, here, too?

falexwolf (approved):
Can you please add a little test?

lamindb/_annotate.py
sunnyosun: This is not user facing, and I need to pass the path and the backed object in order to register artifact and feature sets

PR #1632, opened 2024-05-07
Add links to tutorials of bionty registries

Zethson:


Zethson:
Is there a reason why the CI hits the timeout?

PR #1639, opened 2024-05-08
🚸 No need to pass organism if validating on ids

sunnyosun:


tests/test_from_values.py
```diff
@@ -49,20 +49,20 @@ def test_from_values_organism():
     from lnschema_bionty import Gene, settings
 
     settings._organism = None
-
     with pytest.raises(AssertionError):
         Gene.from_values(["ABC1"], Gene.symbol)
+    # no organism is needed if the values are ensembl gene ids
```
falexwolf, line 54: Thank you!

PR #1649, opened 2024-05-10
♻️ Do not attempt to manage public storage locations

falexwolf:


falexwolf:
This got already pushed to main in meanwhile via another PR.

PR #1663, opened 2024-05-17
♻️ Rename `.type` to `.dtype` in `Feature` and `FeatureSet` & refactor feature types

falexwolf:
https://laminlabs.slack.com/archives/C046F63LJJ2/p1715940665403049

### DIFF
--- a/lamindb/_artifact.py
+++ b/lamindb/_artifact.py
@@ -216,8 +216,7 @@ def get_stat_or_artifact(
             try:
                 n_workers = len(psutil.Process().cpu_affinity())
             except AttributeError:
-                # Fallback method
-                n_workers = os.cpu_count()
+                n_workers = psutil.cpu_count()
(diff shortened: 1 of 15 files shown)


lamindb/_artifact.py
```diff
@@ -212,7 +213,11 @@ def hash_size(file):
                 file_size = file.stat().st_size
                 return hash_file(file, file_size)[0], file_size
 
-            n_workers = len(psutil.Process().cpu_affinity())
+            try:
+                n_workers = len(psutil.Process().cpu_affinity())
+            except AttributeError:
+                # Fallback method
+                n_workers = os.cpu_count()
```
Koncopd, line 220: probably better to use `psutil.cpu_count()` here instead of `os.cpu_count()`

falexwolf: Ok!

Issue #1614, opened 2024-05-03
Increase Annotate coverage


Zethson:
https://laminlabs.slack.com/archives/C04FPE8V01W/p1713907661078619

https://app.codecov.io/gh/laminlabs/lamindb/blob/main/lamindb%2F_annotate.py

PR #1665, opened 2024-05-18
:sparkles: Improve coverage of annotate

Zethson:
Fixes https://github.com/laminlabs/lamindb/issues/1614


tests/test_annotate.py
```diff
@@ -74,11 +106,70 @@ def test_annotator(df, categoricals):
...
+
+
+def test_unvalidated_data_object(df, categoricals):
+    annotate = ln.Annotate.from_df(df, categoricals=categoricals)
+    with pytest.raises(ValidationError) as error:
+        annotate.save_artifact()
+    assert "Data object is not validated" in str(error.value)
+
+
+def test_clean_up_failed_runs(mock_transform):
```
Zethson, line 146: Don't need the fixture here anymore @Zethson

falexwolf:
The 5 failures in your test_artifact are all referring to this single line of code:

https://github.com/laminlabs/lamindb/blob/adf1e8393a7470cfd359623754f25bec870056fd/tests/test_artifact.py#L339-L348

As you see from the error message, this triggers only for non-registered storage and only if ln.track() isn't active

I don't think you registered `./non_registered_storage/` so maybe you're calling ln.track() without clearing it?

falexwolf:
Shall we merge?

Zethson:
I won't be able to revisit this PR today/tomorrow so we can merge and I might follow up.

May 23, 2024 10:33:33 AM Alex Wolf ***@***.***>:

PR #1674, opened 2024-05-22
🚸 `ln.track()` returns `run`, better duplicate detection and search, prettier `.describe()`, and more

falexwolf:
- `ln.track()` returns `run`
- better duplicate detection and search
- prettier `.describe()`
- feature values decoupled from feature sets
- more interactivity in `lamin save`
- `create` flag in `.from_values()`
- reorder fields in dataframe & record representations
- auto-detect feature types in dict-style assignment & raise better validation errors

### DIFF
--- a/docs/introduction.ipynb
+++ b/docs/introduction.ipynb
@@ -702,6 +702,7 @@
    "outputs": [],
    "source": [
     "collection = ln.Collection([artifact, artifact2], name=\"my RNA-seq collection\", version=\"1\")\n",
+    "collection.save()\n",
     "collection.describe()\n",
     "collection.view_lineage()"
    ]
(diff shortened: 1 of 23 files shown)


falexwolf:
I keep iterating through different versions

This is where I am right now

image

And here with types

image

Zethson:
@falexwolf cool! Feels much more reduced.

I wonder whether should keep the `'` given that in the case of `experimental_factors` they are super distracting because some 10x assays also contain `'` in their names. I know that we want to keep colors minimal, but maybe just highlighting the `,` would serve as better separators than the much more cluttering `'`?

falexwolf:
Yes, legitimate! Let me make another big change, and then we'll discuss polishing!

PR #1679, opened 2024-05-24
🔊 Better catching exceptions for graphviz install

sunnyosun:


felix0097:
I've pushed another fix @sunnyosun. The problem was that the `display()` call does not raise an exception but rather just prints the exception to the console. 

The problem gets fixed if I call `u._repr_mimebundle_()` manually: `display(u._repr_mimebundle_(), raw=True)`. Then the `ExecutableNotFound` exception gets raised + handled properly. 

I've checked it on my laptop, now everything works as expected. What do you think?

Zethson:
@felix0097 we should avoid calling underscore methods (private API that can change at any time) from external dependencies if possible. Is there an alternative?

sunnyosun:
Thank you @felix0097! Let's merge this for now and keep an eye on API change.

PR #1686, opened 2024-06-01
:sparkles: Enable Ruff security rules (bandit)

Zethson:
Enabling security rules in Ruff:

```
lamindb/_annotate.py:962:9: S101 Use of `assert` detected
lamindb/_annotate.py:1168:5: S101 Use of `assert` detected
lamindb/_artifact.py:971:9: S101 Use of `assert` detected
lamindb/_can_validate.py:226:9: S110 `try`-`except`-`pass` detected, consider logging the exception
lamindb/_collection.py:103:9: S101 Use of `assert` detected
lamindb/_feature.py:142:5: S101 Use of `assert` detected
lamindb/_finish.py:106:13: S602 `subprocess` call with `shell=True` identified, security issue
lamindb/_finish.py:127:13: S602 `subprocess` call with `shell=True` identified, security issue
lamindb/core/_feature_manager.py:379:5: S101 Use of `assert` detected
lamindb/core/_feature_manager.py:421:13: S101 Use of `assert` detected
lamindb/core/_feature_manager.py:432:17: S101 Use of `assert` detected
lamindb/core/_feature_manager.py:434:17: S101 Use of `assert` detected
lamindb/core/_feature_manager.py:538:9: S101 Use of `assert` detected
lamindb/core/_feature_manager.py:541:9: S101 Use of `assert` detected
lamindb/core/_feature_manager.py:569:9: S101 Use of `assert` detected
lamindb/core/_feature_manager.py:599:9: S101 Use of `assert` detected
lamindb/core/_label_manager.py:69:9: S112 `try`-`except`-`continue` detected, consider logging the exception
lamindb/core/_label_manager.py:105:13: S110 `try`-`except`-`pass` detected, consider logging the exception
lamindb/core/_mapped_collection.py:110:9: S101 Use of `assert` detected
lamindb/core/_run_context.py:47:24: S324 Probable use of insecure hash functions in `hashlib`: `md5`
lamindb/core/_run_context.py:117:16: S113 Probable use of requests call without timeout
lamindb/core/_settings.py:122:9: S101 Use of `assert` detected
lamindb/core/_sync_git.py:28:9: S602 `subprocess` call with `shell=True` identified, security issue
lamindb/core/_sync_git.py:39:9: S607 Starting a process with a partial executable path
lamindb/core/_sync_git.py:40:9: S602 `subprocess` call with `shell=True` seems safe, but may be changed in the future; consider rewriting without `shell`
lamindb/core/_sync_git.py:61:9: S602 `subprocess` call with `shell=True` identified, security issue
lamindb/core/_sync_git.py:71:9: S101 Use of `assert` detected
lamindb/core/_sync_git.py:85:13: S607 Starting a process with a partial executable path
lamindb/core/_sync_git.py:86:13: S602 `subprocess` call with `shell=True` seems safe, but may be changed in the future; consider rewriting without `shell`
lamindb/core/_sync_git.py:96:9: S602 `subprocess` call with `shell=True` identified, security issue
lamindb/core/_track_environment.py:18:66: S602 `subprocess` call with `shell=True` identified, security issue
lamindb/core/storage/_backed_access.py:210:13: S110 `try`-`except`-`pass` detected, consider logging the exception
lamindb/core/storage/_backed_access.py:308:17: S110 `try`-`except`-`pass` detected, consider logging the exception
lamindb/core/storage/paths.py:50:5: S101 Use of `assert` detected
lamindb/integrations/_vitessce.py:42:13: S101 Use of `assert` detected
tests/conftest.py:36:9: S607 Starting a process with a partial executable path
tests/conftest.py:36:51: S602 `subprocess` call with `shell=True` seems safe, but may be changed in the future; consider rewriting without `shell`
tests/test_run_context.py:107:9: S602 `subprocess` call with `shell=True` identified, security issue
Found 38 errors.
```

Just a show off for now! We can discuss later


falexwolf:
Great! Let's discuss this tomorrow!

falexwolf:
How does bandit compare to GitHub's CodeQL (see one of the GA jobs).

Zethson:
> How does bandit compare to GitHub's CodeQL (see one of the GA jobs)?

I don't really see any or a lot of overlap. CodeQL seems to be pretty superficial for Python and only checks the most obvious issues. The advantages of CodeQL are that we can also use it for other languages and can use it's query language if we ever need to but I don't see a use-case for us.

Zethson:
(coverage may go down because we have a couple of untested exceptions now). I can add coverage for these in a followup PR quickly tomorrow.

falexwolf:
> I don't really see any or a lot of overlap. CodeQL seems to be pretty superficial for Python and only checks the most obvious issues. The advantages of CodeQL are that we can also use it for other languages and can use it's query language if we ever need to but I don't see a use-case for us.

So, the overlap is so small because CodeQL doesn't cover as many security issues as bandit for Python? The nature of the tools is the same in that they're both static code analyzers that detect security issues, correct?

lamindb/core/_settings.py
```diff
@@ -108,7 +108,8 @@ def sync_git_repo(self, value) -> None:
         For example: `ln.sync_git_repo = https://github.com/laminlabs/redun-lamin`
         """
         self._sync_git_repo = sanitize_git_repo_url(value)
-        assert self._sync_git_repo.startswith("https://")
+        if not self._sync_git_repo.startswith("https://"):
+            raise ValueError("Github repository URL must start with 'https://'.")
```
falexwolf, line 112: This should be "git repository", because GitLab and others will work, too.

lamindb/core/_sync_git.py
```diff
@@ -24,8 +24,7 @@ def get_git_repo_from_remote() -> Path:
         f"running outside of synched git repo, cloning {repo_url} into {repo_dir}"
     )
     result = subprocess.run(
-        f"git clone --depth 10 {repo_url}.git",
-        shell=True,
+        ["git", "clone", "--depth", "10", f"{repo_url}.git"],
```
falexwolf, line 27: I'm pretty sure the `shell=True` is there for a purpose. 🤔 

I have to admit that I never attempted to understand when it's necessary and when not; but I remember that some things didn't work without it. I'd be astonished if this works without it (meaning I added it "just for fun").

Is it possible that the Mac environment is at fault and that's why this might not be caught by CI? I will see if I can still locally run the tests for this.

lamindb/core/_sync_git.py
```diff
@@ -24,8 +24,7 @@ def get_git_repo_from_remote() -> Path:
         f"running outside of synched git repo, cloning {repo_url} into {repo_dir}"
     )
     result = subprocess.run(
-        f"git clone --depth 10 {repo_url}.git",
-        shell=True,
```
falexwolf, line 28: I'm pretty sure the shell=True is there for a purpose. 🤔

I have to admit that I never attempted to understand when it's necessary and when not; but I remember that some things didn't work without it. I'd be astonished if this works without it (meaning I added it "just for fun").

Is it possible that the Mac environment is at fault and that's why this might not be caught by CI? I will see if I can still locally run the tests for this.

lamindb/core/_sync_git.py
```diff
@@ -68,9 +65,8 @@ def get_git_commit_hash(blob_hash: str, repo_dir: Path | None = None) -> str | N
     if commit_hash == "" or result.returncode == 1:
         return None
     else:
-        assert (
-            len(commit_hash) == 40
-        ), f"commit hash |{commit_hash}| is not 40 characters long"
+        if not len(commit_hash) == 40:
```
falexwolf, line 68: This will be hard to test. 😅 

Maybe we can remove it entirely. 🤔 I think I wanted to assert that the git query was successful and didn't contain an error message. Checking for the length of the parsed output was the proxy.

When this goes wrong, the important thing is that `commit_hash` is printed so that we know what error was raised by git.

lamindb/core/storage/_backed_access.py
```diff
@@ -208,7 +208,9 @@ def safer_read_partial(elem, indices):
                 ds = CSRDataset(elem)
                 result = _subset_sparse(ds, indices)
             except Exception:
-                pass
+                logger.debug(
+                    "Encountered an exception while attempting to subset a sparse dataset by indices."
```
falexwolf, line 212: If we log something, here, we should also log `e` where `e` is the raised instance of `Exception`.

lamindb/integrations/_vitessce.py
```diff
@@ -39,9 +39,12 @@ def save_vitessce_config(vitessce_config: VitessceConfig, description: str) -> A
             if "url" not in file:
                 raise ValueError("Each file must have a 'url' key.")
             filename = file["url"].split("/")[-1]
-            assert filename.endswith(
+            if not filename.endswith(
```
falexwolf, line 42: Great that we'll get a test for this!

tests/conftest.py
```diff
@@ -33,7 +33,7 @@ def pytest_sessionfinish(session: pytest.Session):
         ln.UPath("s3://lamindb-test/.lamindb").rmdir()
     ln.setup.delete("lamindb-unit-tests", force=True)
     # shutil.rmtree("./outside_storage")
-    run("docker stop pgtest && docker rm pgtest", shell=True, stdout=DEVNULL)
+    run("docker stop pgtest && docker rm pgtest", shell=True, stdout=DEVNULL)  # noqa: S602
```
falexwolf, line 36: Is S602 saying that `shell=True` is OK?

falexwolf (approved):
This is overall great!

The only "bad thing" concerns the few cases where exceptions now need to be triggered in tests, but it's very hard to.

I used `assert` in those instances because I consciously wanted to avoid having to test these exceptions.

Because assert is only used for consistency checks that are meant to provide feedback to users/devs in some edge cases, I thought it'd be OK security-wise.

If somebody attempted to bypass these assert statements, it'd be OK: the "attacker" would get a more cryptic error behavior downstream.

Hence: In those hard-to-test cases, I'd stick with the assert rather than introducing an additional line of code and an additional test with essentially no value to users & devs. I consider maintaining that additional line and test more harmful that beneficial; in particular in a phase of this project where the code overall can hopefully still much consolidated (I mean, the code works, but it needs to be refactored to become more maintainable).

In cases where there is value to the user (say the allowed path suffixes for the vitessce export), there should never be an assert.

lamindb/core/_sync_git.py
Zethson: For `shell=False` to work, we need to pass the command as a list. The subprocess module then knows what the command is and what the arguments are.

But ya, we can unpack and split them.

lamindb/core/_sync_git.py
Zethson: I don't really see why `shell=True` would be necessary here although I also had weird quirks with it in the past. Generally, `shell=True` is necessary when you're using shell features such as wild cards, piping, and other things.

falexwolf: Test works! No worries about the arg list. Can stay as is.

lamindb/integrations/_vitessce.py
Zethson: The whole function is only tested via notebooks actually at the moment.

tests/conftest.py
Zethson: Yes! https://docs.astral.sh/ruff/rules/subprocess-popen-with-shell-equals-true/

1. Since this is only a test, we're fine with having `shell=True` being used.
2. This uses `&&` which is the shell interprets so we need `shell=True` (excluding the possibility to rewrite - wasted effort here for a test)

Zethson:
> Thanks for clarifying! For me to understand: The overlap is so small because CodeQL doesn't cover as many security issues as bandit for Python? The nature of the tools is the same in that they're both static code analyzers that detect security issues, correct?

Correct. CodeQL can also test for unreachable code and other things but we have all of that covered with other tools already. I am currently not impressed by CodeQL but enabling it doesn't hurt.

> Hence: In those hard-to-test cases, I'd stick with the assert rather than introducing an additional line of code and an additional test with essentially no value to users & devs

Yeah that's fair. However, either we:
1. Go back to asserts and tell Bandit to ignore them.
2. Keep the exceptions and tell codecov to ignore them (if it works cough). Ignoring hard to test things with codecov is totally fine.

I vote for option 2.

Zethson:
> Is it possible that the Mac environment is at fault and that's why this might not be caught by CI? I will see if I can still locally run the tests for this.

Mac uses zsh instead of bash by default, right? But that shouldn't change the behavior. So either the test is bad, or it does work without `shell=True` because I don't see it using any shell specific features.

falexwolf:
> I vote for option 2.

I think option 1 with a `noqa` and a brief comment next to it would be simpler because we wouldn't need to mess with codecov and we don't have these verbose sections in the code.

I care much more about simplicity than about the exact codecov coverage number. If there is no or minimal value add through additional complexity, I'd always go with what's simpler.

In this case, I'd say there is zero value add in a few instances where assert was replaced. Introducing complexity both within the code and the coverage CI doesn't seem to merit it. We also need to keep in mind that whatever convention is introduced in this PR will be mandated for future contributors. Explaining people how to "not count a specific except block in coverage" seems like a lot of luggage. 

It's also error prone because some people might take it as an excuse to not test triggering the exception. Currently, we have this beautiful & simple rule that _every exception needs to be triggered in tests_. There can't be a single try except block that's not triggered.

If what you suggest means starting to distinguish between "meaningful exceptions" and "exceptions we add to make bandit happy" my reaction is: ohhhhh, let's please not do this. Nobody will get it.

So, unless I miss something, I vote for option 1. It's just a tiny bit of work to revert some of these exceptions back to asserts and add the `noqa` and the comment, and we'll not have headache with this topic going forward.

Zethson:
> It's also error prone because some people might take it as an excuse to not test triggering the exception. Currently, we have this beautiful & simple rule that every exception needs to be triggered in tests. There can't be a single try except block that's not triggered.

I didn't really agree with all of your arguments because they can be made both ways, but this one is good. I'll revert to asserts where reasonable and then merge.

falexwolf:
> I didn't really agree with all of your arguments because they can be made both ways, but this one is good. I'll revert to asserts where reasonable and then merge.

How can any of the other arguments made both ways? 🤔 

> Correct. CodeQL can also test for unreachable code and other things but we have all of that covered with other tools already. I am currently not impressed by CodeQL but enabling it doesn't hurt.

I think I somewhat get this. Thanks!

Zethson:
> Explaining people how to "not count a specific except block in coverage" seems like a lot of luggage.

Now we have to explain how to "circumvent the Bandit rule". We could ignore it in the config, but I actually think that it can be useful if developers think twice about whether this should be an internal `assert` or a user facing exception.

falexwolf:
> We could ignore it in the config, but I actually think that it can be useful if developers think twice about whether this should be an internal assert or a user facing exception.

I fully agree with you. I think each assert statement should have a `noqa: ...` and a comment.

lamindb/_can_validate.py
```diff
@@ -223,8 +223,10 @@ def set_abbr(self, value: str):
     else:
         try:
             self.add_synonym(value, save=False)
-        except Exception:  # pragma: no cover
-            pass
+        except Exception as e:  # pragma: no cover
```
falexwolf, line 226: @sunnyosun - why do we need this? This seems like it should error.

lamindb/_finish.py
```diff
@@ -41,7 +41,10 @@ def finish() -> None:
     if run_context.run is None:
         raise TrackNotCalled("Please run `ln.track()` before `ln.finish()`")
     if run_context.path is None:
-        assert run_context.transform.type not in {"script", "notebook"}
+        if run_context.transform.type in {"script", "notebook"}:
+            raise ValueError(
+                f"Transform type must be 'script' or 'notebook' but is {run_context.transform.type}."
+            )
```
falexwolf, line 47: That would have merited a test, but ok.

lamindb/core/storage/_backed_access.py
```diff
@@ -209,8 +209,10 @@ def safer_read_partial(elem, indices):
             try:
                 ds = CSRDataset(elem)
                 result = _subset_sparse(ds, indices)
-            except Exception:
-                pass
+            except Exception as e:
```
falexwolf, line 212: Is this the only way this can be done, @Koncopd?

lamindb/core/storage/_backed_access.py
```diff
@@ -307,8 +309,10 @@ def safer_read_partial(elem, indices):  # noqa
                 try:
                     ds = CSRDataset(elem)
                     return _subset_sparse(ds, indices)
-                except Exception:
-                    pass
+                except Exception as e:
```
falexwolf, line 312: Same here, @Koncopd.

lamindb/core/_sync_git.py
```diff
@@ -68,9 +65,6 @@ def get_git_commit_hash(blob_hash: str, repo_dir: Path | None = None) -> str | N
     if commit_hash == "" or result.returncode == 1:
         return None
     else:
-        assert (
```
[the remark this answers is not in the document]
Zethson: I thought you said you tested it and it worked without it?

lamindb/_finish.py
Zethson: Will add one!

lamindb/core/_sync_git.py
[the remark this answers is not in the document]
Zethson: Ohh sorry, I misunderstood you.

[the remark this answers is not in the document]
Zethson: I can restore it.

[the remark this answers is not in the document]
falexwolf: Reverted: https://github.com/laminlabs/lamindb/commit/163012a9fa0706e8262929375734e6ee1a2e27ff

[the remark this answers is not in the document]
falexwolf: FYI @Zethson

PR #1699, opened 2024-06-13
♻️ Refactor after upath upgrade

Koncopd:


falexwolf:
This is great, @Koncopd!

PR #1700, opened 2024-06-13
🐛 Fix check_path_is_child_of_root for upaths with different storage_options

Koncopd:
subj


lamindb/_artifact.py
```diff
@@ -278,8 +278,10 @@ def check_path_in_existing_storage(
 
 
 def check_path_is_child_of_root(path: Path | UPath, root: Path | UPath | None) -> bool:
-    path = UPath(path)
-    root = UPath(root)
+    # str is needed to eliminate UPath storage_options
+    # from the equality checks below
```
falexwolf, line 282: Thank you, @Koncopd!

PR #1729, opened 2024-07-15
Add codeql config

Zethson:


Zethson:
Ah, we have the default setup already. So it's fine. It doesn't overlap with Bandit at all

falexwolf:
Got it!

PR #1735, opened 2024-07-16
🐛 Move `mudata` import in TYPE_CHECKING

fredericenard:


falexwolf:
What's the background of this PR, @sunnyosun & @fredericenard? Evidently, it breaks things. Was a dependency on mudata introduced?

Issue #1759, opened 2024-07-27
System error on re-running CellMarker.add_synonym


insavchuk:
<img width="1424" alt="Screenshot 2024-07-27 at 17 46 02" src="https://github.com/user-attachments/assets/6a961838-907b-4793-b479-cd7c59c892f3">

insavchuk:
Full traceback:

insavchuk:
error occurs once in ~10 re-runs of the same cell.

Zethson:
Dear @insavchuk,

thank you for the report! I have a kind request: Could you please always try to copy and paste code and error messages instead of posting screenshots? This makes it much easier for us to debug.

In a fresh instance,
```
import bionty as bt
bt.settings.organism = "human"

bt.CellMarker(name="TBX21").save()
bt.CellMarker.search("TBX21").first().add_synonyms("T-bet")
```

works for me.

Now the issue that you're running into is that you're trying to add a synonym to a record which already is a synonym of a record.

In other words,

```
bt.CellMarker.search(uid="ppzJGtvJuKYK).one()
```

already has the synonym `T-bet` as you can see in your screenshot. I have the impression that your search returns a different record than the one that already has the synonym. Then you are trying to add the synonym to a different record but it's already associated with one.

Could you please ensure that:

```
bt.CellMarker.search("TBX21", field="name").df()
```

only has a single hit?

insavchuk:
Hi Lukas, you had the right suspicion here - I realized that prior to synonym lookup, I had created a new entry for TBX21, and then added a synonym "T-bet" to that. Thanks!
```
bt.CellMarker.search("T-bet").df()
```
<div>
<style scoped>
    .dataframe tbody tr th:only-of-type {
        vertical-align: middle;
    }

    .dataframe tbody tr th {
        vertical-align: top;
    }

    .dataframe thead th {
        text-align: right;
    }
</style>
<table border="1" class="dataframe">
  <thead>
    <tr style="text-align: right;">
      <th></th>
      <th>uid</th>
      <th>name</th>
      <th>synonyms</th>
      <th>gene_symbol</th>
      <th>ncbi_gene_id</th>
      <th>uniprotkb_id</th>
      <th>organism_id</th>
      <th>public_source_id</th>
      <th>run_id</th>
      <th>created_by_id</th>
      <th>updated_at</th>
    </tr>
    <tr>
      <th>id</th>
      <th></th>
      <th></th>
      <th></th>
      <th></th>
[...]

PR #1764, opened 2024-07-29
✨ Better error message for synonym duplications

Zethson:
Fixes https://github.com/laminlabs/lamindb/issues/1759

We are now using `ValueError` instead of `SystemExit` and `AssertionError` because we are primarily performing input validation. `AssertionError` is more for internal behavior.

I decided to keep the error logging so that it's clear why we're displaying records.


Zethson:
I can't get the CI to pass even after several retries:

```
ProgrammingError: relation "lnschema_bionty_organism" does not exist
LINE 1: ...ion", "lnschema_bionty_organism"."source_id" FROM "lnschema_...
```

falexwolf:
Makes sense!

PR #1783, opened 2024-08-05
✨ Improve bio-registry tutorial wording

Zethson:


falexwolf:
Random comment, because I see it a lot: @Zethson, this is a docs improvement (📝) not a new feature (✨) -- I guess you really like the ✨ emoji 😆

Zethson:
Sorry, it's just my default. My brain is autopiloting sometimes ^_^

falexwolf:
And your first edit is another instance of the classic Lukas-Alex stylistic wording clash.

This has been crafted over many iterations:

image

I think it's the most natural way to stress the extension of public ontologies if one also wants to mention plug-in bionty:

> With plug-in {mod}`bionty`, it's easy to import records from public biological ontologies.

You seem to think that this is better:

> We provide the plugin {mod}`bionty`, which makes importing records from public biological ontologies easy.

- It's quite a bit longer and will wrap on many screens.
- It introduces a "We", which we avoid everywhere in the docs. There is no "We". This is not a paper but a product documentation.
- I don't think people care that bionty is "provided" by somebody. It's a complicated phrase. bionty is "just there". Imagine how ridiculous it'd sound if "Apple provided the iPhone". 😄 
- You introduce a relative clause with "which", _which_ makes the sentence more complicated without adding anything.
- There is a dash in "plug-in" missing. I know that its use is somewhat declining, but Grammarly, Claude, and other sources still auto-fix "plugin" to "plug-in". Doesn't your spell checker do this, too?

I might have either gone through too much of YC or generally been criticized for being too wordy too often in my life, but I don't get why you like what I think of as "academic paper-writing style" here in the docs.

Instead of merely rewording a simple sentence in a - to me - _questionable way_, I'd have wondered whether "importing from public ontologies" is the only thing we want to mention in the top-most summary. We had "manage biological registries" before, which isn't great either, as it's not very specific.

As `BioRecord` comes to down to a exposing the same registry functionality as everywhere else, it might in fact be enough to just focus on the "importing aspect".

New suggestion:

> This guide shows how to manage metadata for basic biological entities based on plug-in {mod}`bionty`.

And then maybe somewhere in the introduction:

> In the previous [guide](docs:public-ontologies), we already saw how to access public ontologies. But most of the time, the metadata they provide isn't enough. For instance, you might want to add new cell types or use multiple versions of a gene ontology.

falexwolf:
The `s` was fixed, but the `an` wasn't:
image

Zethson:
Sorry, it shall be changed (back) immediately.

falexwolf:
Wanna give it another shot? 😄 

> Sorry, it shall be changed (back) immediately.

I wouldn't change bit back. I think it's great that you're consolidating docs. I just don't like it if language becomes more complicated without conveying more information.

Zethson:
> There is a dash in "plug-in" missing. I know that its use is somewhat declining, but Grammarly, Claude, and other sources still auto-fix "plugin" to "plug-in". Doesn't your spell checker do this, too?

I looked it up actually. I had the impression that in computational context, most people used plugin whereas plug-in was mostly in the context of things like plug-in hybrid. Anyways, plug-in is 100% older and correct so we shall use it

PR #1795, opened 2024-08-06
📝 Improve curate docs

### LINKED ISSUE #1793: Curate is badly documented
https://docs.lamin.ai/lamindb.curate#lamindb.Curate

Just says "Annotation flow". We need to better document what it checks on initialize and how it should be used with `validate`.

Zethson:
1. Fixes @falexwolf requests concerning bioregistry
2. Fixes #1793 

To be merged after https://github.com/laminlabs/lamindb/pull/1790 because it references not yet merged attributes.

### DIFF
--- a/lamindb/_curate.py
+++ b/lamindb/_curate.py
@@ -764,15 +764,15 @@ class Curate:
     Choose the method corresponding to the object type you want to curate.
     During object creation, any passed categoricals found in the object will be saved.
 
-    2. Run :meth:`~lamindb.DataFrameCurator.validate` to check the data against the defined criteria.
+    2. Run :meth:`~lamindb.core.DataFrameCurator.validate` to check the data against the defined criteria.
     This method identifies which values are already validated (exist in our registries)
     and which are new or potentially problematic.
 
     3. Determine how to handle validated and unvalidated values:
 
-    - Validated values not yet in the registry can be automatically registered using :meth:`~lamindb.DataFrameCurator.add_validated_from`.
-    - Valid and new values can be registered using :meth:`~lamindb.DataFrameCurator.add_new_from`.
-    - All unvalidated values can be accessed using :meth:`~lamindb.DataFrameCurator.non_validated` and subsequently removed from the object at hand.
+    - Validated values not yet in the registry can be automatically registered using :meth:`~lamindb.core.DataFrameCurator.add_validated_from`.
+    - Valid and new values can be registered using :meth:`~lamindb.core.DataFrameCurator.add_new_from`.
+    - All unvalidated values can be accessed using :meth:`~lamindb.core.DataFrameCurator.non_validated` and subsequently removed from the object at hand.
     """
 
     @classmethod
--- a/lamindb/_curate.py
+++ b/lamindb/_curate.py
@@ -764,7 +764,7 @@ class Curate:
     Choose the method corresponding to the object type you want to curate.
     During object creation, any passed categoricals found in the object will be saved.
 
-    2. Run :meth:`~lamindb.Curate.validate` to check the data against the defined criteria.
+    2. Run `validate` to check the data against the defined criteria.
     This method identifies which values are already validated (exist in our registries)
     and which are new or potentially problematic.
 
--- a/lamindb/_curate.py
+++ b/lamindb/_curate.py
@@ -754,14 +754,14 @@ class Curate:
     to facilitate data integration, interpretation and analysis.
 
     The curation flow has several steps:
-    1. Create an :class:`DataFrameCurator` (:meth:`~lamindb.Curate.from_df`), :class:`AnnDataCurator` (:meth:`~lamindb.Curate.from_anndata`),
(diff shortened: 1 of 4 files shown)


lamindb/_curate.py
```diff
@@ -737,7 +748,22 @@ def save_artifact(self, description: str | None = None, **kwargs) -> Artifact:
...
 
 class Curate:
-    """Annotation flow."""
+    """Curation flow.
+
+    Data annotation entails accurately labeling biological datasets with relevant metadata and functional information
+    to facilitate data integration, interpretation and analysis.
+
+    The curation flow has several steps:
+    1. Create an :class:`DataFrameCurator` (:meth:`~lamindb.Curate.from_df`), :class:`AnnDataCurator` (:meth:`~lamindb.Curate.from_anndata`),
```
falexwolf, line 757: This line is intimidating! What about just listing the `from_` constructors and let users explore return type annotations themselves?

Zethson: Yeah I was 50/50 here while writing this. It will look a tiny bit less scary because the `from_` constructors will not expand but I'll consider removing the types.

lamindb/_curate.py
```diff
@@ -737,7 +748,22 @@ def save_artifact(self, description: str | None = None, **kwargs) -> Artifact:
...
+       :class:`MuDataCurator` (:meth:`~lamindb.Curate.from_mudata`) for the object type that you want to curate.
+       During the creation of the object, the passed categoricals will be validated to ensure that they are valid for curation.
```
falexwolf, line 759: This sounds like categories will be validated. I'm somewhat struggling with this choice myself but @sunnyosun insisted it's the right choice: feature names are validated during instantiation.

But this means that we should call out "feature names" here. In theory, to arrive at invariance under feature renaming, we'd like to be able to pass uids at some point. I fear we'd need a flag to indicate what to validated against..., which is gonna be ugly.

Can we re-visit the idea of validating feature identifiers for metadata features during `.validate` just like we validate feature identifiers for the data during `.validate()`?

lamindb/_curate.py
```diff
@@ -737,7 +748,22 @@ def save_artifact(self, description: str | None = None, **kwargs) -> Artifact:
...
+    2. Run :meth:`~lamindb.Curate.validate` which checks the data against the defined criteria.
```
falexwolf, line 760: Great!

lamindb/_curate.py
```diff
@@ -737,7 +748,22 @@ def save_artifact(self, description: str | None = None, **kwargs) -> Artifact:
...
+       It identifies which values are already validated (exist in our registries) and which are new or potentially problematic.
```
falexwolf, line 761: Great!

lamindb/_curate.py
```diff
@@ -737,7 +748,22 @@ def save_artifact(self, description: str | None = None, **kwargs) -> Artifact:
...
+    3. Determine how to handle validated and unvalidated values.
+       Against a public ontology validated values can automatically be registered using :meth:`~lamindb.Curate.add_validated_from`.
+       Valid and new values may be registered using :meth:`~lamindb.Curate.add_new_from`.
```
falexwolf, line 764: I can understand this, I just wouldn't say "may" but "can".

lamindb/_curate.py
Zethson: Discussed with @sunnyosun today. Quoting her:
> columns need to be validated upon init because they are the keys in the categories, so it’d better to error already so people can fix passing categories

lamindb/_curate.py
```diff
@@ -737,7 +748,22 @@ def save_artifact(self, description: str | None = None, **kwargs) -> Artifact:
...
+       All unvalidated values can be accessed using :meth:`~lamindb.Curate.non_validated` and subsequently removed from the object at hand.
```
falexwolf, line 765: What's the overlap with the `InspectResult` class here? It somewhat reminds me of it.

falexwolf:
Thank you for the bio-registries fix, @Zethson! 😉

lamindb/_curate.py
falexwolf: Yes, she told me the same thing when we discussed it 4-5 months ago. I'm still not 100% convinced.

falexwolf: If we're consistent, we should be able to pass `obs_columns=ln.Feature.name` or something like this. I think raising the error during `.validate()` would be more consistent with validating everything else but I might not see all arguments.

sunnyosun: adding obs_columns is straightforward: https://github.com/laminlabs/lamindb/pull/1790/commits/06637dd307968e3c8708c89390d01be9f7f94b71

Zethson: https://laminlabs.slack.com/archives/C04FPE8V01W/p1722956628159469

lamindb/_curate.py
Zethson: https://laminlabs.slack.com/archives/C04FPE8V01W/p1722959524854749

PR #1809, opened 2024-08-13
♻️ Update `MappedCollection` to work with latest version of `scDataLoader`

jkobject:
- checking variables are aligned across datasets
- checking that variables are sorted across datasets (useful in many cases but could be implemented as a side function)
- allowing for a weight scaler and outputting the different elements we are weighting on. (this is necessary for using a weighted random sampler that works on more than 2M elements)
- sorting the categories otherwise the encoder will change everytime it is used

### DIFF
--- a/lamindb/core/_mapped_collection.py
+++ b/lamindb/core/_mapped_collection.py
@@ -7,10 +7,9 @@ from typing import TYPE_CHECKING, Dict, List, Literal, Optional, Union
 
 import numpy as np
 import pandas as pd
-from lamin_utils import logger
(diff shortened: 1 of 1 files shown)


Koncopd:
Hi, @jkobject , thank you. Checking the requests. Are you fine if i rework the implementations?

jkobject:
yup :)

falexwolf:
Shall this be closed in favor of what Sergei did?

falexwolf:
I'll make a release just now!

Koncopd:
Yes, this can be closed.

PR #1812, opened 2024-08-13
✨ New features for MappedCollection

Koncopd:
https://github.com/laminlabs/lamindb/pull/1809
This add the features requested by @jkobject 

>checking variables are aligned across datasets

I have added a function `check_vars_non_aligned` that returns a list of objects' indices that are not aligned with the provided variables list.
```
mapped = MappedCollection(...)
mapped.check_vars_non_aligned(var_list) # returns a list of indices of objects with non-aligned variables
```
I haven't added this check by default because in general it is not needed when `join="inner"` or `"outer"` is used, but i think it is fine to have as a separate function.

>checking that variables are sorted across datasets

I have added `check_vars_sorted`
```
mapped.check_vars_sorted(ascending=True) # returns True if sorted in ascending order
mapped.check_vars_sorted(ascending=False) # descending
```

>allowing for a weight scaler and outputting the different elements we are weighting on

I see that `get_label_weights` didn't work properly before with several obs keys. Now i have fixed that and added proper tests. It concatenates labels from several keys and computes weight based on the number of these concatenated label categories.
I have added the scaler. It is `None` by default.
I have also added  `return_categories` argument, it is `False` by default, but if `True`, the function returns a dictionary with categories and their weight. It is more convenient in my opinion to have it `False` by default because than weight can be provided directly to a sampler.
```
mapped.get_label_weights(obs_keys=["key1", "key2"], scaler=10.) # returns weights for each observation
mapped.get_label_weights(obs_keys=["key1", "key2"], scaler=10., return_categories=True) # returns a dict with categories and their weights.
```
Also improved the docstring of this function.

>sorting the categories otherwise the encoder will change everytime it is used

Done.

### DIFF
--- a/lamindb/core/_mapped_collection.py
+++ b/lamindb/core/_mapped_collection.py
@@ -374,7 +374,7 @@ class MappedCollection:
             label = label.decode("utf-8")
         return label
 
-    def get_label_weights(self, obs_keys: str | list[str]):
+    def get_label_weights(self, obs_keys: str | list[str], scaler: float | None = None):
         """Get all weights for the given label keys."""
         if isinstance(obs_keys, str):
             obs_keys = [obs_keys]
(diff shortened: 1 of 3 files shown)


Koncopd:
@jkobject ok, great, merging then.

PR #1816, opened 2024-08-14
🔥 Remove parsing the nbproject_id backward compat

falexwolf:


falexwolf:
⚠️ At this point in time in the PR, all renaming is done, and I'm actually starting to introduce the new behavior.

PR #1828, opened 2024-08-21
📝 Simplify introduction

falexwolf:


falexwolf:
The diff here doesn't render 🤷 

image

PR #1840, opened 2024-08-23
🎨Fix Future Pandas rename categoricals warning

### LINKED ISSUE #1830: FutureWarning in can_validate
```
lamindb/lamindb/_can_validate.py:408: FutureWarning: The behavior of Series.replace (and DataFrame.replace) with CategoricalDtype is deprecated. In a future version, replace will only be used for cases that preserve the categories. To change the categories, use ser.cat.rename_categories instead.
  result = pd.Series(std_names_db).replace(std_names_bt_mapper).tolist()
```

Zethson:
Fixes #1830


Zethson:
> AttributeError: Can only use .cat accessor with a 'category' dtype

Sigh, do we really have to check dtype now?...

PR #1847, opened 2024-08-26
✅ Split unit tests into unit-core and unit-storage

Koncopd:


.github/workflows/build.yml
```diff
@@ -14,14 +14,15 @@ jobs:
...
+          - "unit-core"
+          - "unit-storage"
           - "tutorial"
           - "guide"
           - "biology"
           - "faq"
           - "storage"
           - "cli"
-    timeout-minutes: 11
+    timeout-minutes: 10
```
falexwolf, line 25: This should be back to 7 min.

falexwolf: We have to invest to get us strictly under 4 min in the upcoming days and weeks.

Koncopd:
Merging this?

PR #1854, opened 2024-08-27
:sparkles: Add issue templates

Zethson:
- Adds issue templates for bugs and features
- Adds buttons to report laminhub issues and to contact us

Should work and look similarly to https://github.com/laminlabs/laminhub-public/issues/new/choose

It's somewhat important to me to encourage people to report their used dependency versions because it's hella difficult to diagnose many issues without it.

Can't test it before merging, but it should be good.

### DIFF
--- a/.github/ISSUE_TEMPLATE/.config.yml
+++ b/.github/ISSUE_TEMPLATE/.config.yml
@@ -1,0 +1,8 @@
+blank_issues_enabled: false
+contact_links:
+  - name: LaminHub issues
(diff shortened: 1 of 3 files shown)


falexwolf (approved):
Awesome! :D Thanks, @Zethson!

Zethson:
OK so the buttons that refer to the other links don't seem to work for some reason here. Maybe we can only add one button? Let me try...

Zethson:
Whopsie, I had the wrong name for the config file. Works now.

PR #1885, opened 2024-09-02
⚡️ Improvements to `save_tiledbsoma_experiment`

Koncopd:
Read local adatas in backed mode for writing to a tiledbsoma store.

### DIFF
--- a/lamindb/core/storage/_tiledbsoma.py
+++ b/lamindb/core/storage/_tiledbsoma.py
@@ -2,11 +2,11 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING, Literal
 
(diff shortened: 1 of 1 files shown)


falexwolf:
Are we sure we want to buy in this complexity? If so, we need a test. I'd rather just keep it the way it was.

Koncopd:
This is the correct way, it is how it is done tin `tiledbsoma` also. And it is tested already by the existing tests.

Koncopd:
No complexity added from our side here, everything is already implemented in `tiledbsoma` itself, i mean it accepts backed anndatas without any problems.

falexwolf:
Ok!

PR #1891, opened 2024-09-04
♻️ Refactor passing an internal storage path

falexwolf:


lamindb/core/storage/_tiledbsoma.py
```diff
@@ -192,5 +188,10 @@ def save_tiledbsoma_experiment(
         )
 
     return Artifact(
-        storepath, key=key, description=description, run=run, revises=revises, _uid=_uid
+        storepath,
+        key=key,
+        description=description,
+        run=run,
+        revises=revises,
+        _is_internal=True,
```
falexwolf, line 196: FYI @Koncopd 

I'm no longer passing `_uid` here because I'm worried about inconsistencies between the uid that's encoded in the storepath and the passed variable.

Instead, I'm now parsing the `uid` from the storepath within the Artifact constructor.

The `_is_internal` variable wouldn't be needed and is just for raising clearer errors to users who should never pass a path or key that contains `.lamindb`.

Koncopd: I see, makes perfect sense.

Koncopd: Also a similar thing happens here https://github.com/laminlabs/lamindb/blob/4d6273cc5c9ea4ac5a6f7ae5c4e73ee26a5bc034/lamindb/_artifact.py#L949C44-L949C52
Does this also need an edit?

falexwolf: The link is broken but I guess you mean this:

https://github.com/laminlabs/lamindb/blob/4d6273cc5c9ea4ac5a6f7ae5c4e73ee26a5bc034/lamindb/_artifact.py#L949

This looks good because it doesn't pass a uid!

falexwolf: Oh, yes, it now raises this error because the constructor thinks it's coming from a user.

```
>               raise ValueError("Do not pass paths inside the `.lamindb` directory.")
E               ValueError: Do not pass paths inside the `.lamindb` directory.
```

So, I'll pass `_is_internal_call`.

Koncopd: The problem is that it can actually come from a user.

lamindb/_artifact.py
```diff
@@ -615,16 +617,29 @@ def __init__(artifact: Artifact, *args, **kwargs):
...
+        artifact._local_filepath = privates["local_filepath"]
+        artifact._cloud_filepath = privates["cloud_filepath"]
+        artifact._memory_rep = privates["memory_rep"]
+        artifact._to_store = not privates["check_path_in_storage"]
+
+    if is_automanaged_path and _is_internal_call:
+        kwargs["_key_is_virtual"] = True
+        if AUTO_KEY_PREFIX in kwargs["key"]:
+            kwargs["key"] = None
+        provisional_uid = artifact.path.stem
```
Koncopd, line 630: I don't think that works at this point.

Koncopd: I mean `artifact.path`.

lamindb/core/storage/_tiledbsoma.py
Koncopd: The path here can be both autogenerated and normal user defined.

falexwolf: I don't think it's a problem because `_is_internal_call` only tells the API that it's OK to pass a `.lamindb/` path. Because the `.open()` API doesn't allow to pass an external path from the user, we should be good.

image

lamindb/_artifact.py
falexwolf: Yes, I realized meanwhile...

This whole PR got much more complicated than I wanted 😞

PR #1898, opened 2024-09-05
noxfile: fix "install" session

ap--:
Hello Lamins,

When following the contribution guide, it is suggested to run `nox -s install`, which fails because the lamindb repo is not under `sub/lamindb`. This change aligns the install sessions behavior with the install_ci sessions behavior used in CI.

Cheers,
Andreas 😃


falexwolf:
Thank you! 🙇

Issue #1919, opened 2024-09-12
🐛 `n_observations` isn't written in `save_tiledbsoma_experiment`


falexwolf:
### Report

<img width="473" alt="image" src="https://github.com/user-attachments/assets/5981b670-ab30-476f-a579-493ef3f1ab6e">

falexwolf:
@Koncopd, this should be a simple bug fix, I hope.

When you make it, can you also add a consistency check in case adata-like artifacts are passed?
```
assert adata.shape[0] = adata_artifact.n_observations
```

PR #1930, opened 2024-09-17
✨ Add n_observations to tiledbsoma artifacts

Koncopd:
Fixes https://github.com/laminlabs/lamindb/issues/1919

### DIFF
--- a/lamindb/core/storage/_tiledbsoma.py
+++ b/lamindb/core/storage/_tiledbsoma.py
@@ -201,5 +201,6 @@ def save_tiledbsoma_experiment(
         _is_internal_call=True,
     )
     artifact.n_observations = n_observations
+    artifact._accessor = "TileDB-SOMA"
 
     return artifact.save()
--- a/lamindb/core/storage/_tiledbsoma.py
+++ b/lamindb/core/storage/_tiledbsoma.py
@@ -174,6 +174,12 @@ def save_tiledbsoma_experiment(
             context=ctx,
         )
 
+    if registration_mapping is not None:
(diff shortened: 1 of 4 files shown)


sunnyosun:
Could we also populate the `accessor` for these artifacts so it's not "unknown" here?
Screenshot 2024-09-17 at 11 49 25

Koncopd:
Yep

Koncopd:
How should i call it? `tiledbsoma`?

sunnyosun:
We currently have "AnnData", "DataFrame", "MuData", so maybe "TileDB-SOMA"?

Koncopd:
Yes, makes sense, added like this.

sunnyosun:
Ha, you did the same already! :D

lamindb/core/storage/_tiledbsoma.py
```diff
@@ -186,11 +192,15 @@ def save_tiledbsoma_experiment(
...
         storepath,
         key=key,
         description=description,
         run=run,
         revises=revises,
         _is_internal_call=True,
-    ).save()
+    )
+    artifact.n_observations = n_observations
+    artifact._accessor = "TileDB-SOMA"
```
falexwolf, line 204: Can we please keep it simple here and use `tiledbsoma` instead?

falexwolf: I'll make the change.

It's mostly for aesthetical reasons, but `TileDB-SOMA` also isn't valid Python code whereas `tiledbsoma` is.

PR #1945, opened 2024-09-18
🚸 Simplify CLI commands

falexwolf:
- Integrate `close` as `--unload` into `load`
- Integrate `logout` as `--logout` into `login`
- Integrate `schema view` as `info view` into `info`

Before | After
--- | ---
image | image

Before | After
--- | ---
image | image

In `lamin-cli`

- https://github.com/laminlabs/lamin-cli/pull/72

### DIFF
--- a/docs/signup-login.md
+++ b/docs/signup-login.md
@@ -32,5 +32,5 @@ lamin login <email> --key <API-key>
 Log out:
 
 ```
-lamin logout
+lamin login --logout
 ```
--- a/noxfile.py
+++ b/noxfile.py
@@ -209,9 +209,11 @@ def docs(session):
             names = name.split(" ")
             section = ""
             if len(names) != 1:
-                section = (
-                    "```\n\n" + "#" * len(names) + " " + " ".join(("lamin", *names[1:]))
-                )
+                if len(names) == 2:
+                    separator = "#" * len(names) + " " + " ".join(("lamin", *names[1:]))
+                else:
+                    separator = ""
+                section = "```\n\n" + separator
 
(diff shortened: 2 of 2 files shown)


Koncopd:
Hm, i think `lamin close` or `lamin logout` is much clearer than `lamin load --unload` / `lamin login --logout`.
Also if i want to logout / close, got to docs and now can't really find how to do this (i am looking at the screenshots after).
I think this change is super confusing.

Koncopd:
Also this requires much more edits
https://github.com/search?q=org%3Alaminlabs%20%22lamin%20close%22&type=code

falexwolf:
Ah, you're right. Making changes: https://github.com/laminlabs/lamindb-setup/commit/d2cfe8b9651ac860efc2173f2fc143fb99629186

falexwolf:
I think these are all. I don't find anymore hits across the core repositories.

falexwolf:
> Hm, i think lamin close or lamin logout is much clearer than lamin load --unload / lamin login --logout.

Yes, it's clearer but it bloats the command number so much that it becomes overwhelming.

> Also if i want to logout / close, got to docs and now can't really find how to do this (i am looking at the screenshots after).

I think the opposite is true. I think people (even Richard) aren't clear on what the heck "closing an instance" even is. I didn't like this for more than 2 years. Richard then suggested "unload" and I think that's much better but having its own command for what is a flag is also weird.

So, when I want to find in the docs "how the heck I'm undoing the loading thing" I'll just see it right away:
image

Same rationales for `logout`. Many products have `auth` instead of `login` which would allow to nest things but that's also complicated.

PR #1953, opened 2024-09-20
✨ Overhaul `save_vitessce_config()` to support multiple artifacts and non-`.zarr`

keller-mark:
## Support multiple artifacts and non-`.zarr`

If the `VitessceConfig` object references multiple artifacts, `save_vitessce_config` will now create a `Collection` and attach the config file to the collection. This makes the "Vitessce button" appear next to the collection rather than the artifact.

## Simplified interface

One can now directly pass `Artifact` objects to the `VitessceConfig` wrapper classes via

- https://github.com/vitessce/vitessce-python/pull/371 
- https://github.com/vitessce/vitessce-python/pull/373

This simplifies the interface for constructing `VitessceConfig` objects:

```diff
vit.AnnDataWrapper(
- adata_url=my_artifact.path.to_url(),
+ adata_artifact=my_artifact,
)
```

Also: The internals of `save_vitessce_config` are now much cleaner as there is no longer a need to do lookups based on urls which were previously needed as the flow was (`artifact -> url -> artifact`).


falexwolf:
Fantastic, Mark!

I also see that this is already released in: https://pypi.org/project/vitessce/3.4.0/

I'll merge it here and release it in lamindb.

keller-mark:
Note that this is a breaking change (will break existing notebooks that use save_vitessce_config if the config was using `adata_url` (would need to switch to `adata_artifact`). Is that ok? If not I can work on a fallback code path that would still do the artifact lookups as well

falexwolf:
Awesome, @keller-mark!

Can you help me understand whether this will break things for non-anndata-like datasets?

falexwolf:
Also, do you have an example for a `spatialdata` dataset that we could show in the guide?

https://docs.lamin.ai/vitessce (source: https://github.com/laminlabs/lamin-spatial/blob/main/docs/vitessce.ipynb)

I think they're almost exclusively working with `spatialdata` on the Hackathon.

falexwolf:
@zimea, if you call `ln.finish()`, we could take a look at the source code that you've used to ingest the Heise 2023 dataset 😄 

https://lamin.ai/scverse/spatialdata-db/transform/qQ5IQkXWOgt90000/

keller-mark:
> Can you help me understand whether this will break things for non-anndata-like datasets?

In https://github.com/vitessce/vitessce-python/pull/371/ I added support for:
- `AnnDataWrapper.adata_artifact`
- `ImageOmeZarrWrapper.img_artifact`
- `ImageOmeTiffWrapper.img_artifact`
- `ObsSegmentationsOmeZarrWrapper.img_artifact`
- `ObsSegmentationsOmeTiffWrapper.img_artifact`

In https://github.com/vitessce/vitessce-python/pull/373 (not yet merged) I added support for `SpatialDataWrapper.sdata_artifact` but some things depend on the JS PR https://github.com/vitessce/vitessce/pull/1849 (e.g., Parquet-formatted Shapes)

falexwolf:
@timtreis, Mark's changes here make working with the `VitessceConfig` object and lamindb more convenient and robust.

Do you happen to be either working with `AnnData`-like datasets or already working off of

- https://github.com/vitessce/vitessce-python/pull/373

during the Hackathon? Either way, we could merge and release this PR and you'd have an improvement.

FYI: @ilan-gold @Zethson

zimea:
@falexwolf done, but it was simply a POC that I can write to the spatialdata-db lamin instance and I planned to delete the artifacts and code.

timtreis:
Hey Alex, 

we won't be directly working with `AnnData`-like objects, but the `SpatialData` objects can contain `n` `AnnData`s. So we might have to perform hierarchical validation, but @Zethson is aware of that and is thinking hard from what I can see right now. 

So far I've been working on Ilan's branch which Mark's new one is based on, but will check out the other one

falexwolf:
@keller-mark, updating the main user guide within the lamin docs here:

- https://github.com/laminlabs/lamin-spatial/pull/24

PR #1970, opened 2024-09-25
⚡️ Speed up `from_values`

sunnyosun:
3.7s → 1.3s from public source (records not yet registered):
Screenshot 2024-09-29 at 16 03 46

1.1s → 0.1s from lamindb registry (records already registered):
Screenshot 2024-09-29 at 16 07 21


falexwolf:
Are you going to move it to `CanValidate` in this PR or another one?

PR #1973, opened 2024-09-25
⚡️ Speed up describe

sunnyosun:
Before: **10s** on `laminlabs/cellxgene`, `us-west-2`:
Screenshot 2024-09-27 at 14 39 03

After: **2.4s** on `laminlabs/cellxgene`, `us-west-2`:
Screenshot 2024-09-27 at 14 32 35

Before: 7.1s on `laminlabs/lamin-dev`, `us-east-1`:
Screenshot 2024-09-27 at 14 39 25

After: 0.6s on `laminlabs/lamin-dev`, `us-east-1`:
Screenshot 2024-09-27 at 14 33 33


falexwolf:
Given all the complex logic introduced here, I think we should have a dedicated test that checks that the .describe() output is correct, for the SQLite and the Postgres backend. The next person who needs to work on this code (hopefully just maintenance) will be lost otherwise.

The unit tests run on Postgres, so one could just add this check to the transfer test in the unit tests.

Most notebook tests run on SQLite, so one could add the same test to the transfer notebook.

PR #1983, opened 2024-09-27
🚸 `lamin load` → `lamin connect` & `lamin get` → `lamin load`

Koncopd:
This PR implements @ap--'s suggestions to improve the command name consistency of the CLI.

It's backward compatible.

Before | After
--- | ---
image | image

Based on the below issue:

- https://github.com/laminlabs/pfizer-lamin-usage/issues/25

Needs:

- https://github.com/laminlabs/lamin-cli/pull/79

### DIFF
--- a/docs/faq/setup.ipynb
+++ b/docs/faq/setup.ipynb
@@ -19,7 +19,7 @@
    },
    "outputs": [],
    "source": [
-    "!lamin load --unload"
+    "!lamin disconnect"
    ]
   },
   {
--- a/docs/setup.ipynb
+++ b/docs/setup.ipynb
@@ -111,16 +111,16 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "## Load an instance"
+    "## Connecting to an instance"
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Load your own instance:\n",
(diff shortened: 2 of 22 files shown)


falexwolf:
Looks great! I'll review this in detail during the weekend.

falexwolf:
I'll polish this a bit more now and hope to merge it this morning.

Zethson:
This is great! Should we have command pairs in one row to simplify how we display he commands?

e.g.

Connect/disconnect              |||  Connect/disconnect to an instance

I think that it would still convey that these are independent commands but it would make the list of commands much shorter. The version that @Koncopd implemented is very explicit and not bad at all, but we might get away with something a bit more grouped?

Zethson:
Random feedback that you can totally ignore if you don't think that it's helpful:

I would consider switching `info` and `init` to have `init` before `info`.
I find the term `entity` not very descriptive.
The explanations in `configure` are also not very helpful.

falexwolf:
> Should we have command pairs in one row to simplify how we display the commands?

I think this could look great but it might also take too much horizontal space. Feel free to give it a shot!

> I would consider switching info and init to have init before info.

Fine for me! Rationale for that `init` is at the bottom is that it's almost never used.

> I find the term entity not very descriptive.
> The explanations in configure are also not very helpful.

Any ideas for making it better?

PR #2007, opened 2024-10-01
💄 Strip `NotebookNotSaved` from report after `ln.finish()`

Koncopd:


falexwolf:
Problem is that this will also strip the output of any other tool that uses finish, say wandb

falexwolf:
Two ways of resolving: (1) add more constraints that account for the lamindb namespace (2) require it to be the last cell

Can you check whether there is such a requirement within ln.finish to be called in the last cell if it's the global Context object?

I believe we had this for publish() in nbproject and I'd say it makes only sense to have it; doesn't it?

Koncopd:
> Problem is that this will also strip the output of any other tool that uses finish, say wandb
Hm, no, why do you think so?

falexwolf:
Ok, then good to merge! :D

Issue #1860, opened 2024-08-28
🎨 Confusing name: `Curate.add_validated_from`


falexwolf:
The problem with the below is that values in a public ontology are not validated in the LaminDB definition. Hence, the name of the method is highly confusing. 

<img width="559" alt="image" src="https://github.com/user-attachments/assets/54621a47-f475-4147-b87e-ae0a02d664d6">

How shall this method be called?

falexwolf:
This here is still a quite basic thing that we should get resolved, shouldn't we? I think you wanted to work on it at some point @Zethson?

Zethson:
I'll think about it next week but if we change the default of `Curator` to also validate against records without a `Source` record even if a `Source` is passed, the comment above would certainly be false and the name fine.

falexwolf:
We need a big clear simple box that defines what “validated” means. This is urgent as it’s so basic.

In my mind “validated” == “contained in registry or an underlying ontology source”.

If this is what “validated” is curator.add_validated_from should not be user-facing but records that are not yet in the ontology source should be added to registry upon the first call of .validate() that touches them.

What’s important about this “auto-import-from-the-ontology-source” is that the definition gets easier in practice: validated record == record exists in a registry. Hence, the whole complexity of the ontology source is no longer important as soon as the record is imported.

It’s important that we have an easy practical definition of “validated” that’s free of “validating against source” and “validating against another instance” both of which are complicated things.

Zethson:
> In my mind “validated” == “contained in registry or an underlying ontology source”.

Yes that's also where @sunnyosun and me ended up. However, I think the current behavior when `sources` is passed to `Curate` would break this assumption because it would only validate against the specific ontology source and NOT the remainder of the registry (basically values that have no source). This needs to be changed for consistency.

+1 the rest of what you wrote.

falexwolf:
Can we make this a priority then so that it's in a LaminDB release ASAP? Asking both of you @Zethson @sunnyosun

Zethson:
While I could figure it out, I think @sunnyosun knows this part better and is probably substantially faster? @sunnyosun I'm happy to look into it if you want me to. Just let me know.

sunnyosun:
What about we run `add_validated_from("all")` under the hood of `.validate()`? Essentially all validated records will be saved before running validation. This way we make `add_validated_from` no longer user facing.

falexwolf:
[...]

PR #2022, opened 2024-10-07
📝 Document how to subclass `Curator`

Zethson:
- Adds a guide on how to subclass the Curator base classes

Comes with:

- https://github.com/laminlabs/lamin-docs/pull/164


falexwolf:
This looks already promising!

Often I wish that we'd delete one docs page for any new page added; but this here isn't such a case.

Small note: Can you please use `ln.track(uid)` instead of
image

falexwolf:
Should one make it more explicit that this guide is about subclassing `Curator` to enable convenient curation of datasets of any format rather than "custom curation" of any dataset, the topic of https://docs.lamin.ai/curate-any?

To rephrase: The term "custom" may not be specific enough because there are two ways of custom curation. One via subclassing Curator (aka "high-level") and one via using the "low-level" lamindb API directly.

Zethson:
Went more into the direction of `Subclassing Curator`. However, I don't know whether this header gives away how useful this feature is because I think that it's one of the best features of lamindb.

falexwolf:
>  However, I don't know whether this header gives away how useful this feature is because I think that it's one of the best features of lamindb.

Then how about "Custom Curators" or similar and the "curate-any" guide is renamed to "Curate using the low-level registry API" or something?

falexwolf:
Now starting a detailed review. Thanks for this, Lukas!

docs/curate.md
```diff
@@ -5,4 +5,5 @@
 
 curate-df
 curate-any
+curate-subclass
```
falexwolf, line 8: If this is the recommended way to curate custom datasets we should put it first. Or we integrate it into curate-any and the curate-any guide starts with subclassing the Curator and only then walks through how to do it low-level.

3 documents on curation is potentially a bit intimidating in particular when curate-any will almost never be used and should be more like a footnote?

falexwolf:
The previous commit refactored the changes made in this PR. Let me know if anything doesn't make sense, Lukas!

I'll now add another commit to simplify the docstring of `Curatator`:

image

We should then cross-link guides and find an answer to this question:

https://github.com/laminlabs/lamindb/pull/2022#discussion_r1802696408

falexwolf:
Also added this:

- https://github.com/laminlabs/lamin-docs/pull/164

falexwolf:
So, I made an attempt at the curator docstring, but this still throws me off: 

- https://github.com/laminlabs/lamindb/issues/1860

We can't have an easy explanation for what the Curator does if we don't have a simple definition of what validation is.

I think this PR here is a great opportunity to fix this once and for all.

lamindb/_curate.py
```diff
@@ -950,29 +950,27 @@ def save_artifact(
...
-    - Values that can successfully validated and already exist in the registry.
-    - Values which are new and not yet validated or potentially problematic values.
-
-    3. Determine how to handle validated and non-validated values:
-
-    - Validated values not yet in the registry can be automatically registered using :meth:`~lamindb.core.DataFrameCurator.add_validated_from`.
-    - Valid and new values can be registered using :meth:`~lamindb.core.DataFrameCurator.add_new_from`.
-    - All unvalidated values can be accessed using :meth:`~lamindb.core.DataFrameCurator.non_validated` and subsequently removed from the object at hand.
+    - validated values not yet in the registry can be automatically registered using :meth:`~lamindb.core.DataFrameCurator.add_validated_from`
+    - new values found in the data can be registered using :meth:`~lamindb.core.DataFrameCurator.add_new_from`
```
falexwolf, line 972: What is the difference between "new values" and "non-validated values"?

docs/curate.md
Zethson: I think it is the recommended way if you want to enforce structure for a specific format and especially across people. However, if you're curating a dataset as a one off, subclassing `Curator` is too complex and annoying. Then the content in `curate-any` applies better I'd argue.

lamindb/_curate.py
Zethson: I think that the old text was not precise enough. The new one looks fine now.

falexwolf:
@Zethson, I'm done here. I'd love for you to drive this home. 😇

Zethson:
I decided to keep it as 3 documents for now instead of merging them and also showing this last. I'll keep a close eye on this and with a few days of reflection I expect myself to make a few small changes.

The `validated_from` issue we're aware of and it'll be fixed in subsequent PRs.

falexwolf:
Sounds good! We'll polish more.

Re the very last commit: I don't like the `{doc}` syntax for the cellxgene ref because it shows the long title and hence makes the sentence & line very long. I know there is syntax to make the intersphinx work without showing the title, but I keep forgetting it. I'll fix this in another iteration.

PR #2032, opened 2024-10-08
🐛 Do not double track runs

falexwolf:


lamindb/core/_label_manager.py
```diff
@@ -39,7 +39,7 @@ def get_labels_as_dict(self: Artifact | Collection, links: bool = False):
         "links_collection",
         "links_artifact",
         "links_feature_set",
-        "previous_runs",
+        "_previous_runs",
```
falexwolf, line 42: @sunnyosun, how do you account for this whole list in the postgres version of describe()?

PR #2041, opened 2024-10-10
🎨  for circular input/outputs

sunnyosun:


falexwolf:
This is great as it also helps us to understand the edge cases in which users manage to fabricate a cycle.

PR #2044, opened 2024-10-11
🐛 Return include fields when one of them is empty

sunnyosun:
Print warning for no-value columns in df include:
Screenshot 2024-10-11 at 14 21 25


andreassteffen:
thank you so much! Will test later.

sunnyosun:
More fixes and tests here: https://github.com/laminlabs/lamindb/pull/2045

PR #2061, opened 2024-10-16
✨ Error when renaming internal features and labels

sunnyosun:
When to error:
1. If a feature is an internal feature, aka it is a member of a featureset.
    Screenshot 2024-10-18 at 15 55 42
2. If a label is an internal label, aka it is associated with an internal feature.
    Screenshot 2024-10-18 at 15 55 33

To bypass the renaming error:
1. `artifact.features.make_external(feature)`: remove feature from all featuresets.
2. `artifact.labels.make_external(label)`: keep the label associated with artifacts, but remove set `feature=None` in the link tables. This label will no longer associated with any internal faetures.

Full renaming UX:
https://lamin.ai/laminlabs/lamindata/transform/u1s4bAB3sliY0000

### DIFF
--- a/lamindb/_record.py
+++ b/lamindb/_record.py
@@ -7,7 +7,7 @@ import dj_database_url
 import lamindb_setup as ln_setup
 from django.db import connections, transaction
 from django.db.models import IntegerField, Manager, Q, QuerySet, Value
-from lamin_utils import logger
+from lamin_utils import colors, logger
 from lamin_utils._lookup import Lookup
 from lamindb_setup._connect_instance import (
     get_owner_name_from_identifier,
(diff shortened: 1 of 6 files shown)


falexwolf:
Can you please add more production-ready screenshots or use pretty text snippets? This is a big new feature that needs a high-quality announcement in the docs.

Also: there is a grammatical error in the user dialogue which makes it feel somewhat haphazard:

image

I know this is well thought but the PR overall doesn't give the appearance.

falexwolf:
The other big thing that misses is the definition of "internal" vs. "external" feature. It's nowhere defined in the docs. So, guides & API docs should be updated. Even the PR title can't be understood if this isn't documented.

falexwolf:
> what error?

It's "rename a feature" not "rename feature".

For the second line, one could maybe do: "These artifacts contain the old name of the feature: [...]"

Because "not validated" sounds a bit technical and doesn't explain specifically what the problem is.

sunnyosun:
Feel free to reword the logging if you don't like it, I asked for your review before.

falexwolf:
@sunnyosun, instead of screenshots, can you put really pretty formatted text messages? This is one of the most important features of the past months and we should have a polished announcement for it.

I also want to use this in the lakehouse tutorial (without this feature, lamindb wouldn't be a lakehouse).

sunnyosun:
> @sunnyosun, instead of screenshots, can you put really pretty formatted text messages? This is one of the most important features of the past months and we should have a polished announcement for it.
> 
> I also want to use this in the lakehouse tutorial (without this feature, lamindb wouldn't be a lakehouse).

Done!

PR #2081, opened 2024-10-18
🔥 Remove curator.add_validated_from

sunnyosun:


falexwolf:
Can you please add comprehensive documentation to every PR? This is a big breaking change. Users need a migration guide and a rationale for why they need to deal with this change.

PR #2092, opened 2024-10-22
✨ Python 3.12 support

Zethson:
- Uses 3.12 instead of 3.11 in the CI
- Specifies 3.12 support in our classifiers
- Removes support for Python 3.13

Docs don't build with 3.12:

```
Extension error:
Could not import extension ablog (exception: No module named 'pkg_resources')
```

See https://github.com/laminlabs/lndocs/issues/105

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,7 +36,7 @@ jobs:
       - uses: actions/setup-python@v5
         if: ${{ matrix.group != 'storage' }}
         with:
-          python-version: "3.11"
+          python-version: "3.12"
       - name: cache pre-commit
         uses: actions/cache@v4
         with:
@@ -134,7 +134,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
         with:
-          python-version: "3.11"
+          python-version: "3.12"
       - run: |
           python -m pip install -U uv
           uv pip install --system coverage[toml]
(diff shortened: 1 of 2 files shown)


Koncopd:
We should not use only 3.12 in the CI, it should still have 3.9, 3.10 and 3.12 i think.

Koncopd:
The problem with `ablog` is due to our pin actually https://github.com/laminlabs/lndocs/blob/d5247a9fc957a9b74091a9e6ed6d79c713bca106/pyproject.toml#L11

Zethson:
> We should not use only 3.12 in the CI, it should still have 3.9, 3.10 and 3.12 i think.

I think 3.9 as a lower bound and 3.12 as the upperbound is fine. We can of course have a bigger matrix but why?

> The problem with ablog is due to our pin actually https://github.com/laminlabs/lndocs/blob/d5247a9fc957a9b74091a9e6ed6d79c713bca106/pyproject.toml#L11
> pinging @falexwolf
> The latest version of ablog doesn't use pkg_resources.

I already created an issue for this and assigned @falexwolf :)

Koncopd:
> I think 3.9 as a lower bound and 3.12 as the upperbound is fine. 

Agree, you are right.

Zethson:
Merging this now because we still build the docs with 3.10. We can upgrade the docs job later.

PR #2097, opened 2024-10-22
✨ Add n_observations function to read n_obs from AnnData and tiledbsoma

Koncopd:
Re https://laminlabs.slack.com/archives/C04FPE8V01W/p1726658747846539

This function infers `n_obs` from `anndata` (both `zarr` and `hdf5`) and `tiledbsoma`. I am reluctant to add this to `Artifact` by default though.

### DIFF
--- a/lamindb/core/storage/__init__.py
+++ b/lamindb/core/storage/__init__.py
@@ -19,6 +19,7 @@ Array accessors.
 from lamindb_setup.core.upath import LocalPathClasses, UPath, infer_filesystem
 
 from ._backed_access import AnnDataAccessor, BackedAccessor
+from ._observations import n_observations
 from ._tiledbsoma import save_tiledbsoma_experiment
 from ._valid_suffixes import VALID_SUFFIXES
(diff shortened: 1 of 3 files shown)


Koncopd:
Do we want it to be used by default?

falexwolf:
This looks perfectly great. We should just call this in `Artifact()` by default if a corresponding path is called within say `from_anndata()`.

Koncopd:
Ok, makes sense. It still needs a bit of improvement though.
Should we also have `.from_tiledbsoma`?

falexwolf:
> Ok, makes sense. It still needs a bit of improvement though.
Should we also have .from_tiledbsoma?

Yes, I think so!

falexwolf:
Can this be closed now?

Koncopd:
No, i still need to transfer the `tiledbsoma` related code.

Koncopd:
Implemented in https://github.com/laminlabs/lamindb/pull/2438 and https://github.com/laminlabs/lamindb/pull/2404

PR #2103, opened 2024-10-23
✨ Truncate standardization output

Zethson:
- Limits output of standardized and unsaved values to max 10 values in line with other print limits
- Fixes an intendation issue of the warning output


sunnyosun:
Hmm, auto-saving doesn’t look right to me, .standardize is meant to be a non-write operation, similar to .validate and .inspect.
The user pattern is expected to be:
```
standardized_names = bt.Gene.standardize(names)
records = bt.Gene.from_values(standardized_names)
ln.save(records)
```
I wouldn’t make un-explicit save calls in low-level API. I think it’s OK in Curator because that’s a high-level class. For standardize I don’t think we should do it.

Issue #2091, opened 2024-10-21
AttributeError: 'str' object has no attribute 'field' when passing "symbol" instead of "bt.Gene.symbol"


Zethson:
### Report

```
!lamin init --storage ./test-bug

import lamindb as ln
import pandas as pd
import anndata as ad
import numpy as np

adata = ad.AnnData(X=np.random.rand(2, 1),
                   obs={"test": ["val1", "val2"]},
                   var=pd.DataFrame(index=["TP53"]))

curator = ln.Curator.from_anndata(adata, var_index="symbol", categoricals={"test": ln.ULabel.name})

curator.add_new_from("all")
```

results in

```
{
	"name": "AttributeError",
	"message": "'str' object has no attribute 'field'",
	"stack": "---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[1], line 14
      8 adata = ad.AnnData(X=np.random.rand(2, 1),
      9                    obs={\"test\": [\"val1\", \"val2\"]},
     10                    var=pd.DataFrame(index=[\"TP53\"]))
     12 curator = ln.Curator.from_anndata(adata, var_index=\"symbol\", categoricals={\"test\": ln.ULabel.name})
---> 14 curator.add_new_from(\"all\")

File ~/PycharmProjects/lamindb/lamindb/_curate.py:272, in DataFrameCurator.add_new_from(self, key, organism, **kwargs)
    270     raise ValueError(\"Cannot pass additional arguments to 'all' key!\")
    271 self._kwargs.update({\"organism\": organism} if organism else {})
--> 272 self._update_registry(key, validated_only=False, **self._kwargs, **kwargs)

File ~/PycharmProjects/lamindb/lamindb/_curate.py:286, in DataFrameCurator._update_registry(self, categorical, validated_only, **kwargs)
    284 def _update_registry(self, categorical: str, validated_only: bool = True, **kwargs):
    285     if categorical == \"all\":
--> 286         self._update_registry_all(validated_only=validated_only, **kwargs)
    287     elif categorical == \"columns\":
    288         self._save_columns(validated_only=validated_only, **kwargs)

File ~/PycharmProjects/lamindb/lamindb/_curate.py:516, in AnnDataCurator._update_registry_all(self, validated_only, **kwargs)
    514 \"\"\"Save labels for all features.\"\"\"
    515 logger.info(\"saving validated records of 'var_index'\")
[...]

PR #2105, opened 2024-10-23
✨ var index support

Zethson:
Fixes https://github.com/laminlabs/lamindb/issues/2091

- Adds support for str in `var_index`. However, this currently assumes that the `var_index` is `Gene`. I am not sure whether this is the right way to go about it. Maybe we should actually just not do this.

WDYT @sunnyosun ?


sunnyosun:
I'd suggest you add a type check instead.

Issue #2060, opened 2024-10-16
Add a function to recursively filter parents/children


sunnyosun:
https://laminlabs.slack.com/archives/C046F63LJJ2/p1728638400378999?thread_ts=1728315596.529489&cid=C046F63LJJ2

PR #2106, opened 2024-10-23
✨ Recursively query parents and children

Koncopd:
This PR adds `query_parents` and `query_children` methods to `HasParents` class. This allows to query all parents and parents of parents etc. recursively  for all records that have `parents` (instances of subclasses of `HasParents`).

Example:
```
import lamindb as ln

label1 = ln.ULabel(name="label1")
label2 = ln.ULabel(name="label2")
label3 = ln.ULabel(name="label3")
label1.save()
label2.save()
label3.save()
label1.children.add(label2)
label2.children.add(label3)
label3.query_parents() # returns a QuerySet with label1 and label2
label1.query_children() # returns a QuerySet with label2 and label3
```
Closes https://github.com/laminlabs/lamindb/issues/2060

### DIFF
--- a/lamindb/_parents.py
+++ b/lamindb/_parents.py
@@ -15,11 +15,37 @@ from ._record import get_name_field
 if TYPE_CHECKING:
     from lnschema_core.types import StrField
 
+    from lamindb.core import QuerySet
+
 LAMIN_GREEN_LIGHTER = "#10b981"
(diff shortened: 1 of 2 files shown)


falexwolf:
This looks very elegant both in terms of the actual implementation and the user-facing API design. Kudos!

The only thing that's missing: a nice user-facing description of this PR.

Issue #2117, opened 2024-10-29
Loading dataset from a different instance that does not have all schemas of the current instance errors


Zethson:
### Report

```
lamin init --storage ./test-perturbation --schema bionty,wetlab,findrefs

ln.track("HIRTYxL3aZc70000")

adata = ln.Artifact.using("laminlabs/lamindata").get(uid="Xk7Qaik9vBLV4PKf0001").load()
```

results in

```
{
	"name": "ProgrammingError",
	"message": "relation \"findrefs_reference\" does not exist
LINE 1: SELECT 1 AS \"a\" FROM \"findrefs_reference\" INNER JOIN \"findre...
                             ^
",
	"stack": "---------------------------------------------------------------------------
UndefinedTable                            Traceback (most recent call last)
File ~/miniconda3/envs/pertpy/lib/python3.12/site-packages/django/db/backends/utils.py:105, in CursorWrapper._execute(self, sql, params, *ignored_wrapper_args)
    104 else:
--> 105     return self.cursor.execute(sql, params)

UndefinedTable: relation \"findrefs_reference\" does not exist
LINE 1: SELECT 1 AS \"a\" FROM \"findrefs_reference\" INNER JOIN \"findre...
                             ^


The above exception was the direct cause of the following exception:

ProgrammingError                          Traceback (most recent call last)
Cell In[5], line 1
----> 1 adata = ln.Artifact.using(\"laminlabs/lamindata\").get(uid=\"Xk7Qaik9vBLV4PKf0001\").load()
      2 adata.obs.head(3)

File ~/PycharmProjects/lamindb/lamindb/_artifact.py:994, in load(self, is_run_input, **kwargs)
    992     access_memory = load_to_memory(cache_path, **kwargs)
    993 # only call if load is successfull
--> 994 _track_run_input(self, is_run_input)
    995 return access_memory

File ~/PycharmProjects/lamindb/lamindb/core/_data.py:465, in _track_run_input(data, is_run_input, run)
    458             is_valid = True
    459         return (
    460             data.run_id != run.id
    461             and not data._state.adding  # this seems duplicated with data._state.db is None
    462             and is_valid
    463         )
--> 465     input_data = [data for data in data_iter if is_valid_input(data)]
    466     input_data_ids = [data.id for data in input_data]
    467 if input_data:

File ~/PycharmProjects/lamindb/lamindb/core/_data.py:457, in _track_run_input.<locals>.is_valid_input(data)
    450 else:
[...]

PR #2132, opened 2024-11-05
🐛 Fix transfering artifacts from a source instance with fewer schema modules

sunnyosun:
Resolves: https://github.com/laminlabs/lamindb/issues/2117

### DIFF
--- a/tests/storage/test_transfer.py
+++ b/tests/storage/test_transfer.py
@@ -12,7 +12,7 @@ def test_transfer_from_remote_to_local():
     ln.ULabel.filter().delete()
     bt.CellType.filter().delete()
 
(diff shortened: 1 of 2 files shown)


tests/storage/test_transfer.py
```diff
@@ -100,8 +100,17 @@ def test_transfer_from_remote_to_local():
     assert artifact2.organisms.get(name="mouse") == bt.settings.organism
     assert artifact.features["obs"].get(name="organism").uid == "existing"
 
+    # test transfer form a remote instance with less schema modules to a local instance (laminlabs/lamin-site-assets)
```
falexwolf, line 103: It should be called "source instance" and "fewer schema modules than target instance".

"local" and "remote" aren't relevant concepts here.

Koncopd:
We are running at the same time.

PR #2134, opened 2024-11-06
📝 Add gene symbol usage and  mapping FAQ

Zethson:
See: https://github.com/laminlabs/pfizer-lamin-usage/issues/76 and https://github.com/laminlabs/bionty/issues/143

- Fixes a Future warning in `can_validate`
- Adds a FAQ section on the usage of Gene symbols and mapping to ENSEMBL IDs


falexwolf:
I do like this FAQ and in particular the whole context story, but this PR and the entry in the changelog doesn't get this across at all. It gives off the impression of a random FAQ.

Can you motivate a bit in the PR description why this was added?

---

It looks like ruff no longer validates strikt indendation.

We shouldn't have hanging indendation anywhere in the codebase and the docs. This throws me off:

image

For as long as we don't cover this with linting, let's watch out, please @sunnyosun @Zethson @Koncopd

PR #2135, opened 2024-11-06
🚸 Improve search

sunnyosun:
Key improvements:
- prioritize `startswith` and isolated phrases (e.g. "naive B cell", "B cell, ..." over "club cell" when searching "b cell")
- sort by shorter names first

Note: `centrocyte` appears in the "b cell" search because there's a perfect match of "B cell" in the description, same for "t cell" results.

LaminDB Before|LaminDB After |Hub
---|---|---
Screenshot 2024-11-07 at 12 24 25|Screenshot 2024-11-07 at 12 42 48|Screenshot 2024-11-07 at 12 21 23
Screenshot 2024-11-07 at 12 24 31|Screenshot 2024-11-07 at 12 43 10|Screenshot 2024-11-07 at 12 22 12
Screenshot 2024-11-07 at 12 24 17|Screenshot 2024-11-07 at 12 43 18|Screenshot 2024-11-07 at 12 22 34
Screenshot 2024-11-07 at 13 10 44|Screenshot 2024-11-07 at 13 10 15|Screenshot 2024-11-07 at 13 11 40
Screenshot 2024-11-07 at 16 06 53|Screenshot 2024-11-07 at 21 43 05|Screenshot 2024-11-07 at 16 11 32
Screenshot 2024-11-07 at 16 09 37|Screenshot 2024-11-07 at 21 43 25|Screenshot 2024-11-07 at 16 11 58

### DIFF
--- a/lamindb/_record.py
+++ b/lamindb/_record.py
@@ -260,12 +260,27 @@ def _search(
     for field in fields:
         exact_expression |= Q(**{f"{field}__{case_sensitive_i}exact": string})
 
-    # Layer 2: Contains full phrase
+    # Layer 2: Contains the isolated phrase (prioritize names ending with the search phrase)
+    iso_phrase_expression = Q()
+    for field in fields:
+        iso_phrase_expression |= Q(
+            **{f"{field}__{case_sensitive_i}contains": f" {string}"}
+        )
+        iso_phrase_expression |= Q(
+            **{f"{field}__{case_sensitive_i}contains": f"{string} "}
+        )
+        iso_phrase_expression |= Q(
+            **{f"{field}__{case_sensitive_i}contains": f" {string} "}
+        )
+        iso_phrase_expression |= Q(**{f"{field}__{case_sensitive_i}startswith": string})
+        iso_phrase_expression |= Q(**{f"{field}__{case_sensitive_i}endswith": string})
+
+    # Layer 3: Contains full phrase as a phrase
     phrase_expression = Q()
     for field in fields:
         phrase_expression |= Q(**{f"{field}__{case_sensitive_i}contains": string})
 
-    # Layer 3: Contains individual terms and pairs
+    # Layer 4: Contains individual terms and pairs
     terms_expression = Q()
     for field in fields:
         for term in search_terms:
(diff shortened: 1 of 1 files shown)


Koncopd:
@sunnyosun @falexwolf here is the benchmark for this PR https://github.com/laminlabs/lamindb/pull/2141

Koncopd (approved):

falexwolf:
These are great improvements!

I wish there was good way in laminhub to document such changes but I guess there isn't for now.

falexwolf:
@fredericenard, please also take a look here.

And then we'll discuss in the benchmarking PR how we proceed with organizing code across lamindb, bionty and laminhub.

falexwolf (approved):

falexwolf (changes requested):
One thing that's totally not covered is longer search phrases along the lines of "CD8-positie cytokine T cell"

So, I believe we need a benchmarking case for this.

Koncopd:
Added more searches to the benchmark. 
`test_search_synonyms`, commented out in the benchmark.

sunnyosun:
Fixed synonyms. The only thing is that it's getting a bit slower now ~0.8s per search (before 0.2-0.3s) because of all the different layers.

lamindb/_record.py
```diff
@@ -218,8 +221,30 @@ def _search(
                     ) from error
             else:
                 fields.append(field)
+    if not name_field:
+        name_field = (
+            fields[0] if not hasattr(registry, "_name_field") else registry._name_field
+        )
+
+    def tokenize_search_string(search_str: str) -> list[str]:
+        # Split the string
```
Zethson, line 230:
```suggestion
```

Zethson: Self documenting code here.

Zethson (commented):
This is really cool!

I know that no review of me was requested but I thought I'd leave a couple of comments still. Feel free to ignore them!

1. Would it be possible to add 1-3 sentences of high level motivation and explanation of this new layered algorithm to the docstring or where appropriate, please?
2. We have the nice search benchmark now. Is there a way to add slightly more sophisticated tests to ensure that no changes of this code will lead to regressions of the search performance? @Koncopd benchmarking notebook is great but it's not regularly run, is it?

These things can also be added later if you think that they're useful...

Zethson:
Ohh and can we get rid of this now?
image
🚀

falexwolf:
Yes, removing will be the goal.

image

@Koncopd is making a push to consolidate all 3 search algorithms into one clearly documented and benchmarked solution.

Sunny's PR here has good ideas but it's not suitable for the hub due to performance. So, the hope is Sergei can replicate the UX with server-side code (essentially correctly using postgres plugins). We can still use Sunny's code for dataframes and sqlite; it'll give the same results but just run slower which is OK in that context.

Issue #2143, opened 2024-11-08
🚸 Validate minimum `uid` length


falexwolf:
Can't be shorter than 12 characters.

falexwolf:
Encountered by @fredericenard

PR #2144, opened 2024-11-08
🚸 Validate minimum `uid` length

falexwolf:
- https://github.com/laminlabs/lamindb/issues/2143


sunnyosun:
This is a breaking change, the uid field has been used to sync some external ids based on the definition is "the universal id of the record".

falexwolf:
Your concern is addressed, @sunnyosun:

- https://github.com/laminlabs/lamindb/pull/2167

PR #2152, opened 2024-11-11
⬆️ Upgrade anndata to 0.11.0

Koncopd:


noxfile.py
```diff
@@ -90,7 +90,7 @@ def install_ci(session, group):
         run(session, "uv pip install --system huggingface_hub")
     elif group == "unit-storage":
         extras += "aws,zarr,bionty"
-        run(session, "uv pip install --system tiledbsoma")
+        run(session, "uv pip install --system tiledbsoma>=1.15.0rc3")
```
Zethson, line 93: Could you please add a comment here why you've added this upperbound? Else people have to trace it back to this PR to understand that this is required for anndata 0.11.0 support.

PR #2157, opened 2024-11-12
✨ Improve runtime formatting

Zethson:
Before:
`→ finished Run('7xgSkdkp') after 0:10:06.762290 at 2024-11-12 12:14:24 UTC`

After:
`→ finished Run('7xgSkdkp') after 0h 10m 06s at 2024-11-12 12:14:24 UTC`

We can hyperoptimize this by removing leading 0s but I think this is fine.

### DIFF
--- a/lamindb/_finish.py
+++ b/lamindb/_finish.py
@@ -233,9 +233,9 @@ def save_context_core(
 
     # finalize
     if not from_cli:
-        run_time = (
-            run.finished_at - run.started_at.split(".")[0]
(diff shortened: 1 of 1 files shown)


falexwolf:
(I really might be missing something but I have this deja vu of prettifying date strings many years ago with very simple usage of the stdlib rather than string wrangling and splitting and all the not-so-pretty things you're doing in this PR.)

falexwolf:
I also like about this PR that there is a before after 😇

PR #2162, opened 2024-11-12
🎨 Improve runtime formatting code

Zethson:
See https://github.com/laminlabs/lamindb/pull/2157#issuecomment-2470812207

Claude and chatGPT both suggested the earlier version with this as alternative. Suggested in https://stackoverflow.com/questions/538666/format-timedelta-to-string. Note that the earlier solution is also suggested in this stackoverflow thread!

Less "manual" string wrangling is generally better and more robust - I agree. In this case, it's still a bit meh especially as we're performing some math.

I can live with either solution.


Zethson:
@falexwolf do you prefer this solution? I might have a slight preference for this new one but no strong feelings.

falexwolf:
Interesting! I like this solution much better than the previous one. 🤔 

But I'm surprised that nobody suggests using the stdlib for it: https://docs.python.org/3/library/datetime.html#strftime-and-strptime-format-codes. Am I missing something?

In any case, not showing `0h` and `0m` if a run just takes `5s` would be really nice given how much work goes into this now.

Zethson:
That's because these are different objects. a `timedelta` object is not a `datetime` object.

> how much work goes into this now.

Not a lot - just a bit more communication :)

falexwolf:
You're right, sorry!

Claude suggests:
```
days = td.days
seconds = td.seconds
hours = seconds // 3600
minutes = (seconds % 3600) // 60
secs = seconds % 60
print(f"{days}d {hours}h {minutes}m {secs}s")  # 1d 2h 30m 0s
```

Zethson:
Claude suggested to me the one that this PR is currently implementing :) I mean divmod or not doesn't make a difference anymore but if you prefer `//` I can adapt it.

falexwolf:
No, as I said I like the solution you have here. I'd just print `5s` instead `0h 0m 5s`. So, if you could please account for that! 🙏

PR #2163, opened 2024-11-12
⚡️ Improve speed and relevance of search

Koncopd:
This changes `Record.search()` approach to rule-based ranking.

It uses a different few rules now:

**Exact matches are given preference.**

`bt.CellType.search("B cell")` returns a record with `name="B cell"` or `synonyms="...|B cell|..."` as the top record.

**Then occurrences as sub phrase are ranked highly.**

`bt.CellType.search("B cell")` ranks record with `name, description, etc ="... B cell of ..."` occurrences highly.

**Occurrences at the beginning or end  of a subphrase are also given a high rank.**

`ln.Artifact.search("Experim")` or `ln.Artifact.search("eriment")` return records with `description="... Experiment ..."`

**It also just checks `contains`, but it has the lowest rank score.**

All these rules are combined into a single ranking value.


Koncopd:
Examples of searches are here https://6733f36a39f9b718a068cc2b--lamindb-qnwk.netlify.app/faq/benchmark-search

Koncopd:
@falexwolf this is the approach we discussed yesterday, the results look reasonable for me and the speed is good.
I didn't add any postgres specific search here, can do.

falexwolf:
Great!

Can you add extensive discussion, benchmarking results, and related resources to the top of this PR?

And link a report on `laminlabs/lamindata` that documents the performance?

Koncopd:
Hm, i am not sure where to put the search benchmark.
I need a separate postgres instance for it, so maybe better to keep as a notebook.
But maybe not in FAQ.

I would just check that the top result is actually the full match, not the top N ordering.

laminhub maybe, not sure it is a good idea for bionty.

Koncopd:
Also as you see it fails with fuzzy matching, i can add trigram or fulltext search in postgres for this case, i.e. when there are no results from the normal search or maybe just do both if it is postgres and combine.

falexwolf:
Do you want to call to clarify?

Koncopd:
A bit later today.

falexwolf:
Ok! Just ping me!

> Also as you see it fails with fuzzy matching

Along the lines of what we discussed, this would now be the desired behavior. We shouldn't get a hit anymore on fuzzy matching just as you don't get a hit when you search GitHub or Slack with fuzzy matching.

What should be tested though is that `suggest_records_with_similar_names` is still working as expected: https://github.com/laminlabs/lamindb/blob/dfdd732a5083c6677f7e345348f86087481e3c57/lamindb/_record.py#L74

I fear there is no unit test for this yet. Can you please add a unit test for it?

PR #2170, opened 2024-11-14
🩹 Fix anndata io `read_zarr` warning

Zethson:
```
anndata/_io/__init__.py:12: FutureWarning: Importing read_zarr from `anndata._io` is deprecated. Please use anndata.io instead.
  warnings.warn(
```


Koncopd:
No, this is incorrect, `anndata.io` was introduced in 0.11.0, we can't really import like this due to the need to support earlier versions.

Zethson:
Ah yeah, that's true.

PR #2172, opened 2024-11-14
✨ Add an artifact loader for `.yaml`

rcannood:
* Fix incorrect documentation for `load_image`
* Add loaders for `.gif` and `.yaml`.

In addition, we could add a data loader for `.rds`, which when called, produces an error message that you should use laminr for loading `.rds` files. WDYT?

### DIFF
--- a/lamindb/core/loaders.py
+++ b/lamindb/core/loaders.py
@@ -109,7 +109,7 @@ def load_json(path: UPathStr) -> dict:
 
 
 def load_image(path: UPathStr):
-    """Display `.svg` in ipython, otherwise return path."""
+    """Display `.jpg`, `.png` or `.gif` in ipython, otherwise return path."""
     if is_run_from_ipython:
         from IPython.display import Image, display
 
@@ -141,6 +141,7 @@ FILE_LOADERS = {
     ".h5mu": load_h5mu,
     ".jpg": load_image,
     ".png": load_image,
+    ".gif": load_image,
     ".svg": load_svg,
 }
 
--- a/lamindb/core/loaders.py
+++ b/lamindb/core/loaders.py
@@ -10,6 +10,7 @@
    load_h5mu
    load_html
    load_json
+   load_yaml
    load_image
    load_svg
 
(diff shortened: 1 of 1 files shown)


rcannood:
FYI -- laminlabs/laminr#81 added a data loader for .rds.

falexwolf:
Thanks, Robrecht! Great!

> In addition, we could add a data loader for .rds, which when called, produces an error message that you should use laminr for loading .rds files. WDYT?

That's a nice idea, yeah!

falexwolf:
We need a test here for the new `.yaml` artifact loader:

https://github.com/laminlabs/lamindb/blob/996a41640d693cd5254170a65aa140ac5078f416/tests/core/test_artifact.py#L714

@Koncopd, do you want to finalize and rename the test function to `test_artifact_loaders()`?

Koncopd:
Do we want to add `pyyaml` to the deps?

falexwolf:
No, I'd not add pyyaml to the dependencies because we have too many anyway and nothing in lamindb requires it.

Koncopd:
Ok, i wanted to push some things here but again we have the problem with CI for PRs not from this repo.

Koncopd:
> Thanks, Robrecht! Great!
> 
> > In addition, we could add a data loader for .rds, which when called, produces an error message that you should use laminr for loading .rds files. WDYT?
> 
> That's a nice idea, yeah!

@rcannood @falexwolf i think a warning is more reasonable for that https://github.com/laminlabs/lamindb/pull/2188

lamindb/core/loaders.py
```diff
@@ -107,9 +108,17 @@ def load_json(path: UPathStr) -> dict:
         data = json.load(f)
     return data
 
+def load_yaml(path: UPathStr) -> dict:
+    """Load `.yaml` to `dict`."""
+    import yaml
```
Koncopd, line 113: Maybe except `ImportError` and log a warning that the package is not installed and return a path.

Koncopd:
`load_yaml` needs a test. 

There is also a comment from me, otherwise good to merge when the comments are resolved.

Koncopd:
This looks inactive, so moving it here https://github.com/laminlabs/lamindb/pull/2270

PR #2177, opened 2024-11-14
✨ Enable Django field validation and Literal validation

sunnyosun:
Enable broad validation of records via Django's [field validation](https://docs.djangoproject.com/en/5.1/ref/validators/) and, for enums, based on type annotation with `Literal`.

## Validating enums via `Literal`
- literal validation is done from type hints. This allows to validate, e.g., `Transform.type`, `GeneticTreatment.system`.
- this is enabled for all schemas.

```python
GeneticTreatmentSystem = Literal[
    "CRISPR Cas9",
    "CRISPRi",
    "CRISPRa",
    "shRNA",
    "siRNA",
    "transgene",
    "transient transfection",
]

class GeneticTreatment:
    system: GeneticTreatmentSystem = models.CharField()
```

```python
>>> import wetlab as wl
>>> wl.GeneticTreatment(name="new treatment", system="new-sys")
FieldValidationError: 
  system: new-sys is not a valid
    → Valid values are: CRISPR Cas9, CRISPRa, CRISPRi, shRNA, siRNA, transgene, transient transfection
```

## Django field validation
- calls [`.clean_fields()`](https://docs.djangoproject.com/en/5.1/ref/models/instances/#django.db.models.Model.clean_fields) during record init.
- introduced `lnschema_core.fields` to set defaults for django fields (e.g. CharField), all custom schema should import from `lnschema_core.fields` instead of `django.db.models`.
- enabled for models that inherit the `ValidateFields` class. In these model definition, make sure to import fields from `lnschema_core.fields` instead of `django.db.models`! If using `django.db.models`, please specify `blank=True`.
- added `faq:validate-fields`.

```python
from lnschema_core.models import ValidateFields
from lnschema_core.fields import CharField

class Reference(Record, ..., ValidateFields):
    ...

>>> import findrefs as fr
>>> fr.Reference(name="my ref", doi="abc.ef", url='myurl.com')
FieldValidationError: 
  url: myurl.com is not valid
    → Enter a valid URL.
  doi: abc.ef is not valid
    → Must be a DOI (e.g., 10.1000/xyz123 or https://doi.org/10.1000/xyz123)
```

## Performance of django validation
The django validation comes with performance cost, therefore it is only enabled for models that are inherited from the `ValidateFields` class.

Here is a test with inserting 70k records with or without django validation:
Screenshot 2024-11-15 at 10 55 18

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,7 +22,7 @@ jobs:
           - "faq"
           - "storage"
           - "cli"
-    timeout-minutes: 10
+    timeout-minutes: 15
 
     steps:
       - uses: actions/checkout@v4
--- a/.gitignore
+++ b/.gitignore
@@ -58,6 +58,7 @@ default_storage_unit_core
 default_storage_unit_storage
 test.ipynb
 run-tests
+test-django-validation/
 
 # General
 .DS_Store
--- a/docs/bio-registries.ipynb
+++ b/docs/bio-registries.ipynb
@@ -304,7 +304,7 @@
    "source": [
     "We'd like to load the corresponding records in our in-house registry to annotate a dataset.\n",
     "\n",
-    "To this end, you'll typically use {class}`~lamindb.core.CanValidate.from_values`, which will both validate & retrieve records that match the values."
+    "To this end, you'll typically use {class}`~lamindb.core.CanCurate.from_values`, which will both validate & retrieve records that match the values."
    ]
   },
   {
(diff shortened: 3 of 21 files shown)


falexwolf:
I'd rename `RecordValidationError` to `FieldValidationError`.

image

falexwolf:
@sunnyosun, I put in some effort to make these examples prettier and easier to understand. Do you want to backpart into this PR description so that this looks even better?

- https://github.com/laminlabs/lamin-docs/pull/183

PR #2184, opened 2024-11-18
📝 Move `tutorial` and `tutorial2` to `lamin-docs`

falexwolf:
This is to prepare integrating both tutorials and consolidating the content with what's in the introduction.

See:

- https://github.com/laminlabs/lamin-docs/pull/178


falexwolf:
> Looks like it is pretty bad for coverage.

I inspected this and think that's mostly a fluke. And if there are indeed some lines that I somehow forgot to cover elsewhere, more unit tests have to be filled in.

falexwolf:
Ok, I remember. There are indeed 9 lines in `_parents()` that need a more comprehensive lineage setup. This was so far somewhat implicitly tested with the notebooks. The much worse thing is that we don't even test the resulting plots, @sunnyosun. Maybe Claude can help us come up with a comprehensive testing framework for this; there is a ton of somewhat ad hoc logic in `_parents.py`. I admit that this was one of the pieces of most work in Scanpy to get the plotting testing right (with a big bulk of it done by Fidel).

image

Issue #2154, opened 2024-11-11
Can the Curator.validate() function be more efficient?


Zethson:
### Description of feature

I was just debugging something and noticed that the `curator.validate()` function seems to call `validate() very many times per passed categorical. I observed 10 calls to `validate()` per categorical. Are all of these necessary? Why are there so many calls?

Obviously this has a big performance penalty.

falexwolf:
Nice one: you used the "Feature" template for a UX/enhancement topic. 😇 😆

sunnyosun:
There's a single `inspect` call in the `validate_categories` function, if with `using_key`, there are 2 `inspect` calls, one on each instance.

can you point out where did you see 10 validate calls?

Zethson:
There's two performance issues for the Curator

https://github.com/laminlabs/lamindb/issues/2217

so I'll close mine

PR #2186, opened 2024-11-18
🗑️ Deprecate add_new_from_columns

sunnyosun:


falexwolf:
And is this a good time to look into?

- https://github.com/laminlabs/lamindb/issues/2217
- https://github.com/laminlabs/lamindb/issues/2154

sunnyosun:
Looking into performance now!

lamindb/_curate.py
```diff
@@ -265,28 +277,87 @@ def add_new_from(self, key: str, organism: str | None = None, **kwargs):
...
+        )
+        pass
 
-        Args:
-            organism: The organism name.
-            **kwargs: Additional keyword arguments to pass to the registry model.
+    def _replace_synonyms(
+        self, key: str, syn_mapper: dict, values: pd.Series | pd.Index
+    ):
+        # replace the values in df
```
Zethson, line 299: Not very useful comment because the next line should give it away and have a self explanatory variable name 
```suggestion
```

If anything, this comment should contain justification of what we're doing next.

lamindb/_curate.py
```diff
@@ -265,28 +277,87 @@ def add_new_from(self, key: str, organism: str | None = None, **kwargs):
...
+        std_values = values.map(lambda x: syn_mapper.get(x, x))
```
Zethson, line 300:
```suggestion
        std_values = values.map(lambda unstd_val: syn_mapper.get(unstd_val, unstd_val))
```

Makes it clearer that we default to unstandardized values if we don't find a match.

lamindb/_curate.py
```diff
@@ -265,28 +277,87 @@ def add_new_from(self, key: str, organism: str | None = None, **kwargs):
...
+        # remove the standardized values from self.non_validated
+        non_validated = [i for i in self.non_validated[key] if i not in syn_mapper]
```
Zethson, line 302:
```suggestion
        non_validated = list(set(self.non_validated[key]) - set(syn_mapper))
```

Probably more efficient like this?

lamindb/_curate.py
```diff
@@ -265,28 +277,87 @@ def add_new_from(self, key: str, organism: str | None = None, **kwargs):
...
+                [f'"{k}" → "{v}"' for k, v in syn_mapper.items()], sep=""
+            )
+            s = "s" if n > 1 else ""
+            logger.success(
+                f'standardized {n} synonym{s} in "{key}": {colors.green(syn_mapper_print)}'
+            )
+        return std_values
+
+    def standardize(self, key: str):
+        """Standardize the dataset.
```
Zethson, line 320: Not the most informative description because that's more or less what the function name already gives away.
```suggestion
        """Replace synonyms with standardized values.
```

Something like this.

lamindb/_curate.py
```diff
@@ -265,28 +277,87 @@ def add_new_from(self, key: str, organism: str | None = None, **kwargs):
...
+
+        Inplace modification of the input dataset.
```
Zethson, line 322:
```suggestion
        Modifies the input dataset inplace.
```

lamindb/_curate.py
```diff
@@ -265,28 +277,87 @@ def add_new_from(self, key: str, organism: str | None = None, **kwargs):
...
         """
-        self._kwargs.update({"organism": organism} if organism else {})
-        self._save_columns(validated_only=False, **self._kwargs, **kwargs)
+        avail_keys = list(self.non_validated.keys())
+        if len(avail_keys) == 0:
+            logger.warning("No need to standardize")
```
Zethson, line 326:
```suggestion
            logger.warning("values are already standardized.")
```

lamindb/_curate.py
```diff
@@ -265,28 +277,87 @@ def add_new_from(self, key: str, organism: str | None = None, **kwargs):
...
+            syn_mapper_print = _print_values(
+                [f'"{k}" → "{v}"' for k, v in syn_mapper.items()], sep=""
+            )
+            s = "s" if n > 1 else ""
+            logger.success(
+                f'standardized {n} synonym{s} in "{key}": {colors.green(syn_mapper_print)}'
+            )
+        return std_values
+
+    def standardize(self, key: str):
```
Zethson, line 319: Please document the `key` parameter.

Zethson: Why is `key` not an `Iterable`? I get it - "all" is special but what if users want to pass several keys?

lamindb/_curate.py
```diff
@@ -265,28 +277,87 @@ def add_new_from(self, key: str, organism: str | None = None, **kwargs):
...
+        """Standardize the dataset.
+
+        Inplace modification of the input dataset.
         """
-        self._kwargs.update({"organism": organism} if organism else {})
-        self._save_columns(validated_only=False, **self._kwargs, **kwargs)
+        avail_keys = list(self.non_validated.keys())
```
Zethson, line 324: This list construction might not be necessary because `self.non_validated.keys()` is already iterable. We're constructing this list every time we call `standardize` resulting in a very minor performance penalty.

lamindb/_curate.py
```diff
@@ -311,6 +385,8 @@ def _update_registry_all(self, validated_only: bool = True, **kwargs):
     def validate(self, organism: str | None = None) -> bool:
         """Validate variables and categorical observations.
 
+        This method also registers the validated records in the current instance.
```
Zethson, line 388: Might need 1 more sentence of context when that happens? Validated against external instances or public ontologies? Else, they'd already be registered in the current instance, right?

lamindb/_curate.py
```diff
@@ -400,14 +472,15 @@ class AnnDataCurator(DataFrameCurator):
...
     Args:
-        data: The AnnData object or an AnnData-like path.
-        var_index: The registry field for mapping the ``.var`` index.
-        categoricals: A dictionary mapping ``.obs.columns`` to a registry field.
-        using_key: A reference LaminDB instance.
-        verbosity: The verbosity level.
-        organism: The organism name.
-        sources: A dictionary mapping ``.obs.columns`` to Source records.
-        exclude: A dictionary mapping column names to values to exclude.
+        data: `ad.AnnData | UPathStr` The AnnData object or an AnnData-like path.
```
Zethson, line 475: See above - not sure why we need type hints?

lamindb/_curate.py
```diff
@@ -529,14 +600,16 @@ def add_new_from_var_index(self, organism: str | None = None, **kwargs):
 
         Args:
             organism: The organism name.
-            **kwargs: Additional keyword arguments to pass to the registry model.
+            **kwargs: Additional keyword arguments to pass to create new records.
```
Zethson, line 603: See above.

lamindb/_curate.py
```diff
@@ -529,14 +600,16 @@ def add_new_from_var_index(self, organism: str | None = None, **kwargs):
...
         """
         self._kwargs.update({"organism": organism} if organism else {})
         self._save_from_var_index(validated_only=False, **self._kwargs, **kwargs)
 
     def validate(self, organism: str | None = None) -> bool:
         """Validate categories.
 
+        This method also registers the validated records in the current instance.
```
Zethson, line 611: See above.

lamindb/_curate.py
```diff
@@ -576,6 +649,27 @@ def validate(self, organism: str | None = None) -> bool:
         self._validated = validated_var and validated_obs
         return self._validated
 
+    def standardize(self, key: str):
```
Zethson, line 652: See above for several comments.

lamindb/_curate.py
```diff
@@ -631,17 +722,17 @@ class MuDataCurator:
     the object should be recreated using :meth:`~lamindb.Curator.from_mudata`.
 
     Args:
-        mdata: The MuData object to curate.
-        var_index: The registry field for mapping the ``.var`` index for each modality.
+        mdata: `MuData` The MuData object to curate.
```
Zethson, line 725: See above for typehint comment.

lamindb/_curate.py
```diff
@@ -807,21 +886,21 @@ def add_new_from_var_index(
         Args:
             modality: The modality name.
             organism: The organism name.
-            **kwargs: Additional keyword arguments to pass to the registry model.
+            **kwargs: Additional keyword arguments to pass to create new records.
```
Zethson, line 889: See above.

lamindb/_curate.py
```diff
@@ -836,15 +915,17 @@ def add_new_from(
             key: The key referencing the slot in the DataFrame.
             modality: The modality name.
             organism: The organism name.
-            **kwargs: Additional keyword arguments to pass to the registry model.
+            **kwargs: Additional keyword arguments to pass to create new records.
```
Zethson, line 918: See above.

lamindb/_curate.py
```diff
@@ -864,49 +945,38 @@ def validate(self, organism: str | None = None) -> bool:
...
+            logger.info(f'validating categoricals in modality "{modality}"...')
+            mods_validated &= adata_curator.validate(**self._kwargs)
+            if len(adata_curator.non_validated) > 0:
+                self._non_validated[modality] = adata_curator.non_validated  # type: ignore
+            logger.print("")
+
+        self._validated = obs_validated & mods_validated
         return self._validated
 
+    def standardize(self, key: str, modality: str | None = None):
```
Zethson, line 968: See above for key and docstring comments

lamindb/_curate.py
```diff
@@ -1158,7 +1212,7 @@ def validate_categories(
         source: The source record.
         exclude: Exclude specific values from validation.
         standardize: Whether to standardize the values.
-        validated_hint_print: The hint to print for validated values.
+        hint_print: The hint to print to fix non-validated values.
```
Zethson, line 1215:
```suggestion
        hint_print: The hint to print that suggests fixing non-validated values.
```

lamindb/_curate.py
```diff
@@ -1257,6 +1308,43 @@ def _log_mapping_info():
         return False, non_validated
 
 
+def standardize_categories(
+    values: Iterable[str],
+    field: FieldAttr,
+    using_key: str | None = None,
+    organism: str | None = None,
+    source: Record | None = None,
+):
```
Zethson, line 1317: typehint return please and consider adding a docstring. This is not an underscore function and has many lines of code. It'd deserver at least a 1 liner.

Zethson (changes requested):
Great!

1. We are losing about 1.5% coverage with this PR. That's a bit much.
2. I can't properly comment on the notebooks so I'll make some messy comments here.
In `curate-df`:
> In LaminDB, validation means verifying that values exist in metadata registries. Validation criteria define which registry fields should contain these values.

Technically correct but if I were to read this for the first time, it would sound a bit lose to me. I would feel more comfortable if it were clearer that in many cases these "values that exist in metadata registries" are based on public ontologies.

I'm not sure how useful distinguishing between "__Validated label & feature identifier__" and "validated record" is. At least the description didn't give much more away to me.
3. The `ux` of `standardize()` is a bit weird to me. I'd consider defaulting the `key` to `all`. I'd also consider supporting `Iterable`. And should it be called `key`? `key` is programmers language and maybe we should consider something more meaningful to users.

I made a commit with some minor changes to not bug you with nitpicks.

sunnyosun:
> We are losing about 1.5% coverage with this PR. That's a bit much.
No, I added a lot more tests, we in fact gained 0.5% coverage, the codecov report up there didn't update, see here: https://app.codecov.io/gh/laminlabs/lamindb/pull/2186?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=laminlabs

sunnyosun:
> 3. The ux of standardize() is a bit weird to me. I'd consider defaulting the key to all. I'd also consider supporting Iterable. And should it be called key? key is programmers language and maybe we should consider something more meaningful to users.

Thought about it, but I think sometimes user wants to standardize in other ways. It's a mutating operation to their dataset, so I want to be careful.

Calling a for loop for `.standardize` is super easy; I'd not support that to minimize our maintenance.

`key` is just a generic word I can come up with that can cover all different formats (obs columns, var.index, mudata stuff), I'd like to be more specific if you have a better name.

Zethson:
I reverted my accidental submodule push with that short commit @sunnyosun :) I agree ofc.

tests/core/test_curate.py
```diff
@@ -98,17 +97,45 @@ def mock_transform():
...
+        _ = curator.non_validated
     validated = curator.validate()
+    assert curator.non_validated == {
+        "cell_type": ["cerebral pyramidal neuron", "astrocytic glia"],
+        "donor": ["D0001", "D0002", "DOOO3"],
+    }
     assert validated is False
 
+    # deprecated method
+    curator.add_new_from_columns()
```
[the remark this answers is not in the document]
sunnyosun: it's a deprecated method doesn't do anything but needs to be covered, what do you suggest to do?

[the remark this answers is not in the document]
Zethson: I guess a `pragma: no cover` is good enough for that function assuming it calls the non-deprecated one that is covered.

If only `pragma: no cover` worked as expected sigh.

tests/core/test_curate.py
```diff
@@ -215,13 +243,49 @@ def test_clean_up_failed_runs():
 
 @pytest.mark.parametrize("to_add", ["donor", "all"])
 def test_anndata_curator(adata, categoricals, to_add):
+    adata = adata.copy()
+    # must pass an organism
+    with pytest.raises(ValidationError):
+        bt.settings._organism = None  # make sure organism is not set globally
```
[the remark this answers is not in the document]
sunnyosun: This involves a lot of changes, will do in a different PR.

tests/core/test_curate.py
```diff
@@ -327,7 +396,41 @@ def test_mudata_curator(mdata):
...
+        },
+        "rna": {
+            "cell_type": ["astrocytic glia"],
+            "donor": ["D0001", "D0002", "DOOO3"],
+            "var_index": ["TCF-1"],
+        },
+    }
+
+    # lookup
+    _ = curator.lookup()
```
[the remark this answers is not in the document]
sunnyosun: I think calling it to make sure it doesn't error is enough.

tests/core/test_curate.py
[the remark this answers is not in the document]
Zethson: Why? Isn't it something like:

```
old_organism = bt.settings._organism
bt.settings._organism = None
# more code
bt.settings._organism = old_organism
```

?

lamindb/_curate.py
```diff
@@ -165,24 +182,21 @@ def __init__(
         self._fields = categoricals or {}
         self._columns_field = columns
         self._using_key = using_key
+        # TODO: change verbosity back
```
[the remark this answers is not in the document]
sunnyosun: No, it's not easy, because there's no clear end step for Curator. I need to think more here.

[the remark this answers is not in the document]
sunnyosun: Maybe I'll just change back after `.validate` and `save_artifact`

lamindb/_curate.py
```diff
@@ -222,9 +234,12 @@ def _check_valid_keys(self, extra: set = None) -> None:
                 raise TypeError(f"{name} must be a dictionary!")
             valid_keys = set(self._df.columns) | {"columns"} | extra
             nonval_keys = [key for key in d.keys() if key not in valid_keys]
+            n = len(nonval_keys)
```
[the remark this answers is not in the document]
sunnyosun: I think it's more consistent to show a number here.

lamindb/_curate.py
```diff
@@ -265,28 +277,87 @@ def add_new_from(self, key: str, organism: str | None = None, **kwargs):
         Args:
             key: The key referencing the slot in the DataFrame from which to draw terms.
             organism: The organism name.
-            **kwargs: Additional keyword arguments to pass to the registry model.
+            **kwargs: Additional keyword arguments to pass to create new records
```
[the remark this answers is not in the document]
sunnyosun: What do you suggest? It's needed for populating things like `add_new_from("perturbation", system="CRISPR-Cas9")`, otherwise it will error as system is required in `wl.GeneticPerturbation`.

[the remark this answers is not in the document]
Zethson: Hmm they're more or less just for developers and not for users. So we could remove the docstring or label them for internal use only or just leave it as you have it.

P.S. I find it funny that you intuitively used `GeneticPerturbation` when it's called `GeneticTreatment`. Maybe we should just rename them...

falexwolf:
I've been confused about the choice `Treatment` for years meanwhile. It'd be great to have a paragraph somewhere that explains why that chose versus something else. Some people also prefer "intervention" but I think that's less common in biology.

lamindb/_curate.py
sunnyosun: I'm just seeing how it's done here: https://docs.lamin.ai/lamindb.artifact

lamindb/_curate.py
sunnyosun: It's consistent with `key` for `add_new_from`, for such important operations, it's better users know what they are doing. `all` is already a convenient thing I added, don't think iterable is needed here, `key` also indicate it's singular.

Zethson:
> Thought about it, but I think sometimes user wants to standardize in other ways. It's a mutating operation to their dataset, so I want to be careful.
> 
> Calling a for loop for .standardize is super easy; I'd not support that to minimize our maintenance.

Well, `all` is equally as dangerous and we are supporting that. It is easy to call a loop, but it breaks the UX a bit because when using the `curator` users usually don't write loops for `validate()` either. It's very cheap for us to support `Iterable` and I don't think that it adds a lot of maintenance burden? It should be easy to implement and test.

> key is just a generic word I can come up with that can cover all different formats (obs columns, var.index, mudata stuff), I'd like to be more specific if you have a better name.

Yeah, that's a lot...maybe:
1.  `values_key`?
2. `categorical_key`? - for consistency with `categoricals` but it would not cover `var_index`...
3. `col`? `var_index` is also a column.
4. `col_key`

tests/core/test_curate.py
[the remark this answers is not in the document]
sunnyosun: Because there's this one: https://github.com/laminlabs/lamindb/blob/2e298635f61ee745acf41fb66b3d711eb79c1726/tests/core/test_artifact.py#L45

I need to go through all the test functions to see which ones needs it.

Zethson:
> But if it's called `key` or `values_key` or anything, it implies it's a singular, making it Iterable is very confusing. At least I always find thing confusing if they accept a sigular and a list at the same time. I think having `all` is enough to make the UX you described above, an Iterable is overkill and even more confusing to name.
> 
> Yes "all" is equally dangerous, that's why I'm not making it a default and we should not advertise using it unless necessary in all our guides.

I think behavior that I've seen before is that `None` means all. In other words, if we didn't have the `all` special parameter and rather defaulted to `None` which means all of them, we could more easily enable Iterable.

falexwolf:
There is a lot of stuff going on here and I'll not read through all code and all comments unless you ask me to.

So, I'll just focus on getting the "what the hell does validation even mean?" question right. The first 5 min in which the user gets confronted with the topic "Curating datasets".

Before | After
--- | ---
image | image
image | image

I noticed that sometimes we say "validation criteria" and sometimes "validation constraint". We should only use one of them. Which?

Zethson:
> Again, I think users should not do things blindly when it comes to add_new_from and standardize.

I agree with `add_new_from` 100%. But `standardize()` should be fine. Yes, it's inplace but it only matches values that are in the instance or the public ontologies. I don't see the big risks here.

We can gather some feedback from our users first and then still act later if necessary.

sunnyosun:
> > Again, I think users should not do things blindly when it comes to add_new_from and standardize.
> 
> I agree with `add_new_from` 100%. But `standardize()` should be fine. Yes, it's inplace but it only matches values that are in the instance or the public ontologies. I don't see the big risks here.
> 
> We can gather some feedback from our users first and then still act later if necessary.

For example, Felix's issue with some numbers in synonyms will be a disaster if applied blindly. So I think users still need to be careful with `standardize`.

PR #2189, opened 2024-11-19
🎨 Prepare release

sunnyosun:


Koncopd:
Please don't release yet, the new error for nonexisting buckets has important consequences for `lamindb-setup` sync code.

falexwolf:
> Please don't release yet, the new error for nonexisting buckets has important consequences for lamindb-setup sync code.

Do we have to start upper bound botocore; are all lamindb-setup releases right now broken because they decided to throw another error? Are we using such low level logic that we're prone to such a change?

Koncopd:
@falexwolf we surely have, i also see the new errors when checking manually, without installing anything new or reinstalling. It looks like s3 service itself began to issue new errors in some cases for some reason, i can't really understand why for now. For some paths it gives `PermissionError`, for some `FileNotFound`.

falexwolf:
> @falexwolf we surely have proper bounds, i also see the new errors when checking manually, without installing anything new or reinstalling. It looks like s3 service itself began to issue new errors in some cases for some reason, i can't really understand why for now. For some paths it gives PermissionError, for some FileNotFound.

Understood. Then let's sleep on this and investigate tomorrow.

This seems a bigger issue. If S3 broke the API that'd be pretty crazy, so I hope there some temporary fluke there.

Koncopd:
It doesn't really break anything, everything continues working, just some code here https://github.com/laminlabs/lamindb-setup/blob/ab825efcfd48361e0a330cc5384dc9c581c99c8e/lamindb_setup/core/upath.py#L366 becomes problematic for a rare edge case when trying to sync non-existing cloud object.

Koncopd:
So these 2 paths started giving `PermissionError` now for some reason:
```
s3://lndb-setup-ci/pbmc68k.h5ad
s3://inexistent-bucket-239809834/pbmc68k.h5ad
```
Just random non-existing buckets still result in `FileNotFoundError`.
very strange

Koncopd:
No, i don't see these buckets. I suspect s3 enabled some kind of cache on their side for frequently accessed buckets. In this case it doesn't really matter, because the normal case of non-existent file in an existing bucket is still fine. I have edited the upload notebook a bit to explain this new behavior and also check for non-existent file in an existing bucket. I don't think we have any problems with this because now we know what to expect in some cases. Frequently accessed non-existent bucket is super edge case and even that doesn't really break anything for us i believe.

falexwolf:
Great! 👌

PR #2210, opened 2024-11-25
📝 Prettify gene symbols guide & logging

falexwolf:
Before:
```
! Curating gene symbols is discouraged. See FAQ for more details.
```

After:
```
! indexing datasets with gene symbols can be problematic: https://docs.lamin.ai/faq/symbol-mapping
```

I have a case where I consciously want to use gene symbols for the sake of illustrating something; not for the sake of curating a large body of data.

Hearing some feedback from some users this seems to be very common. So, I'd not make the warning as harsh.

I was missing the URL so that I could read up on this.

Made the warning lowercase consistent with all other logging messages (including warnings).

Replace hanging indentation with vertically aligned indendation.

See: 

- https://github.com/laminlabs/lamindb/issues/2211

### DIFF
--- a/docs/faq/symbol-mapping.ipynb
+++ b/docs/faq/symbol-mapping.ipynb
@@ -4,14 +4,14 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "# Why should I validate against gene symbols?"
+    "# Why should I not index datasets with gene symbols?"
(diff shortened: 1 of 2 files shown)


Zethson:
Thank you!

Zethson (approved):

Issue #2039, opened 2024-10-10
✨ Support `datetime` as a feature `dtype`


falexwolf:
```python
ln.Feature(name='publish_date', dtype='datetime').save()
```
leads to the error
```python
ValueError: dtype is datetime but has to be one of 'number', 'int', 'float', 'cat', 'bool', 'cat[...]'!
```

It turns out we're also not supporting it for feature values: https://lamin.ai/laminlabs/lamindata/transform/U2T805mjF6TL0000

Internal discussion: https://laminlabs.slack.com/archives/C062BMXLHRN/p1728554457256399

PR #2218, opened 2024-11-26
✨ Improve annotating with, retrieving & removing feature values

falexwolf:
Note: More comprehensive examples will be added in a follow up PR that will prettify visualizations of features. This PR and the following will be announced together.

Improved annotating artifacts with feature values:

1. support date- and datetime-like feature values with new `dtype` for `Feature`: `"date"` and `"datetime"`
2. clearer feature overview that separates internal from external features
3. enable removing feature value annotations via `artifact.features.remove_values()`
4. consistent handling of annotated values via sets as they come from aggregations, before they were treated as lists
5. refactor the code to simplify and make more maintainable, performance issue still persists though

On the last 3rd point, see https://lamin.ai/laminlabs/lamindata/transform/BblTiuKxsb2g0002

Resolves:

- https://github.com/laminlabs/lamindb/issues/2039

### DIFF
--- a/lamindb/_feature.py
+++ b/lamindb/_feature.py
@@ -21,6 +21,8 @@ FEATURE_TYPES = {
     "int": "int",
     "float": "float",
     "bool": "bool",
+    "date": "date",
+    "datetime": "datetime",
     "str": "cat",
     "object": "cat",
 }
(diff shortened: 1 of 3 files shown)


falexwolf:
Claude [helped me](https://claude.ai/share/692dd602-db3c-4825-8d58-c0bd49d26a33) with this commit, @sunnyosun 

- https://github.com/laminlabs/lamindb/pull/2218/commits/6cb05af458679b4b9e5a49d756f564f64ed65a39

I'm surprised that there is so much room for simplification. 🤔 

Do you spot something shady?

I was just asking it to separate the message building logic from the data gathering because evidently the former doesn't depend on the SQL dialect.

PR #2228, opened 2024-11-28
✨ Add Curator for tiledbsoma stores

Koncopd:


lamindb/_curate.py
```diff
@@ -1058,6 +1062,490 @@ def save_artifact(
...
+    if (n := len(nonval_keys)) > 0:
+        s = "s" if n > 1 else ""
+        are = "are" if n > 1 else "is"
+        raise ValidationError(
+            f"the following {n} key{s} passed to {name} {are} not allowed: {colors.yellow(_print_values(nonval_keys))}"
+        )
+
+
+class SOMACurator(BaseCurator):
+    """Curation flow for ``tiledbsoma``.
```
falexwolf, line 1075: Inconsistent use of double backticks and single backticks. You can use single backticks through out.

lamindb/_curate.py
```diff
@@ -1058,6 +1062,490 @@ def save_artifact(
...
+
+    See also :class:`~lamindb.Curator`.
+
+    Args:
+        experiment_uri: The uri of a `tiledbsoma.Experiment`.
```
falexwolf, line 1080: What's is a URI? Please explain.

lamindb/_curate.py
```diff
@@ -1058,6 +1062,490 @@ def save_artifact(
...
+        var_index: The registry fields for mapping the ``.var`` indices for measurements.
+            Should be in the form `{"measurement name": ("var column", field)}`.
+            These keys should be used in the flattened form ('{measurement name}__{column name in .var}')
```
falexwolf, line 1083: The should be backticks around the expression so that it renders as code: 
image

lamindb/_curate.py
```diff
@@ -1058,6 +1062,490 @@ def save_artifact(
...
+            in `.standardize` or `.add_new_from`, see the output of `.var_index`.
+        categoricals: A dictionary mapping ``.obs`` columns to a registry field.
```
falexwolf, line 1085: This is only for categorical obs columns, not for any others.

lamindb/_curate.py
```diff
@@ -1058,6 +1062,490 @@ def save_artifact(
...
+        obs_columns: The registry field for mapping the ``.obs`` columns.
```
falexwolf, line 1086: for mapping `.obs.columns`. It's otherwise confusing with `categoricals`, which is about the `.obs` columns content and where you say "mapping the `.obs` columns to a registry field".

lamindb/_curate.py
```diff
@@ -1058,6 +1062,490 @@ def save_artifact(
...
+        using_key: A reference LaminDB instance.
```
falexwolf, line 1087: let's please not expose `using_key` for any new Curator class. It should disappear from all.

(Unless you tell me we absolutely need this @Zethson @sunnyosun, I continue to think that this is a horribly complicated design choice and a very bad name for an argument. I'll make a new GitHub issue on this topic or try to find where we previously discussed this).

lamindb/_curate.py
```diff
@@ -1058,6 +1062,490 @@ def save_artifact(
...
+        organism: The organism name.
+        sources: A dictionary mapping ``.obs`` columns to Source records.
+        exclude: A dictionary mapping column names to values to exclude.
```
falexwolf, line 1090: Which column names? Exclude from what?

(I guess I'm touching on broader questions in API design for Curator but given this is a new interface we should keep it as minimal as possible to just enable minimal functionality rather than propagating bad design choices.)

lamindb/_curate.py
```diff
@@ -1058,6 +1062,490 @@ def save_artifact(
...
+
+    Examples:
+        >>> import bionty as bt
+        >>> curate = ln.Curator.from_tiledbsoma(
+        ...     "store.tiledbsoma",
```
falexwolf, line 1095: Can you choose another name that indicates that this is a path and is user-provided? Maybe the following:
```
./my_array_store.tiledbsoma
```

lamindb/_curate.py
```diff
@@ -1058,6 +1062,490 @@ def save_artifact(
...
+        self._non_validated_values: dict[str, list] | None = None
+        self._validated_values: dict[str, list] = {}
+        # filled by _check_save_keys
+        self._n_obs: int | None = None
+        self._valid_obs_keys: list[str] | None = None
+        self._valid_var_keys: list[str] | None = None
+        self._var_fields_flat: dict[str, FieldAttr] | None = None
+        self._check_save_keys()
+
+    def _check_save_keys(self):
```
falexwolf, line 1135: What does this do? A high-level comment?

falexwolf (approved):
This is good from my end. The bigger questions are for @sunnyosun and @Zethson about conventions for the user-facing API design and code organizations.

This is another thread.

@sunnyosun should also review this PR though.

lamindb/_curate.py
```diff
@@ -1058,6 +1062,490 @@ def save_artifact(
...
+    def _non_validated_values_field(self, key: str) -> tuple[list, FieldAttr]:
+        assert self._non_validated_values is not None  # noqa: S101
+
+        if key in self._valid_obs_keys:
+            field = self._obs_fields[key]
+        elif key in self._valid_var_keys:
+            ms = key.partition("__")[0]
+            field = self._var_fields[ms][1]
+        else:
+            raise KeyError(f"key {key} is invalid!")
```
Zethson, line 1305: This is not covered and could be tested?

lamindb/_curate.py
```diff
@@ -1058,6 +1062,490 @@ def save_artifact(
...
+                if len(non_val) > 0:
+                    validated = False
+                    self._non_validated_values[key] = non_val
+                else:
+                    self._validated_values[key] = values
+        self._validated = validated
+        return self._validated
+
+    def _non_validated_values_field(self, key: str) -> tuple[list, FieldAttr]:
+        assert self._non_validated_values is not None  # noqa: S101
```
Zethson, line 1297: Why do we need an assert here and can't raise a proper error?

lamindb/_curate.py
```diff
@@ -1058,6 +1062,490 @@ def save_artifact(
...
+            field = self._obs_fields[key]
+        elif key in self._valid_var_keys:
+            ms = key.partition("__")[0]
+            field = self._var_fields[ms][1]
+        else:
+            raise KeyError(f"key {key} is invalid!")
+        values = self._non_validated_values.get(key, [])
+        return values, field
+
+    def add_new_from(self, key: str):
```
Zethson, line 1309:
```suggestion
    def add_new_from(self, key: str) -> None:
```

lamindb/_curate.py
sunnyosun: Agreed, let's hide it.

lamindb/_curate.py
```diff
@@ -1058,6 +1062,490 @@ def save_artifact(
         return self._artifact
 
 
+def _maybe_validation_error(nonval_keys: list[str], name: str):
```
Zethson, line 1065:
```suggestion
def _maybe_validation_error(nonval_keys: list[str], name: str) -> None:
```

I don't really like the name of this function because "validation" is extremely generic and overloaded heavily in this context. It's more like "maybe_curation_keys_error" or something like this.

lamindb/_curate.py
```diff
@@ -1058,6 +1062,490 @@ def save_artifact(
...
+                field=self._columns_field,
+                organism=organism,
+                raise_validation_error=False,
+            )
+        for ms in self._var_fields:
+            var_key, var_field = self._var_fields[ms]
+            organism = check_registry_organism(
+                var_field.field.model, self._organism
+            ).get("organism")
+            # todo: add type
```
Zethson, line 1513: Is this still open?

noxfile.py
```diff
@@ -86,10 +86,11 @@ def install_ci(session, group):
     if group == "unit-core":
         extras += "bionty,aws,gcp,zarr,fcs,jupyter"
         run(session, "uv pip install --system huggingface_hub")
+        run(
+            session, "uv pip install --system tiledbsoma==1.15.0rc3"
```
Zethson, line 90: Please add a note why you're pinning this version.

Zethson (approved):
Awesome! I recognized quite many parts of the code and we can look out for ways to consolidate them in the future I hope.

lamindb/_curate.py
Zethson: I think it's useful to extend the CxG curator. By default, it uses the `laminlabs/cellxgene` instance but when extending it (e.g. with perturbations) you want to now not curate against `laminlabs/cellxgene` but your own instance.

sunnyosun (approved):

lamindb/_curate.py
falexwolf: Because then we'd need to test raising it and it's an internal function? We agreed on using asserts for that and are doing this everywhere in the code base for internal consistency checks.

Zethson: Ah yeah right - I think I forgot about that. We discussed it.

tests/core/test_curator.py
```diff
@@ -441,3 +444,154 @@ def test_mudata_curator(mdata):
     bt.CellType.filter().delete()
     ln.FeatureSet.filter().delete()
     bt.Gene.filter().delete()
+
+
+def test_soma_curator(adata, categoricals):
+    tiledbsoma.io.from_anndata("curate.tiledbsoma", adata, measurement_name="RNA")
+
+    with pytest.raises(ValidationError) as error:
```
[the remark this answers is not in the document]
falexwolf: I've been using `assert err.exconly() == "..."` to test error messages for years now and if something like the above exists that'd allow simplifying many tests 😅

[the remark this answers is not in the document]
Zethson: See https://github.com/laminlabs/lamindb/blob/1b8f66dbd7e248368d11960f5da8d47decc01bf1/tests/core/test_curator.py#L313 where I used it.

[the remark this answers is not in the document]
falexwolf: Nice! 😄

lamindb/_curate.py
Koncopd: fixed

lamindb/_curate.py
Koncopd: done

lamindb/_curate.py
Koncopd: fixed

lamindb/_curate.py
Koncopd: fixed

lamindb/_curate.py
Koncopd: done

lamindb/_curate.py
Koncopd: Added the comment.

lamindb/_curate.py
Koncopd: done

lamindb/_curate.py
Koncopd: thx, added

noxfile.py
Koncopd: Added.

tests/core/test_curator.py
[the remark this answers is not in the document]
Koncopd: Thx, using this.

lamindb/_curate.py
Zethson: This is a hack for `cellxgene` and I've already complained many times that we need it. We need a way to exclude values from being validated against that are benign "default" values.

Example:

CxG pins ontology A against version B. However, if the corresponding column is not present in the object, we need to add it with default values such as "unknown". These values such as "unknown" would then lead to `validation` failures though because they are not found in ontology A of version B. The `exclude` parameter allows us to exclude specific values from validation.

We can pick the conversation up on slack again because I consider this a hack and not a solution. For now, we need it.

Zethson: @Koncopd could you slightly improve all "exclude docstrings" given my explanation (if it was clear enough) in all places where it's used, please?

Koncopd: Could you just give me a docstring to paste?

lamindb/_curate.py
```diff
@@ -1058,6 +1062,491 @@ def save_artifact(
...
+        experiment_uri: A local or cloud path to a `tiledbsoma.Experiment`.
+        var_index: The registry fields for mapping the `.var` indices for measurements.
+            Should be in the form `{"measurement name": ("var column", field)}`.
+            These keys should be used in the flattened form (`'{measurement name}__{column name in .var}'`)
+            in `.standardize` or `.add_new_from`, see the output of `.var_index`.
+        categoricals: A dictionary mapping categorical `.obs` columns to a registry field.
+        obs_columns: The registry field for mapping the names of the `.obs` columns.
+        organism: The organism name.
+        sources: A dictionary mapping `.obs` columns to Source records.
+        exclude: A dictionary mapping column names to values to exclude.
```
Zethson, line 1089:
```suggestion
        exclude: A dictionary mapping column names to values to exclude from validation.
            When specific :class:~bionty.Source instances are pinned and may lack default values (e.g., "unknown" or "na"), using the exclude parameter ensures they are not validated.
```

@Koncopd

lamindb/_curate.py
```diff
@@ -1058,6 +1062,491 @@ def save_artifact(
...
+
+    Examples:
+        >>> import bionty as bt
+        >>> curate = ln.Curator.from_tiledbsoma(
```
Zethson, line 1093:
```suggestion
        >>> curator = ln.Curator.from_tiledbsoma(
```

Thought we're using that everywhere.

lamindb/_curate.py
```diff
@@ -1058,6 +1062,491 @@ def save_artifact(
...
+                if "public", the lookup is performed on the public reference.
+        """
+        return CurateLookup(
+            categoricals=self._obs_fields,
+            slots={"columns": self._columns_field, **self._var_fields_flat},
+            using_key=using_key or self._using_key,
+            public=public,
+        )
+
+    def standardize(self, key: str):
```
Zethson, line 1385:
```suggestion
    def standardize(self, key: str) -> None:
```

lamindb/_curate.py
```diff
@@ -1058,6 +1062,491 @@ def save_artifact(
...
+                field=self._columns_field,
+                key="columns",
+                using_key=self._using_key,
+                validated_only=True,
+                organism=organism,
+                source=self._sources.get("columns"),
+                exclude=self._exclude.get("columns"),
+            )
+
+    def validate(self):
```
Zethson, line 1215:
```suggestion
    def validate(self) -> None:
```

lamindb/_curate.py
Koncopd: fixed

lamindb/_curate.py
Koncopd: added

PR #2236, opened 2024-11-29
💄 Create a more intuitive `artifact.describe()`

sunnyosun:
Before | After
--- | ---
`small_dataset1: AnnData` | https://lamin.ai/laminlabs/lamindata/artifact/mKo82UispVElpGtP0000
Screenshot 2024-11-29 at 21 08 42 | Screenshot 2024-12-01 at 22 47 53
`Papalexi21: MuData` | https://lamin.ai/laminlabs/lamindata/artifact/yjmQ0s0ekreNlMIA0000
Screenshot 2024-11-30 at 00 22 09 | Screenshot 2024-12-01 at 22 53 05Screenshot 2024-12-01 at 22 21 41Screenshot 2024-12-01 at 22 54 20
Screenshot 2024-12-01 at 21 58 27 | Screenshot 2024-12-01 at 21 54 52
Screenshot 2024-12-01 at 21 59 41 | Screenshot 2024-12-01 at 22 00 20


falexwolf:
We can't put the suffix as the heading because there are cases in which there is no suffix.

And we want to draw away attention from nuances of whether an AnnData is stored as zarr or parquet or a DataFrame as parquet or ipc etc.

Using  a combination of `_accessor` and `suffix` is better; but also this can sometimes both be not defined; e.g. for custom folders.

falexwolf:
It might be helpful to call out "Feature sets" somewhere on the representation, maybe next to "Dataset". 🤔 

Re the term "Dataset" for the materialized data on S3: It's probably a pretty good way of saying it. Alternatives could be "Schema" or "Storage schema" or "Schema in storage" or "Dataset schema". 🤔

While feature sets are a technical concept, they are key to understanding how datasets can be queried. Eliminating the term entirely from `.describe()` and GUI Detail Page seems dangerous.

falexwolf:
I think this is a bit confusing

image

How about "Dataset / .feature_sets" instead? Dim styling is good.

falexwolf (approved):

PR #2241, opened 2024-12-02
✅ More tests for Artifact.df()

falexwolf:
FYI @sunnyosun @Zethson @Koncopd -- there is a helper function here to compare df equality. Given it's often a downstream format for us, we'll likely need it in many places.

### DIFF
--- a/tests/core/test_describe_df.py
+++ b/tests/core/test_describe_df.py
@@ -12,7 +12,9 @@ def check_df_equality(actual_df: pd.DataFrame, expected_df: pd.DataFrame):
 
     Special handling for columns containing sets and NaN values.
     """
-    pd.testing.assert_index_equal(actual_df.index, expected_df.index)
+    # do not test indices by default
+    # pd.testing.assert_index_equal(actual_df.index, expected_df.index)
+    expected_df.index = actual_df.index
     assert set(actual_df.columns) == set(expected_df.columns)
     for col in expected_df.columns:
         # Detect if column contains sets by checking first non-null value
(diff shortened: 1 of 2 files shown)


tests/core/test_describe_df.py
```diff
@@ -1,14 +1,69 @@
...
+from lamindb.core import datasets
 from lamindb.core._data import _describe_postgres
-from lamindb.core.datasets import small_dataset1
 
 
-def test_curate_annotate_df():
-    ## Define the schema of the dataset & its metadata
+def check_df_equality(actual_df, expected_df):
+    """
+    Checks equality between two DataFrames, with special handling for columns containing sets
```
Zethson, line 12: Nit: Move 1 line up.

tests/core/test_describe_df.py
```diff
@@ -1,14 +1,69 @@
...
+    and NaN values.
 
+    Args:
+        actual_df (pd.DataFrame): The DataFrame to test
```
Zethson, line 16:
```suggestion
        actual_df: The DataFrame to test
```

tests/core/test_describe_df.py
```diff
@@ -1,14 +1,69 @@
...
+        expected_df (pd.DataFrame): The DataFrame with expected values
```
Zethson, line 17:
```suggestion
        expected_df: The DataFrame with expected values
```

tests/core/test_describe_df.py
```diff
@@ -1,14 +1,69 @@
...
+
+    Returns:
+        bool: True if DataFrames are equal, raises AssertionError otherwise
```
Zethson, line 20:
```suggestion
        True if DataFrames are equal, raises AssertionError otherwise
```

tests/core/test_describe_df.py
```diff
@@ -1,14 +1,69 @@
...
+    """
+    # Check index
```
Zethson, line 22: This and below has lots of comments that are of questionable use

Zethson (commented):
Since it's not in draft mode I allow myself to nag a bit :) Yes, Claude produced this. But let's clean it up.

falexwolf:
Addressed your comments, @Zethson!

PR #2250, opened 2024-12-03
🐛 Fix IPython import error

Koncopd:
Do not import `IPython` directly in `.view.py`

### DIFF
--- a/lamindb/_view.py
+++ b/lamindb/_view.py
@@ -5,7 +5,6 @@ import importlib
 import inspect
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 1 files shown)


Koncopd:
@falexwolf you added `display_df_with_descriptions` [here](https://github.com/laminlabs/lamindb/pull/2238). The problem is that it doesn't really check if `IPython` is present like in other parts of the code in the file. I have just moved the imports inside the function but i think we also a variant of `display_df_with_descriptions` when it not IPython, but just a terminal.

PR #2253, opened 2024-12-04
🐛 Fix display of (non-categorical) `str` features in `.describe()`

sunnyosun:


falexwolf:
@sunnyosun, can you add a bit of a description of what happened here?

falexwolf:
Do you mean non-categorical features?

sunnyosun:
No, meaning features that are categorical doesn't have associated labels.

PR #2266, opened 2024-12-09
🐛 Better error behavior for repeated calls of `standardize`

### LINKED ISSUE #2262: standardize KeyError output looks bad and should likely not get thrown on rerun
### Add a description

Running:

```
rxrx_curator.standardize("cell_line")
rxrx_curator.add_new_from("split")
```

for the second time leads to:

```
{
	"name": "KeyError",
	"message": "'\"cell_line\" is not a valid key, available keys are: \\'well\\', \\'sirna\\'!'",
	"stack": "---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
Cell In[12], line 1
----> 1 rxrx_curator.standardize(\"cell_line\")
      2 rxrx_curator.add_new_from(\"split\")
      3 rxrx_curator.add_new_from(\"experiment\")

File ~/miniconda3/envs/lamindb/lib/python3.11/site-packages/lamindb/_curate.py:362, in DataFrameCurator.standardize(self, key)
    360 else:
    361     if key not in avail_keys:
--> 362         raise KeyError(
    363             f'\"{key}\" is not a valid key, available keys are: {_print_values(avail_keys)}!'
    364         )
    365     else:
    366         if key in self._fields:  # needed to exclude var_index

KeyError: '\"cell_line\" is not a valid key, available keys are: \\'well\\', \\'sirna\\'!'"
}
```

I'm not sure whether a second run should error. It should much rather somehow determine that nothing needs to be standardized because it was already standardized.

Moreover, the `KeyError` message looks malformed.

Zethson:
Fixes #2262 

# Fixes KeyError formatting

## Before

`KeyError: '\"cell_line\" is not a valid key, available keys are: \\'well\\', \\'sirna\\'!'"`

## After

`KeyError: "'cell_line' is not a valid key, available keys are: 'well', 'sirna'!"`

# Fixes repeated calls of `standardize()`

```python
curator.standardize("disease")
curator.standardize("disease")
```

## Before

led to a traceback with 'KeyError: 'disease' is not a valid key.

## After

`• No unstandardized values found for 'disease'`

# Now hides the `add_new_from_columns` method from sphinx

Before this PR, the deprecated `add_new_from_columns` method was still listed in our documentation. It is now hidden.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,23 +29,28 @@ jobs:
         with:
           submodules: recursive
           fetch-depth: 0
+
       - uses: actions/setup-python@v5
         if: ${{ matrix.group == 'storage' }}
         with:
           python-version: "3.9"
+
       - uses: actions/setup-python@v5
         if: ${{ matrix.group == 'unit-storage' }}
         with:
           python-version: "3.11" # there are problems with installation time with python 3.12
+
       - uses: actions/setup-python@v5
         if: ${{ matrix.group != 'storage' && matrix.group != 'unit-storage'}}
         with:
           python-version: "3.12"
+
       - name: cache pre-commit
         uses: actions/cache@v4
         with:
           path: ~/.cache/pre-commit
           key: pre-commit-${{ runner.os }}-${{ hashFiles('.pre-commit-config.yaml') }}
+
       - name: cache postgres
         if: ${{ matrix.group == 'faq' || matrix.group == 'unit-core' || matrix.group == 'unit-storage' }}
         id: cache-postgres
@@ -63,9 +68,11 @@ jobs:
       - name: install postgres
         if: ${{ matrix.group == 'faq' }}
         run: sudo apt-get install libpq-dev
+
       - name: install graphviz
         if: ${{ matrix.group == 'tutorial' || matrix.group == 'guide' || matrix.group == 'biology' }}
         run: sudo apt-get -y install graphviz
+
       - run: nox -s lint
         if: ${{ matrix.group == 'tutorial' }} # choose a fast-running a group
       - run: nox -s "install_ci(group='${{ matrix.group }}')"
@@ -75,12 +82,14 @@ jobs:
           aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           aws-region: us-east-1
       - run: nox -s "build(group='${{ matrix.group }}')"
+
       - name: upload coverage
         uses: actions/upload-artifact@v4
         with:
           name: coverage--${{ matrix.group }}
           path: .coverage
           include-hidden-files: true
+
       - name: upload docs
         if: ${{ matrix.group == 'tutorial' || matrix.group == 'guide' || matrix.group == 'biology' || matrix.group == 'faq' || matrix.group == 'storage' }}
         uses: actions/upload-artifact@v4
@@ -96,6 +105,7 @@ jobs:
         with:
           submodules: recursive
           fetch-depth: 0
+
       - name: checkout lndocs
         uses: actions/checkout@v4
         with:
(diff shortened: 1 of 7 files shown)


falexwolf:
Ahh, main is broken because of the `wetlab` upgrade:

- https://github.com/laminlabs/lamindb/pull/2271

I guess I'll wait with the above until this here is merged.

.github/workflows/build.yml
```diff
@@ -29,23 +29,28 @@ jobs:
         with:
           submodules: recursive
           fetch-depth: 0
+
       - uses: actions/setup-python@v5
         if: ${{ matrix.group == 'storage' }}
         with:
           python-version: "3.9"
+
```
falexwolf, line 37: Why all these new lines? Is this some standard that we're introducing here? So far, from what I've seen and implemented, our standard is no new-lines in the GA yaml files.

lamindb/_curate.py
```diff
@@ -111,6 +111,15 @@ def __repr__(self) -> str:
 class BaseCurator:
     """Curate a dataset."""
 
+    def __init_subclass__(cls, **kwargs):
+        super().__init_subclass__(**kwargs)
+        import sys
+
+        # Deprecated methods
+        if "sphinx" not in sys.modules:
```
falexwolf, line 119: Arrgh, this seems an incredibly manual hack to account for rendering this in sphinx. Is this always needed as part of the deprecation process?

falexwolf:
Due to inconsistent indendation levels, this summary is very hard to read through -- can you please use headings or another way of styling things?

image

.github/workflows/build.yml
Zethson: Yes, we'd be introducing it here. I find the huge blob without newlines utterly unreadable. We don't code like that either. We can discuss it and I can revert it if you're not happy with it.

lamindb/_curate.py
Zethson: This has **always** been a part of the "Lukas deprecration process". See my original PR where I tried like 5 different solutions till I found this one that worked. Now actually, here we might get a way with a different solution because this isn't a Django model where there's a lot more complexity involved but for consistency, we can go for the same process.

falexwolf:
Done:
image

lamindb/_curate.py
falexwolf: Got it! :D

Zethson:
If you still have your source, we can go for B. It's less overwhelming and I like numbered changes. Easier to refer to.

.github/workflows/build.yml
falexwolf: I personally don't care for as long as we're consistent and follow whatever standard there is about this! No need to revert from my end, but a brief note in #eng-devops-disc would be good as I might just have reverted that in another PR thinking it's unconscious.

lamindb/_curate.py
falexwolf: Got it! :D

.github/workflows/build.yml
falexwolf: I personally don't care for as long as we're consistent and follow whatever standard there is about this! No need to revert from my end, but a brief note in #eng-devops-disc would be good as I might just have reverted that in another PR thinking it's unconscious.

Zethson: Let me correct myself, we are already using newlines in several schema modules.

lamindb/_curate.py
sunnyosun: I don't think we need to hide it, showing deprecates isn't enough? I don't think we do this for all deprecated methods, do we?

.github/workflows/build.yml
sunnyosun: I'm not sure, I think with - is in fact fine

falexwolf:
> If you still have your source, we can go for B. It's less overwhelming and I like numbered changes. Easier to refer to.

Yeah, it's in the version drop down:

image

I just fixed the indenting but left everything else you had the same.

lamindb/_curate.py
Zethson: No, we hide deprecated methods because we don't want users to see nor use them. They might even not have any functionality like this one. Our API docs are already extremely cluttered especially with some more Django stuff. I'd always hide deprecated methods.

sunnyosun (approved):

PR #2270, opened 2024-12-10
✨ Add an artifact loader for .yaml

Koncopd:
Builds on and extends https://github.com/laminlabs/lamindb/pull/2172


Koncopd:
test (storage) failure was fixed in https://github.com/laminlabs/lamindb/pull/2266, i don't want to introduce conflicts so leave it failing here.

PR #2273, opened 2024-12-11
🐛 Fix very long runtimes for Artifact.describe

Koncopd:
- Identified a few bottlenecks with a profiler and fixed them.

Now 
```
ln.Artifact.get("NiaMP8YQMbmkdOgnAa1T").describe()
ln.Artifact.get("LmlgCMQRhSEHtcbG0000").describe()
```
runs around 10 sec each instead of several minutes.

- Fixed describe when artifact is queried with `using`.
`ln.Artifact.using("laminlabs/cellxgene").get("g0RcSSYe5vQKzSWYkhMc").describe()` was failing before.

4-10 sec depending on annotations still feels pretty slow though, so it is probably not the final iteration.

### DIFF
--- a/lamindb/core/_django.py
+++ b/lamindb/core/_django.py
@@ -140,18 +140,21 @@ def get_artifact_with_related(
                     artifact, {i["featureset"]: i["slot"] for i in v}
                 )
 
-    m2m_any_expr = reduce(
-        lambda a, b: a | b,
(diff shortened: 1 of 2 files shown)


Zethson:
@Koncopd does this also improve the memory usage? Can I try it again now?

Koncopd:
I hope so, but i haven't checked.

PR #2289, opened 2024-12-17
🎨 Proper type check of FeatureSet rename

sunnyosun:
Amends https://github.com/laminlabs/lamindb/pull/2288

FeatureSet should be allowed to rename without checking, since the dataset itself will not contain the name.


falexwolf:
A test would also be good here, @sunnyosun.

PR #2290, opened 2024-12-17
✨ Add `SpatialDataCurator`

Zethson:
- Adds a SpatialDataCurator that allows the curation of [SpatialData](https://github.com/scverse/spatialdata) objects
- Currently depends on PR 806 in SpatialData and expects the sample level (or dataset level) metadata to be stored in a single zattrs key (here `sample`)

```python
import lamindb as ln
import bionty as bt
import pandas as pd
from spatialdata.datasets import blobs

sdata = blobs()
sdata.attrs["sample"] = {
    "assay": "Visium Spatial Gene Expression",
    "disease": "Alzheimer's dementia",
    "developmental_stage": "very early",
}
sdata.tables["table"].var.index = [
    "ENSG00000139618",  # BRCA2
    "ENSG00000157764",  # BRAF
    "ENSG00000999999"   # Does not exist - to test add new 
]
sdata.tables["table"].obs["region"] = pd.Categorical(
        ["region 1"] * 13 + ["region 2"] * 13
    )

curator = ln.Curator.from_spatialdata(
    sdata,
    var_index={"table": bt.Gene.ensembl_gene_id},
    categoricals={
        "sample": {"assay": bt.ExperimentalFactor.name,
                   "disease": bt.Disease.name,
                   "developmental_stage": bt.DevelopmentalStage.name},
        "table": {"region": ln.ULabel.name},
    },
    organism="human",
)

curator.validate()
curator.add_new_from_var_index("table")
curator.add_new_from(key="developmental_stage", accessor="sample")
curator.add_new_from(key="region", accessor="table")

curator.standardize(key="disease", accessor="sample")

artifact = curator.save_artifact(description="blob spatialdata")
artifact.describe()
```

image

- Moves the docs to Python 3.12 because 3.10 caused issues
- Introduces a new pattern in the build CI using `dorny/paths-filter@v3` to condition the groups that are executed. I tested several options and I can only say now that it executes the spatial group but I can't test with this PR the cases where it should NOT execute the spatial group. I'll follow up if necessary.


Koncopd:
We had this many times, it is `uv` problem, sometimes its resolver goes mad and tries to install very old package versions. One solution is to install with `pip`, another is to find the package causing the problem and add `uv` install manually at the beginning with an appropriate lower bound.

sunnyosun:
No I feel almost the entire curator file should be just another repo, it feels so heavy to run all the CI with tiledbsoma and spatialdata and many more in the future if I want to make an unrelated change.

I get that Alex is a huge fan of monorepo, but if we continue supporting many 100 more formats, do we really want to run all these CI in every PR push?

falexwolf:
I think we also should eliminate `lamin_spatial` which in my mind should never have seen the light of day. 😅

falexwolf:
Folders
```
curators/
   __init__.py. # this has the whole slew of existing curators
   _spatialdata/  # this has what's currently in lamin-spatial and triggers the corresponding tests that are also in lamin-spatial
       __init__.py
```

Docs:
```
curators/
- DataFrameCurator
- AnnDataCurator
- SpatialDataCurator
```

New suggested idiom:
```python
ln.curators.SpatialDataCurator(...)
```
instead of
```python
ln.Curator.from_spatialdata(...)
```

Zethson:
Followup in https://github.com/laminlabs/lamindb/issues/2303

With the next release we can yank `lamin-spatial` from pypi and I'll remove the `SpatialDataCurator` code there.

falexwolf:
> With the next release we can yank lamin-spatial from pypi and I'll remove the SpatialDataCurator code there.

Great! :D -- One PyPI package less!

.github/workflows/build.yml
```diff
@@ -22,28 +22,29 @@ jobs:
...
-
     steps:
       - uses: actions/checkout@v4
         with:
           submodules: recursive
           fetch-depth: 0
 
-      - uses: actions/setup-python@v5
-        if: ${{ matrix.group == 'storage' }}
+      - uses: dorny/paths-filter@v3
```
falexwolf, line 33: Interesting! I didn't know about this before.

Is it correct that you chose this over the native working-directory trigger because you wanted keep a single `build.yml`?

lamindb/core/__init__.py
```diff
@@ -81,18 +81,18 @@
     ValidateFields,
 )
 
-from lamindb._curate import (
+from lamindb._query_manager import QueryManager
+from lamindb._query_set import QuerySet, RecordList
+from lamindb.core._feature_manager import FeatureManager, ParamManager
+from lamindb.core._label_manager import LabelManager
+from lamindb.curators import (
     AnnDataCurator,
```
falexwolf, line 89: Ah, do we really want to re-export all curators at the root-level? That's going to clutter the root API because they don't even share a common prefix.

And `CurateLookup` and `BaseCurator` are not even user-facing interfaces.

I'd have the user-facing API under `curators`.

lamindb/curators/__init__.py
```diff
@@ -1712,6 +1715,80 @@ def from_tiledbsoma(
             exclude=exclude,
         )
 
+    @classmethod
```
falexwolf, line 1718: I'd not add the duplicated interface.

We can discuss this but I think the new way of calling curators should be
```python
ln.curators.SpatialDataCurator(sdata, ...)
```
instead of 
```python
ln.Curator.from_spatialdata()
```

Because we only want one way of doing things it should only be the `ln.curators.SpatialDataCurator(sdata, ...)` way I'd say. 🤔

.github/workflows/build.yml
falexwolf: But does this actually work? The last CI run ran all jobs. I'd have expected to only run the "spatial" job.

image

falexwolf: The idea is to only run the spatialdata tests when there is a change to spatialdata-related code. No other tests should then run.

Let's discuss tomorrow!

Zethson: > Is it correct that you chose this over the native working-directory trigger because you wanted keep a single build.yml?

For now! See https://github.com/laminlabs/lamindb/issues/2303

> But does this actually work? The last CI run ran all jobs. I'd have expected to only run the "spatial" job
> The idea is to only run the spatialdata tests when there is a change to spatialdata-related code. No other tests should then run.

The last CI is supposed to run all jobs with the current way I've designed it. It touched more than just the spatial code. I hear you that you want to exclusively only run parts that were touched and this will require more thinking - especially because some code pieces here are not isolated. For example, touching the other Curators can totally break the SpatialdataCurator. Hence, the spatial CI is triggered if any of the Curator code is touched.

lamindb/curators/__init__.py
Zethson: This will happen but not with this PR: https://github.com/laminlabs/lamindb/issues/2303

.github/workflows/build.yml
falexwolf: Ok, if this is how it behaves it's all great!

I'll play with this. I think it's very important that this is well organised.

sunnyosun:
This PR broke cellxgene-lamin because the import filename was changed from ._curate to curators...

Zethson:
> This PR broke cellxgene-lamin because the import filename was changed from ._curate to curators...

Sorry - I'll fix that

falexwolf:
I think the cellxgene-lamin package should also disappear and be part of lamindb just like spatial curator

sunnyosun:
I'm already doing it!

falexwolf:
☺️

sunnyosun:
Also this fix is now gone since this PR: https://github.com/laminlabs/lamindb/pull/2299

It's probably due to you deleting the _curator.py file before rebasing...

Zethson:
Sorryyyy also on me then, yes..

Issue #2301, opened 2024-12-19
🚸 Shall we make `aws` and `ourprojects` default dependencies?


falexwolf:
Goal is to cut down the optionality/complexity of installation.

While purely local usage or purely GCP wouldn't need it, it seems that these are edge cases. Almost anybody will at some point want to pull something from an S3 bucket. Hence I vote to remove the `aws` extra and add the dependencies to the defaults.

For `ourprojects`, it also seems that we shouldn't require an extra but simply install it. It has no additional dependencies at all and is just a few 100 lines of Python code.

`bionty` & `jupyter` indeed bring some additional dependencies; hence these merit extras.

@Zethson @sunnyosun @Koncopd

falexwolf:
This is how this looks:

- https://github.com/laminlabs/lamindb/pull/2302

Zethson:
`aws` I can see but I think @Koncopd knows better whether there'll be any potential conflicts in existing environment. TBH I was sometimes surprised at how many issues dependencies like boto3 caused.

Concerning `ourprojects` - generally also yes as long as we pin it like you're doing.

PR #2302, opened 2024-12-19
🚸 Simplify installation & API overview

falexwolf:
## Installation

Before:
```bash
pip install "lamindb[aws,bionty,jupyter,ourprojects]"
```

After:
```bash
pip install "lamindb[bionty,jupyter]"
```

## Docs

Before | After
--- | ---
image | image

## Materials

- https://github.com/laminlabs/lamindb/issues/2301

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -16,15 +16,9 @@
 [submodule "sub/clinicore"]
 	path = sub/clinicore
 	url = https://github.com/laminlabs/clinicore
-[submodule "sub/findrefs"]
-	path = sub/findrefs
-	url = https://github.com/laminlabs/findrefs
 [submodule "sub/ourprojects"]
 	path = sub/ourprojects
 	url = https://github.com/laminlabs/ourprojects
 [submodule "sub/cellregistry"]
 	path = sub/cellregistry
 	url = https://github.com/laminlabs/cellregistry
-[submodule "sub/omop"]
-	path = sub/omop
-	url = https://github.com/laminlabs/omop.git
--- a/docs/api.md
+++ b/docs/api.md
@@ -4,25 +4,11 @@
 
 ```{toctree}
 :maxdepth: 1
-:caption: CLI
+:caption: CLI & lamindb
 :hidden:
 
 cli
-```
-
-```{toctree}
-:maxdepth: 1
-:caption: Python
-:hidden:
-
 lamindb
-bionty
-ourprojects
-findrefs
-cellregistry
-wetlab
-clinicore
-omop
 ```
 
 ```{toctree}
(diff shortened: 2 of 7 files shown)


Koncopd:
But why the submodules were removed?

Koncopd:
This will also break a ton of CI scripts across the repos.
I really don't like that aws dependency is installed by default now, it is very problematic because sometimes people because sometimes it can break existing environments, so if people have their environment with `s3fs` etc configured, it is better not to touch it.

Koncopd:
I think such things should be discussed before merging.

Koncopd:
Ok, about breaking i am probably not right, it was some time before, but was simplified.

Koncopd:
No, thinking about this, it can still cause problems.

falexwolf:
> No, thinking about this, it can still cause problems.

What problems can it cause? I think it will reduce problems!

Assume that somebody has an incompatible version of s3fs installed. If they forget to provide the `aws` extra during installation, they'll never get feedback on clashing version requirements.

Now, they'll immediately get feedback if something clashes with what lamindb expects for `aws`.

I believe this might have been the cause of some bugs where people didn't notice anything until they call `.load()` or `.cache()` and then get a cryptic error and don't know what to do. That error would have been avoided if `aws` is simply part of `lamindb`.

falexwolf:
> But why the submodules were removed?

`findrefs` is deprecated in favor of `ourprojects`.

`omop` isn't a real use case; this requires much more discussion and is at this point confusing. There is still a link to https://omop.lamin.ai/

The docs job was beyond 6 min before this; now it's back to 3 min because omop is so huge.

falexwolf:
> I think such things should be discussed before merging.

Yes, I know. Sorry, I discussed it 4 times with myself and then considered it's high probability that we actually want this and be done with it so that I could finalize a downstream dependency at the same time.

Koncopd:
> What problems can it cause? I think it will reduce problems!

Yes, thinking yet again, you are right here. I had problems on sagemaker, where it worked with a given environment but broke with `lamindb[aws]`, but it was long time ago, there are no sagemaker environments like this anymore and the structure of aws deps changed.
The only thing i don't like it having `boto3` here https://github.com/laminlabs/lamindb-setup/blob/8426994be65f5a2cfded28d241c402e5ac207e7b/pyproject.toml#L32 that is used nowhere in `lamindb`, but was added just not to break existing environments where `boto3` is installed and used.

Koncopd:
If we have clients in the future that rely only on `gcp`, we will need to reintroduce `aws` extra i think.

falexwolf:
> If we have clients in the future that rely only on gcp, we will need to reintroduce aws extra i think.

I don't think so based on this argument from the issue

> While purely local usage or purely GCP wouldn't need it, it seems that these are edge cases. Almost anybody will at some point want to pull something from an S3 bucket. 

Also people on GCP will at some point want to access data from some AWS bucket. And AFAIK boto3 and s3fs aren't big; these are just a few python files.

Koncopd:
Yes, this makes sense, you are right.

Issue #2265, opened 2024-12-06
🚸 ln.track should prompt to enter any key instead of y or n with an error


Koncopd:
### Add a description

subj

falexwolf:
Can then also make this message analogous so that it no longer raises an error but the same analogous dialogue in a notebook.

<img width="671" alt="image" src="https://github.com/user-attachments/assets/b5ec29ed-7f64-49b2-b9a7-769b490e4b74" />

ap--:
Just fyi, because a user ran into this:

There can be a deadlock in older jupyterlab versions where when the input is displayed and a user manually reruns the cell, the kernel deadlocks. Might be worth not showing the input when running in older jupyter lab versions

https://github.com/jupyterlab/jupyterlab/pull/8713
https://github.com/jupyterlab/jupyterlab/pull/16032

Seems to be fixed in:
- **3.x branch:** 3.1.0+
- **4.x branch:** 4.2.0+

falexwolf:
Thanks for sharing! We'll account for this in one way or the other. 👌

falexwolf:
This is comprehensively addressed here:

- https://github.com/laminlabs/lamindb/pull/2308

PR #1982, opened 2024-09-26
🚸 Minimal `ln.track()`

falexwolf:
Before | After
--- | ---
image | image

Resolves:

- https://github.com/laminlabs/lamindb/issues/1976

Needs:

- https://github.com/laminlabs/lamin-cli/pull/77

PR #2308, opened 2024-12-28
🚸 Improve `ln.track()`

falexwolf:
# Summary

This improves `ln.track()` by 

1. enabling users to organize transforms in folders
2. letting users omit the `uid` argument
3. automating the versioning process (requirement for 1.)
4. enabling to save scripts and notebooks without running them (corollary of 1.)
5. avoiding the interactive prompt in a notebook and the throwing of an error in a script (corollary of 1.)
6. no longer requiring titles in notebooks

# Examples

## Organize transforms in folders

Before this PR, `lamindb` didn't know how to deal with `/`-separated values for transform `key`.

Now, you can have any folder structure in your instance, e.g.:

```
my-notebook.ipynb
my_script.py
projectA/my-notebook.ipynb
projectA/my_script.py
projectB/my-notebook.ipynb
projectB/my_script.py
```

Upon calling `lamin load ...`, you mirror this structure in your current working directory and work with them as you would in git or other versioned file systems (Google Drive etc.): if you rename a file or move it into a different folder, you initiate a new version family.

## Init a transform

Before:
```python
>>> ln.track()
✗ to track this script, run: ln.track("PC0eB2QPm0jW0000")
```

After:
```python
>>> ln.track()
→ created Transform('2D4W8qklAPLg0000'), started new Run('zXWSu267...') at 2024-12-30 05:13:07 UTC
```

## Bump version of transform

Before:
```python
>>> ln.track("PC0eB2QPm0jW0000")
Source code changed, bump revision by setting:

ln.track("PC0eB2QPm0jW0001")
```

After:
```python
>>> ln.track()
→ there already is a transform with key 'my_test_script.py', creating new version 'PC0eB2QPm0jW0001'
→ created Transform('PC0eB2QPm0jW0001'), started new Run('gl5Nvt0m...') at 2024-12-30 05:58:54 UTC
```

# How does version management work

When called without a `uid`, `ln.track()` auto-generates `uid` values modeling a revision process similar to other tools (git, Google Docs, etc.):

1. if there is no transform with the same `key`, make a new `uid`
2. if there is a transform with the same `key`, bump the version `uid`

# Degradations

The main downsides of enabling `ln.track()` without `uid` are

1. Users may rename a script or notebook and then be confused that they initiate a new version family if they call `ln.track()` without passing a `uid` from the previous version family
2. There are now two ways of calling `ln.track()`; one that passes the `uid`, which allows to maintain version families across renaming (unlike git); one without, which leads to disadvantage 1.

Given the 6 improvements/simplifications of the workflow listed above, I think that maintaining the explicit passing of a `uid` to `ln.track()` for advanced users to address 1. is OK given the reduced learning curve for beginning users that this PR introduces and the convenience that also advanced users should feel in most cases (often, when renaming, one actually wants to create a new version family).

# Materials

Needs:

- https://github.com/laminlabs/lamin-cli/pull/99

Continues a sequence of improvements for `ln.track()`:

- https://github.com/laminlabs/lamindb/pull/2176
- https://github.com/laminlabs/lamindb/pull/1982
- https://github.com/laminlabs/lamindb/pull/1816

Fixes the below and other informal reports about the user journey by less experienced users:

- https://github.com/laminlabs/lamindb/issues/2265


falexwolf:
Still need to support the case in which multiple folders are created on the UI at a later point and a user loads one of them and there is no `uid` in the script/notebook.

falexwolf:
This is still WIP. I need to do another refactor.

Koncopd:
Interesting, but i doubt that key is reliable enough for this. But i haven't looked at the code yet and don't know the exact logic. Probably makes sense to discuss this sometime.

falexwolf:
> Interesting, but i doubt that key is reliable enough for this. But i haven't looked at the code yet and don't know the exact logic. Probably makes sense to discuss this sometime.

I have weighed pros & cons for quite some time now and will document them in the PR description. I don't think there is a robustness issue; in my mind there are only UX issues where the advantages of this (non-breaking) change outweigh the disadvantages of omitting the `uid` in many cases.

Koncopd:
This should still prompt users to add `uid` to `track`, in a non-breaking manner maybe. I expect a ton of problems from relying on key alone.

Issue #2137, opened 2024-11-07
IntegrityError: UNIQUE constraint failed: bionty_gene.ensembl_gene_id when repeatedly validating example dataset 


Zethson:
### Report

```
!lamin init --storage ./run-tests --name run-tests --schema bionty

import lamindb as ln
import bionty as bt

adata = ln.core.datasets.anndata_pbmc68k_reduced()

curator = ln.Curator.from_anndata(adata, var_index=bt.Gene.ensembl_gene_id, organism="human")
curator.validate()
curator.validate()
```

leads to

```
{
	"name": "IntegrityError",
	"message": "UNIQUE constraint failed: bionty_gene.ensembl_gene_id",
	"stack": "---------------------------------------------------------------------------
IntegrityError                            Traceback (most recent call last)
File ~/miniconda3/envs/lamindb/lib/python3.11/site-packages/django/db/backends/utils.py:105, in CursorWrapper._execute(self, sql, params, *ignored_wrapper_args)
    104 else:
--> 105     return self.cursor.execute(sql, params)

File ~/miniconda3/envs/lamindb/lib/python3.11/site-packages/django/db/backends/sqlite3/base.py:354, in SQLiteCursorWrapper.execute(self, query, params)
    353 query = self.convert_query(query, param_names=param_names)
--> 354 return super().execute(query, params)

IntegrityError: UNIQUE constraint failed: bionty_gene.ensembl_gene_id

The above exception was the direct cause of the following exception:

IntegrityError                            Traceback (most recent call last)
Cell In[3], line 1
----> 1 curator.validate()

File ~/PycharmProjects/lamindb/lamindb/_curate.py:548, in AnnDataCurator.validate(self, organism)
    543     logger.important(
    544         f\"validating metadata using registries of instance {colors.italic(self._using_key)}\"
    545     )
    547 # add all validated records to the current instance
--> 548 self._update_registry_all()
    550 validated_var, non_validated_var = validate_categories(
    551     self._adata.var.index,
    552     field=self._var_field,
   (...)
    558     **self._kwargs,  # type: ignore
    559 )
    560 validated_obs, non_validated_obs = validate_categories_in_df(
    561     self._adata.obs,
    562     fields=self.categoricals,
   (...)
    566     **self._kwargs,
    567 )

File ~/PycharmProjects/lamindb/lamindb/_curate.py:517, in AnnDataCurator._update_registry_all(self, validated_only, **kwargs)
[...]

PR #2312, opened 2025-01-02
🎨 Standardize name

sunnyosun:
Prevents mistakes like: https://github.com/laminlabs/lamindb/issues/2137

This PR makes sure that standardize only performs on name field.

### DIFF
--- a/lamindb/_can_curate.py
+++ b/lamindb/_can_curate.py
@@ -200,7 +200,7 @@ def _inspect(
                     f" {colors.italic('.from_values()')}"
                 )
 
-            nonval = bionty_result.non_validated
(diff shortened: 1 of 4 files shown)


falexwolf:
Great! Just can you add more color to the description of the PR with a small example so that everybody reading this understands what's changing? It's even quite hard for me (even though I figured it out). One reason is also that the PR title isn't descriptive enough. Can you please improve it?

Maybe just the one you mentioned here: https://github.com/laminlabs/lamindb/issues/2137#issuecomment-2567652731

sunnyosun:
The example shows synonyms are only mapped when field is "symbol", not "ensembl_gene_id". I don't think "This PR makes sure that standardize only performs on name field." has grammar error.

PR #2315, opened 2025-01-02
♻️ Eliminate mocks from base

Koncopd:


Koncopd:
I think codecov is off here.

PR #2323, opened 2025-01-05
🚚 Database migrations for lamindb 1.0a1

falexwolf:
Bumps `__version__` from `0.78a1` to `1.0a1`.

Internal instances migrated:

- `laminlabs/lamindata`
- `laminlabs/lamin-site-assets`
- `laminlabs/lamin-dev`
- `laminlabs/arrayloader-benchmarks`
- `laminlabs/bionty-assets`

## Other changes

- upgraded Python to `>=3.10` and Django to `>=5` because we're using `db_default` -- discussion [here on GitHub](https://github.com/laminlabs/lamindb-setup/pull/927#issuecomment-2578257332)

## Schema changes

- remove `_source_code_artifact` from Transform, it's been deprecated since 0.75 https://github.com/laminlabs/lamindb/pull/2323/commits/222931a61f19aadbc3576bf2024e3ff3435a5500
   - data migration: for all transforms that have `_source_code_artifact` populated, populate `source_code`
- rename `Transform.name` to `Transform.description` because it behaves like `Artifact.description` and it's confusing because people expect versioning to happen based on `name` https://github.com/laminlabs/lamindb/pull/2323/commits/5d68c5e4c2aa4e5d15c8a2e74977903063f35d77 [Claude chat](https://claude.ai/share/29b5c2a5-4922-417f-8d99-2563de50d535)
  - backward compat: 
    - in the `Transform` constructor use `name` to populate `key` in all cases in which only `name` is passed
    - return the same transform based on `key` in case `source_code is None` via `._name_field = "key"`
  - data migrations:
     - there already was a legacy `description` field that was never exposed on the constructor; to be safe, we concatenated potential data in it on the new description field [Claude chat](https://claude.ai/share/29b5c2a5-4922-417f-8d99-2563de50d535) 
     -  for all transforms that have `key=None` and `name!=None`, use `name` to pre-populate `key`
- rename `Collection.name` to `Collection.key` for consistency with Artifact & Transform and the high likelihood of users wanting to organize them hierarchically https://github.com/laminlabs/lamindb/pull/2323/commits/2bef93a87d3830a38fd62e5d8ae25751d69487db
- a `_branch_code` integer on every record to model pull requests https://github.com/laminlabs/lamindb/pull/2323/commits/483fb3705e375b4aa2a42d3a75edfaa3207a6e28
   - include `visibility` within that code
   - repurpose `visibility=0` as `_branch_code=0` as "archive"
   - put an index on it: see [Claude chat](https://claude.ai/share/124f91ac-75e3-4ed3-9c3c-b7ac147cf565)
   - code a "draft" as _branch_code = 2, and "draft prs" as negative branch codes [Claude thread](https://claude.ai/share/124f91ac-75e3-4ed3-9c3c-b7ac147cf565) -- this is better than adding another field as originally planned ([internal Slack thread](https://laminlabs.slack.com/archives/C04FPE8V01W/p1736242336846459?thread_ts=1725632618.145149&cid=C04FPE8V01W))
- rename values "number" to "num" in dtype
- an `.aux` json field on `Record`: [Claude thread](https://claude.ai/share/c3b22db2-763e-47a0-83dc-8ff7f5f2f7ff)
- a SmallInteger `run._status_code` that allows to write `finished_at` in clean up operations so that there is a run time also for aborted runs
- rename `Run.is_consecutive` to `Run._is_consecutive`
- a `_template_id` FK to store the information of the generating template (whether a record is a template is coded via _branch_code)
- rename `_accessor` to `otype` to publicly declare the data format as `suffix, accessor`
- rename `Artifact.type` to `Artifact.kind` ([internal Slack thread](https://laminlabs.slack.com/archives/C04FPE8V01W/p1736003536684769)) [Claude chat](https://claude.ai/share/1d6231a6-aa85-401e-86ed-3baa16a06eeb)
- a FK to artifact `run._logfile` which holds logs
- a `hash` field on `ParamValue` and `FeatureValue` to enforce uniqueness without running the danger of failure for large dictionaries https://github.com/laminlabs/lamindb/pull/2323/commits/9e4b503d876743890753e1f97c8a73aeb91417aa [Claude chat](https://claude.ai/share/46037772-e05e-4b45-b45f-78a46e5857bf)
- an `Artifact._curator: dict` field https://github.com/laminlabs/lamindb/pull/2323/commits/068bcf777d862ebbe2bdf9e8485d39fbcf9bc45b
- add a boolean field `._expect_many` to `Feature`/`Param` that defaults to `True`/`False` and indicates whether values for this feature/param are expected to occur a single or multiple times for every single artifact/run https://github.com/laminlabs/lamindb/pull/2323/commits/1cc1ca868cea459cbb25c519dd236aeb2e5c4b4a
    - for feature 
        - if it's `True` (default), the values come from an observation-level aggregation and a dtype of `datetime` on the observation-level mean `set[datetime]` on the artifact-level
        - if it's `False` it's an artifact-level value and `datetime` means `datetime`; this is an edge case because an arbitrary artifact would always be a set of arbitrary measurements that would need to be aggregated ("one just happens to measure a single cell line in that artifact")
    - for param
        - if it's `False` (default), the values mean artifact/run-level values and `datetime` means `datetime`
        - if it's `True`, the values would be from an aggregation, this seems like an edge case but say when characterizing a model ensemble trained with different parameters it could be relevant
- remove the `.transform` foreign key from artifact and collection for consistency with all other records; introduce a property and a simple filter statement instead (context: https://github.com/laminlabs/laminhub/issues/1498) https://github.com/laminlabs/lamindb/pull/2323/commits/a4e0fb238aa8a41cb6b4324bd2e79a88b245a888
- store provenance metadata for `TransformULabel`, `RunParamValue`, `ArtifactParamValue` https://github.com/laminlabs/lamindb/pull/2323/commits/576d858c6eeffc254fa23a63c605eef60e1602f3
- enable linking projects & references to transforms & collections: https://github.com/laminlabs/ourprojects/pull/8/commits/370a8696702e119d86f5550978322eedbbf0cfc9
- rename `Run.parent` to `Run.initiated_by_run` https://github.com/laminlabs/lamindb/pull/2323/commits/9fbdd6f2a4dce2913879212e76dadd8c9d7787c0
- introduce a boolean flag on artifact that's called `_overwrite_versions`, which indicates whether versions are overwritten or stored separately; it defaults to `False` for file-like artifacts and to `True` for folder-like artifacts - see [Slack thread](https://laminlabs.slack.com/archives/C04TMJPNCBS/p1723987597397739)
- Rename `n_objects` to `n_files` for more clarity https://github.com/laminlabs/lamindb-setup/pull/927/commits/eba36ba4912d06c1faf655b90a5c7660d1f8e077 [Claude chat](https://claude.ai/share/25546de5-7797-4633-8c6f-c438a6f074e4)
- Add a `Space` registry to lamindb with an FK on every `BasicRecord` [internal Notion page](https://www.notion.so/laminlabs/UX-of-managing-access-based-on-teams-spaces-1762aeaa55e18072a1f8fd7309ee15a9) https://github.com/laminlabs/lamindb/pull/2323/commits/a25df165a5395c2f9ab71afa1d12eea77788f03d
- add a name column to run so that a specific run can be used as an analysis and named; for almost all runs it's going to be None
- add a `.type` CharField to `Feature` and `Param` so that features & params can be hierarchically grouped
- remove `_previous_runs` field on everything except Artifact & Collection; we can model this better with events going forward
- a boolean on `ULabel` called `is_concept` to distinguish ontological concepts from the labels themselves; for instance, you would never want to label an artifact with a ulabel `Project`, you'll only want to label with actual project values `Project 1`, `Project 2`, etc. https://github.com/laminlabs/lamindb/pull/2323/commits/6bd7f62dbfd4d2918eea44d23efbbefe543ac9f3

TODO in another PR:

- add a `Metric` and `MetricValue` registry that parallels `Param` and `ParamValue` up to an FK to `run`
   - or rename `Feature` to `Dimension` and `FeatureSet` to `Schema` and let people group metrics through a feature type metric 
- move the Project registry into lamindb with meaningful releationships (e.g. also for ULabel) and make it easy to label things by default with a project
- attach curator to FeatureSet or re-purpose FeatureSet as curator in some way or rename FeatureSet to Schema [Slack thread](https://laminlabs.slack.com/archives/C046F63LJJ2/p1736072921768029?thread_ts=1724051174.864259&cid=C046F63LJJ2)
- think through naming conventions of lamindb-related records: transfer, save_vitessce_config, run reports, etc. and migrate them
- rename "glue" transform type

## Materials

Needs:

- https://github.com/laminlabs/lamin-cli/pull/102
- https://github.com/laminlabs/lamindb-setup/pull/927

Affects:
- https://github.com/laminlabs/laminhub/pull/1810
- https://github.com/laminlabs/clinicore/pull/17
- https://github.com/laminlabs/ourprojects/pull/8
- https://github.com/laminlabs/wetlab/pull/90
- https://github.com/laminlabs/bionty/pull/178

Triggered these issues:
- https://github.com/laminlabs/laminhub/issues/1784
- https://github.com/laminlabs/laminhub/issues/1787

## Did not move forward with these changes

- a `Record._page_md` textfield so that every entity can hold markdown content
     - Versioning would go through the `Event` model with fields
     - Reverted (https://github.com/laminlabs/lamindb/pull/2323/commits/58e46ef675b5d63d4c3db4eb9a817c4c88e104c4) during this PR in favor of: https://github.com/laminlabs/laminhub/issues/1784
- a `Record._public` boolean that indicates whether a record is public [Claude chat](https://claude.ai/share/f9dde651-146a-479b-8cbd-02c1c1a98b4e)
    - Reverted (https://github.com/laminlabs/lamindb/pull/2323/commits/483fb3705e375b4aa2a42d3a75edfaa3207a6e28) because of [this internal Slack discussion](https://laminlabs.slack.com/archives/C04TMJPNCBS/p1736105677542579)
- check that there is an index on paramvalue.value and featurevalue.value, there is None but indexes on jsons are complicated and these tables won't be big so we rather hold off
- consider whether we want to eliminate the link tables associated with "_previous_runs" runs; decided against it [Claude thread](https://claude.ai/chat/b17baa80-3170-4f4d-a455-14b35c7de9a3)
- given that `feature_sets` is a public attribute and given that some people might want to query for `feature_values` or `param_values` outside the `.features` and `.params` managers; we should consider making `_feature_values` and `_param_values` public. This will likely also help people better understand the underlying mechanics if they're ever interested. Decided against because then there would be two ways to query by a feature or param. Everything is already supported through the lamindb-native API and by that considerably simpler than doing this with Django, in particular since https://github.com/laminlabs/lamindb/pull/2238. Also see the run docs where nested queries are exemplified:  image
- rename User to Person? no, better to keep it separate; it's OK to have to tables for the internal concept and things that people want to do with `Person` otherwise


Koncopd:
>automatically tracking artifact folder: https://claude.ai/chat/85b7a953-36d5-4afa-8efc-1bd6886eaefe

The conversation you were looking for could not be found.
what is it actually?

Koncopd:
> a Folder model with registry | key (see internal [Claude discussion](https://claude.ai/chat/85b7a953-36d5-4afa-8efc-1bd6886eaefe))

I don't really understand the point, as we have Storage already. Also we have different storage types, like `http`, `hf`, `s3` etc, not clear what is a folder for `http` or `hf`.

falexwolf:
> The conversation you were looking for could not be found. what is it actually?

This is something for the frontend; I shared the Claude discussion with you. After the whole exploration around the block model this will likely make more sense in `laminhub_client` (https://github.com/laminlabs/laminhub/issues/1784).

Koncopd:
I still don't see it.

falexwolf:
https://claude.ai/share/c554a663-75de-4dda-884e-a910b1b261bd

Koncopd:
Ok, that is clearer to me now, but i am still not sure. If we have a path in `s3`, `gs`, and `http` with the same "folder" in the key, is it really the same folder and do we want to see it as a unified folder really?

falexwolf:
>  If we have a path in s3, gs, and http with the same "folder" in the key, is it really the same folder and do we want to see it as a unified folder really?

`ArtifactFolder` and `TransformFolder` and `CollectionFolder` are merely used in the UI and merely for virtual keys. Let's discuss this here when we get to it: https://github.com/laminlabs/laminhub/issues/1787

Koncopd:
I will reworks types in `anndata_accessor` later, for now we should probably ignore `UP007`.

lamindb/_record.py
```diff
@@ -737,13 +737,19 @@ def save(self, *args, **kwargs) -> Record:
...
+                else:
+                    super(Record, self).save(*args, **kwargs)
                 _store_record_old_name(self)
             self._revises = None
         # save unversioned record
         else:
             check_name_change(self)
-            super(Record, self).save(*args, **kwargs)
+            if isinstance(self, BasicRecord):
+                super(BasicRecord, self).save(*args, **kwargs)
```
Koncopd, line 750: I see that `Record` is a subclass of `BasicRecord`, so the first condition will be always triggered. Is it fine? Anyways the else looks redundant here.

lamindb/models.py
```diff
@@ -805,6 +816,24 @@ class Record(models.Model, metaclass=Registry):
     machine learning or biological models.
     """
 
+    _branch_code: int = models.SmallIntegerField(db_index=True, db_default=1)
```
Koncopd, line 819: Am i right in assuming that the fields with `db_default` won't be populated with anything on model creation (until it is saved). I suspect we can have problems with this behavior.

falexwolf: According to Claude & the whole test suite here not. I asked whether I should declare both and Claude said `default` takes its value from `db_default` unless it's defined.

Checking the Django docs Claude wasn't entirely accurate: this is what happens

image

I'll play with this in an example to see whether it gives us problems. If Django is smart (and it typically is) the object corresponds to 1 in this case and to `None` in the other.

lamindb/_record.py
```diff
@@ -892,3 +898,5 @@ def delete(self) -> None:
 
 for name in METHOD_NAMES:
     attach_func_to_class_method(name, Record, globals())
+    if name not in {"__init__", "lookup", "delete"}:
+        attach_func_to_class_method(name, BasicRecord, globals())
```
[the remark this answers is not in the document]
falexwolf: Right, I didn't notice this because I just made one the super-class of the other.

This will in any case go away in the ditch-monkey-patching refactor. I'll fix it here, too, so that it's logically correct.

lamindb/_record.py
falexwolf: This is the weirdest behavior here; it has to be eliminated in the ditch-monkey-patching refactor. We'll then use `super()` instead.

For whatever reason this is the only way I could get everything to work.

Koncopd: I think you had this problem before you made `Record` a subclass of `BasicRecord`. I think it should work now just fine with `super(BasicRecord, self).save(*args, **kwargs)` only.

Koncopd: Because i believe here `isinstance(self, BasicRecord)` always evaluates to `True` for `self` of both `Record` and `BasicRecord`.

falexwolf: You're right I can very like switch this out.

lamindb/models.py
falexwolf: Duplicated with `default`. I discovered an edge case. Now it's safe even though the code is not as lean.

lamindb/_record.py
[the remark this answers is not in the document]
falexwolf: Refactored it. One indeed needs to attach twice. But again, this will disappear when we remove the monkey patching.

falexwolf:
This is ready. After another review from @sunnyosun we can merge this to main and then deploy the migrations against our own instances to test for a day or two.

On the weekend, we can complete the entire migration.

falexwolf:
FYI: the tests that hit our remote instances are commented out because I'll need to deploy the migrations first; I will only do this once it's reviewed.

sunnyosun:
Could we rename `FeatureManager._add_from` to `FeatureManager.add_from`? (backward compatible)
https://github.com/laminlabs/lamindb/blob/519b1d774aeb4bfcd76e19db6bda103f26daf5c0/lamindb/core/_feature_manager.py#L1184

sunnyosun (approved):

Koncopd (approved):
Note that the docs fail, otherwise looks good to me.

lamindb/_record.py
```diff
@@ -891,4 +893,5 @@ def delete(self) -> None:
     }
 
 for name in METHOD_NAMES:
+    attach_func_to_class_method(name, BasicRecord, globals())
     attach_func_to_class_method(name, Record, globals())
```
Koncopd, line 897: Isn't it enough to add them only to `BasicRecord`? Shouldn't they be inherited then by `Record` without adding them explcitily? Or it doesn't work like that?

lamindb/_transform.py
```diff
@@ -20,23 +20,36 @@ def __init__(transform: Transform, *args, **kwargs):
...
+                f"`name` will be removed soon, please pass '{key}' to `key` instead"
+            )
+        else:
+            if description is None:
+                description = kwargs.pop("name")
+                logger.warning(
+                    f"`name` will be removed soon, please pass '{description}' to `description` instead"
+                )
+            else:
+                raise ValueError("name doesn't exist anymore `description`")
```
Koncopd, line 47: It is unclear to me what this tries to convey.

Koncopd:
I have added a few comment about minor things.

Zethson (commented):
Awesome!

1. `ln.Artifact.filter(visibility=None).df()` gets changed to `ln.Artifact.filter(_branch_code=None).df()` in the deletion FAQ. I don't like that we're suggesting to filter by an underscore attribute. Should we consider not making this an underscore attribute given that `visibility` wasn't either?
2. Generally, for all things visibility, I found `visibility` to be much easier to understand than `_branch_code` although I of course see how much power we gain with it. It reads hacky though for users in my opinion.
3. WDYT about using `obj_type` instead of `otype`? It took me a few seconds too much to understand what you meant.
4. I don't get warm with `kind` at all but well `type` is a very overloaded term...

lamindb/_record.py
falexwolf: No, it's surprisingly not enough! See our other discussion on this. I tried it after you raised it yesterday.

falexwolf:
Thanks, @Zethson. Responses below.

> Should we consider not making this an underscore attribute given that visibility wasn't either?

In fact this is backward compatible and people can keep doing `filter(visibility=0)`. I fixed this in the FAQ.

> Generally, for all things visibility, I found visibility to be much easier to understand than _branch_code although I of course see how much power we gain with it. It reads hacky though for users in my opinion.

💯  `visibility` will stay around at least for a while for the user (this PR doesn't break the calls). `_branch_code` is a private/internal concept that is exposed through "Archive", "Trash", "PR", etc. and until lamindb 2 also with "visibility".

> WDYT about using obj_type instead of otype? It took me a few seconds too much to understand what you meant.

The main reason for `otype` is `dtype`. You bringing up `obj_type` makes me realize a problem I hadn't thought of. We also have `n_objects` and refer to the things in storage as "objects". I think we might want to rename `n_objects` to `n_files`. The latter name again isn't so great because these files can be fragments in an array representation. But ultimately they're blobs like files. Which means we could also use `n_blobs`. I will ask Claude.

But either way I don't like `obj_type` because it doesn't look pretty and isn't a known convention. If anything I'd say `object_type`. But then I guess `otype` is OK because it parallels `dtype`.

> I don't get warm with kind at all but well type is a very overloaded term...

Yes, I agree. But it's still better than `type`. I linked the Claude discussions for everything at the top. It's hard to find something better than `kind`.

.pre-commit-config.yaml
```diff
@@ -27,7 +27,7 @@ repos:
     rev: v0.8.6
     hooks:
       - id: ruff
-        args: [--fix, --exit-non-zero-on-fix, --unsafe-fixes]
```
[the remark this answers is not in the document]
falexwolf: Because I was unable to get information about which things ruff was modifying otherwise. I'll restore it. If I hadn't added 3 exceptions this PR would be full of linting changes and impossible to understand.

lamindb/_collection.py
```diff
@@ -107,27 +105,27 @@ def __init__(
...
-        else VisibilityChoice.default.value
+    _branch_code: int | None = (
+        kwargs.pop("_branch_code") if "_branch_code" in kwargs else 1
     )
-    if "is_new_version_of" in kwargs:
-        logger.warning("`is_new_version_of` will be removed soon, please use `revises`")
-        revises = kwargs.pop("is_new_version_of")
+    if "name" in kwargs:
+        key = kwargs.pop("name")
+        logger.warning(
```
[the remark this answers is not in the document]
falexwolf: You're right. Will do.

lamindb/_query_set.py
```diff
@@ -79,6 +78,29 @@ def one_helper(self):
         return self[0]
 
 
+def get_backward_compat_filter_kwargs(expressions):
```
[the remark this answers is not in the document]
falexwolf: Ah, but then I'd need to change this in 30 files. This is for another PR.

lamindb/_transform.py
```diff
@@ -20,23 +20,36 @@ def __init__(transform: Transform, *args, **kwargs):
...
     reference_type: str | None = (
         kwargs.pop("reference_type") if "reference_type" in kwargs else None
     )
-    if "is_new_version_of" in kwargs:
-        logger.warning("`is_new_version_of` will be removed soon, please use `revises`")
-        revises = kwargs.pop("is_new_version_of")
+    if "name" in kwargs:
+        if key is None:
+            key = kwargs.pop("name")
+            logger.warning(
```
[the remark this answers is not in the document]
falexwolf: Yes, will do.

lamindb/_record.py
Koncopd: Ok, thx, good to know.

lamindb/_transform.py
```diff
@@ -20,23 +20,36 @@ def __init__(transform: Transform, *args, **kwargs):
...
+                f"`name` will be removed soon, please pass '{key}' to `key` instead"
+            )
+        else:
+            if description is None:
+                description = kwargs.pop("name")
+                logger.warning(
```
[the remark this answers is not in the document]
falexwolf: Done!

lamindb/_transform.py
falexwolf: Fixed.

lamindb/_query_set.py
```diff
@@ -79,6 +78,29 @@ def one_helper(self):
...
+        "name": "key",  # backward compat <1.0
+        "visibility": "_branch_code",  # backward compat <1.0
+        "transform": "run__transform",  # for convenience (and backward compat <1.0)
+    }
+    mapped = {}
+    for field, value in expressions.items():
+        parts = field.split("__")
+        if parts[0] in name_mappings:
+            if parts[0] != "transform":
+                logger.warning(
```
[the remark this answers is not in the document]
falexwolf: Fixed.

PR #2326, opened 2025-01-06
🎨 Raise error when modifying key

sunnyosun:
If `artifact` has a non-virtual key or if the key doesn't have the same suffix an update will throw an `InvalidArgument` error:
```python
with pytest.raises(InvalidArgument) as error:
    artifact.key = "new_key"
    artifact.save()
```


lamindb/_record.py
```diff
@@ -740,14 +747,18 @@ def save(self, *args, **kwargs) -> Record:
                 revises._revises = None  # ensure we don't start a recursion
                 revises.save()
                 check_name_change(self)
+                check_key_change(self)
                 super(Record, self).save(*args, **kwargs)
                 _store_record_old_name(self)
+                _store_record_old_key(self)
```
falexwolf, line 753: Rather than making a dedicated function `_store_record_old_key`, can you please just add the two lines here:
```
    if isinstance(record, (Artifact, Transform)):
        record._key = record.key
```

Easier to read and more performant. In parts, we have two many auxiliary functions.

lamindb/_record.py
```diff
@@ -847,6 +864,32 @@ def check_name_change(record: Record):
                 raise RecordNameChangeIntegrityError
 
 
+def check_key_change(record: Record):
+    """Errors if a record's key has falsely changed."""
+    if not record.pk or record.key is None or not hasattr(record, "_key"):
```
falexwolf, line 869: Can you add comments on these conditions?

I don't understand the first two. The last two is moot if you have proper type checks for Transform and Artifact; please refactor.

If record.key is None and people set it to something invalid, that's also a problem; so I don't think this should return here, should it?

lamindb/_record.py
```diff
@@ -847,6 +864,32 @@ def check_name_change(record: Record):
...
+        return
+
+    old_key = record._key
+    new_key = record.key
+
+    old_key_suffix = (
+        record.suffix
```
falexwolf, line 876: Transform does not have `suffix`.

falexwolf: I'd only add a consistency check for artifact at the very end to ensure that the new suffix is still consistent.

For transforms this isn't necessary because we don't have the suffix column.

lamindb/_record.py
```diff
@@ -847,6 +864,32 @@ def check_name_change(record: Record):
...
+        if hasattr(record, "suffix")
+        else extract_suffix_from_path(PurePosixPath(old_key), arg_name="key")
+    )
+    new_key_suffix = extract_suffix_from_path(PurePosixPath(new_key), arg_name="key")
+
+    if new_key is not None and old_key != new_key:
+        if hasattr(record, "_key_is_virtual") and not record._key_is_virtual:
```
falexwolf, line 883: Please proper type check for Artifact.

lamindb/_record.py
```diff
@@ -847,6 +864,32 @@ def check_name_change(record: Record):
...
+            raise RecordKeyChangeIntegrityError(
```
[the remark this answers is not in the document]
sunnyosun: Because `key` is not passed as an argument? So I thought `InvalidArgument` isn't appropriate if it's not errored from init.

lamindb/_record.py
sunnyosun: But _store_record_old_key is used in 3 places, that's why I made a function.

falexwolf: Ok, fine!

lamindb/_record.py
falexwolf: If key was `None` people can set it, yes. It has then to be set for all members of the version family.

lamindb/_artifact.py
```diff
@@ -874,6 +874,7 @@ def replace(
         if key_path.name != new_filename:
             self._clear_storagekey = self.key
             self.key = str(key_path.with_name(new_filename))
+            self._key = self.key
```
Koncopd, line 877: @sunnyosun @falexwolf this is so so confusing, why calling it _key and not _old_key or _previous_key?

falexwolf: You're absolutely right, @Koncopd! I should have pointed this out to. We over-use "simply prefixing with _".

falexwolf: Can this be fixed, please?

falexwolf: I see it was fixed already somewhere.

PR #2332, opened 2025-01-08
✨ Also search git blob hash in remote branches

### LINKED ISSUE #2331: Improve git integration code blob search to also look at branches
### Add a description

https://laminlabs.slack.com/archives/C07DB677JF6/p1736336383055239

Zethson:
Fixes #2331 

**Before:**

Errors when the code blob was not on the default branch.

**After:**
```
code blob hash 313ce3a4116476f1e83a14c17d8d2dcb6138e141 was found in non-default branch(es): origin/feature/ruff
```

### DIFF
--- a/lamindb/core/_sync_git.py
+++ b/lamindb/core/_sync_git.py
@@ -74,11 +74,9 @@ def get_git_commit_hash(blob_hash: str, repo_dir: Path | None = None) -> str | N
     # Hence, we split by new line ("\n") and use the first one
     commit_hash = result.stdout.decode().split("\n")[0]
 
-    # Return None if no commit was found or if the log command failed
     if not commit_hash or result.returncode == 1:
         return None
 
-    # Get the default branch
     default_branch = (
         subprocess.run(
             ["git", "rev-parse", "--abbrev-ref", "origin/HEAD"],
@@ -91,7 +89,7 @@ def get_git_commit_hash(blob_hash: str, repo_dir: Path | None = None) -> str | N
     )
 
     # Find all branches containing the commit
-    branches = subprocess.run(
+    commit_containing_branches = subprocess.run(
         ["git", "branch", "--all", "--contains", commit_hash],
         capture_output=True,
         cwd=repo_dir,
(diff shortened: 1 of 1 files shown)


falexwolf (commented):
Oh my god, that became so much more complicated. 🤔 

9 tested simple lines become 30 untested lines that are hard to review. 🤔

falexwolf (approved):
Looking more at it I think it's fine. I would have tried to keep the untested parts (not finding the thing on the default branch) more minimal. But there is probably no way. And yes, it's perfectly understandable and well-written.

Of course we could alternatively think through how to set up the test but that's going to be hard.

Do you feel good about merging it? I think if you do I'm also good.

falexwolf:
> I actually don't think that it's that hard to review or understand...

You're right! Sorry! I judged too hastily from the mere amount of lines.

falexwolf:
Originally I also had the thought that one would just try it on the default branch kinda running the previous code so that one doesn't need to fetch potentially a large amount of data and make many calls. And then only if one doesn't find something one falls back to fetching etc.

But I ended up agreeing that given how hard the multi-branch scenario is to test right now, this is the safer way of writing it. The performance penalty will be OK unless people are working on a large mono-repo with many branches. That seems unlikely.

No need to overoptimize.

falexwolf:
One thing: When we construct the link to the commit, we need to reflect the branch name I think. 🤔

falexwolf:
No, I guess I'm wrong. The blob database exists independent of the branches. Can you briefly check for your example?

It's this line:

https://github.com/laminlabs/lamindb/blob/ce4e023b9a77c5ac366d21dc82d8471dd5617cf6/lamindb/core/_sync_git.py#L137

Zethson:
> Do you feel good about merging it? I think if you do I'm also good.

While writing I wasn't sure whether I'd want to add several lines of code just to get the branch name but in the end I think being explicit is better.

> No, I guess I'm wrong. The blob database exists independent of the branches. Can you briefly check for your example?
> 
> It's this line:
> 
> https://github.com/laminlabs/lamindb/blob/ce4e023b9a77c5ac366d21dc82d8471dd5617cf6/lamindb/core/_sync_git.py#L137

gives me `https://github.com/laminlabs/redun-lamin/blob/8e8dff0074130441e44a959fa79e2735bf6adb92/docs/workflow.py` which should be correct.

falexwolf:
Yeah, this exists: https://github.com/laminlabs/redun-lamin/blob/8e8dff0074130441e44a959fa79e2735bf6adb92/docs/workflow.py

Great! 😄 

This much nicer now. 

PS: Last correction is the PR title which should be "non-default" instead of "remote". I'll fix it. And ah, you already merged. Fine!

PR #2334, opened 2025-01-09
🥅 Clear error on running track in an interactive session

Koncopd:
Fix https://github.com/laminlabs/pfizer-lamin-usage/issues/109

Also upper bound `zarr`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -49,7 +49,7 @@ jupyter = [
     "mistune!=3.1.0",  # there is a bug in it
 ]
 zarr = [
-    "zarr>=2.16.0",
+    "zarr>=2.16.0,<3.0.0a0", # not yet compatible with 3.0.*
 ]
 fcs = [
     "readfcs>=1.1.9",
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -49,7 +49,7 @@ jupyter = [
     "mistune!=3.1.0",  # there is a bug in it
 ]
 zarr = [
-    "zarr>=2.16.0",
+    "zarr>=2.16.0,<3.0.0a0", # not yet compatible with 3.0.*
```
falexwolf, line 52: We should mention somewhere that people can actually run zarr v3; they just won't be able to use anndata zarr streaming internals.

It's perfectly fine to use zarr v3 artifacts like I'm sure many people will do. But given this is an extra it might not be too confusing.

falexwolf: I'm glad you put in the work long ago to separate this from the core package & dependencies, @Koncopd

Koncopd: No, they actually can't use zarr v3, the imports will error. Actually i think i even should add some additional version check to avoid that if zarr v3 is installed already.

Koncopd: Lets continue here https://github.com/laminlabs/lamindb/pull/2334

PR #2336, opened 2025-01-09
🐛 Prevent installed zarr v3 from crushing lamindb import

Koncopd:


Koncopd:
Another problem is that now installing `zarr v3` crushes `lamindb` import.
image
unfortunately can't do much here because this comes from `anndata`.

PR #2345, opened 2025-01-13
♻️ Refactor key and name change checks

Koncopd:


sunnyosun (approved):
Great, thank you Sergei!

Koncopd:
Some questions regarding the previous PR:
https://github.com/laminlabs/lamindb/blob/0671bdf5188c95118dee09ae31b701bd46b19fbe/lamindb/_record.py#L820
But `Collection` also has `.artifacts`, do we want to check this for `Collection`? Shouldn't the models be specified here explicitly?

Koncopd:
This has type https://github.com/laminlabs/lamindb/blob/0671bdf5188c95118dee09ae31b701bd46b19fbe/lamindb/_record.py#L867
`record: Artifact | Transform` but checks only for `Artifact`. Is it right?

Koncopd:
@falexwolf or @sunnyosun could you check the questions also please?

falexwolf:
I believe that all of Artifact, Collection & Transform should not just have a simple key change.

This here seems to not yet cover all the cases as it was written with Artifact in mind.

But for all of them, there should also be a check for the version family:

https://github.com/laminlabs/lamindb/blob/0671bdf5188c95118dee09ae31b701bd46b19fbe/lamindb/core/versioning.py#L15-L22

This is e.g. used in the Transform constructor:

https://github.com/laminlabs/lamindb/blob/0671bdf5188c95118dee09ae31b701bd46b19fbe/lamindb/_transform.py#L92-L94

So, yes, Sergei; the things you raise are not 100% thought through and I'd appreciate it if you refactored so that also the version family consistency is insured and this message is printed `message_update_key_in_version_family` (see link above).

Koncopd:
https://github.com/laminlabs/lamindb/blob/0671bdf5188c95118dee09ae31b701bd46b19fbe/lamindb/_record.py#L820
@falexwolf @sunnyosun what is the model that is checked here? I want to specify it explicitly. Is it `ULabel`?

sunnyosun:
> https://github.com/laminlabs/lamindb/blob/0671bdf5188c95118dee09ae31b701bd46b19fbe/lamindb/_record.py#L820
> 
> 
> @falexwolf @sunnyosun what is the model that is checked here? I want to specify it explicitly. Is it `ULabel`?

I fear this one can't be specified, this includes all models with `.artifacts` as many-to-many relationship, so essentially all bionty, wetlab, custom registries.

Koncopd:
Ok, git it, thx.

Koncopd:
@falexwolf i am not really familiar with the versioning code, so i will merge this and then can work on the versions key check for key change separately when i familiarize myself more with versioning. Or you can work on this when you have time - https://github.com/laminlabs/lamindb/issues/2347

PR #2349, opened 2025-01-13
🐛 Skip deleting storage when deleting outdated versions of folder-like artifacts

Koncopd:
Now by default, if not explicitly specified with `.delete(storage=True)`, deleting a non-latest version of a mutable `Artifact` (such as `tiledbsoma`) does not delete an underlying storage.

```python
artifact = ln.Artifact.filter(key="some.tiledbsoma", is_latest=False).first()
artifact.delete() # deletes only this specific version and doesn't delete artifact.path
```

Deleting the latest version now deletes all versions so that no artifact records without underlying data survive.
```python
artifact = ln.Artifact.filter(key="some.tiledbsoma", is_latest=True).one() # virtual key
artifact.delete() # deletes all versions and the data in storage
```

Internal [Slack thread](https://laminlabs.slack.com/archives/C07DB677JF6/p1736783041648239).

### DIFF
--- a/lamindb/_artifact.py
+++ b/lamindb/_artifact.py
@@ -1072,7 +1072,11 @@ def delete(
         # only delete in storage if DB delete is successful
         # DB delete might error because of a foreign key constraint violated etc.
         self._delete_skip_storage()
-        if self.key is None or self._key_is_virtual:
+        # by default do not delete storage if deleting only one version of multiple
+        # and the underlying store is mutable
+        if self._overwrite_versions and self.versions.count() > 1:
+            delete_in_storage = False if storage is None else storage
+        elif self.key is None or self._key_is_virtual:
             # do not ask for confirmation also if storage is None
             delete_in_storage = storage is None or storage
         else:
--- a/lamindb/_artifact.py
+++ b/lamindb/_artifact.py
@@ -1069,12 +1069,14 @@ def delete(
             # we can still delete the record
             logger.warning("Could not get path")
             storage = False
+        # check the number of versioned artifacts before deleting the current one
+        is_versioned_mutable = self._overwrite_versions and self.versions.count() > 1
         # only delete in storage if DB delete is successful
         # DB delete might error because of a foreign key constraint violated etc.
         self._delete_skip_storage()
         # by default do not delete storage if deleting only one version of multiple
         # and the underlying store is mutable
-        if self._overwrite_versions and self.versions.count() > 1:
+        if is_versioned_mutable:
             delete_in_storage = False if storage is None else storage
         elif self.key is None or self._key_is_virtual:
             # do not ask for confirmation also if storage is None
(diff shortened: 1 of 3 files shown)


lamindb/_artifact.py
```diff
@@ -1071,8 +1071,17 @@ def delete(
...
+        if self._overwrite_versions and self.is_latest:
+            # includes self
+            for version in self.versions.all():
+                _delete_skip_storage(version)
+        else:
+            self._delete_skip_storage()
+        # by default do not delete storage if deleting only a previous version
+        # and the underlying store is mutable
+        if self._overwrite_versions and not self.is_latest:
+            delete_in_storage = False if storage is None else storage
```
falexwolf, line 1083: Do we want to allow forcing a storage delete here? I don't think so.

I'd issue a warning and say "storage argument is ignored; can't delete store on a previous version".

Otherwise the user is left with a family of records that doesn't have any storage attached to it. That's not useful, is it?

Koncopd: Normally storage should not be specified at all, and it is, i assume a good reason to do this.

Koncopd: I also want `storage` to be fully consistent for all cases, while we have it.

falexwolf:
Given you now have all the logic, can you throw an error if you do any of the following?
```python
artifact = ln.Artifact.filter(key="some.tiledbsoma", is_latest=False).first()
artifact.cache() / .load() / .open()
```

Because all of these would load an inconsistent state.

lamindb/_artifact.py
falexwolf: If you don't want the warning, how about throwing an error? I can't imagine a single use case where you'd want to produce dangling records in that version family.

It's only going to be called by an inexperienced user who will produce a mess with this. Don't you agree?

Koncopd: For edge-case cleanups with corrupted version chain or something like this.

falexwolf: > a corrupted version chain or something like this

In that case the user should get an error telling them to delete the latest version so that the whole thing gets cleaned up.

You definitely don't want to keep around the corrupted version chain (or a similar case). Are you not afraid of the messiness this call will leave behind?

Koncopd: Ok, after some think i agree that you are right here, added ignore and a warning.

Koncopd:
>Given you now have all the logic, can you throw an error if you do any of the following?

Maybe a warning?

falexwolf:
> Maybe a warning?

I'd hard error because this is one of the situations where you'll produce an unexpected behavior and then you need to look ages for the bug in your book-long logs.

Warnings are for things that won't kill you when ignored. I'd say that returning an inconsistent state will necessarily produce a bug.

Koncopd:
You are right, i have added the exception.

falexwolf:
Good to merge from my end! 😄

PR #2353, opened 2025-01-15
🚚 Migrate to 1.0a6

falexwolf:
- Manage `created_at`, `updated_at` on the db-level, make `created_by not editable`
- Better migration strategy for FeatureSet -> Schema
- Add project labels to everything except run, storage & user so that people can see the things that are relevant to their project
- Rename transform `type` `"glue"` to `"linker"` [Slack thread](https://laminlabs.slack.com/archives/C046F63LJJ2/p1716994663729129)


lamindb/models.py
```diff
@@ -199,7 +205,9 @@ class TracksUpdates(models.Model):
     class Meta:
         abstract = True
 
-    updated_at: datetime = DateTimeField(auto_now=True, db_index=True)
```
falexwolf, line 202: Change from Django to server-side time stamps.

PR #2361, opened 2025-01-17
👷 Prettify `build.yml` and adjust coverage calculation if a job isn't run

falexwolf:
We're not always running the `curators` tests because they're almost like another package inside lamindb.


Koncopd:
btw what is with the coverage? Why is it so low now?

falexwolf:
> btw what is with the coverage? Why is it so low now?

curator tests are skipped if no files in curator module are changed; but coverage isn't respecting this; that's what I'm trying to fix here

falexwolf:
Do you spot why the coveragerc might not be respected, @Zethson?

Zethson:
Settings in a `project.toml` take precedence I think. Maybe this config is therefore still seen first and the specific one ignored? Not sure, I need to think further.

noxfile.py
```diff
@@ -157,6 +157,38 @@ def install_ci(session, group):
...
 
 
+@nox.session
+def configure_coverage(session) -> None:
+    """Write a coverage config file, adding extra patterns to omit."""
+    import tomlkit
+
+    groups_str = session.posargs[0]  # first positional argument
+
+    print(groups_str)  # for debugging
```
Zethson, line 167: Nit: Should this still be there?

noxfile.py
```diff
@@ -157,6 +157,38 @@ def install_ci(session, group):
...
+    # Update the omit patterns
+    base_patterns = config["tool"]["coverage"]["run"]["omit"]
+    all_patterns = base_patterns + extra_omit_patterns
+    config["tool"]["coverage"]["run"]["omit"] = all_patterns
+
+    # Write back to pyproject.toml
+    with open(base_config_path, "w") as f:
+        tomlkit.dump(config, f)
+
+    print(base_config_path.read_text())
```
Zethson, line 189: And this?

Zethson (commented):
It's fine if you want to keep those prints around for logging but I'm asking nevertheless

noxfile.py
falexwolf: Yes, we'll need this when we change the group names; this will soon happen as the CI architecture is far from optimal right now.

It's not super intuitive to pass these inputs from the other job around because it's just a long string from the CLI argument. So the easiest is to see it. :)

noxfile.py
falexwolf: Yeah, that's needed to check that things worked out the way they're supposed to work.

You'll not expand this step in the `test` job unless you want to debug it. That's why it's there.

PR #2370, opened 2025-01-23
🎨 Standardize irrespective of source

sunnyosun:


lamindb/_can_curate.py
```diff
@@ -59,6 +59,7 @@ def inspect(
     field: str | StrField | None = None,
     *,
     mute: bool = False,
+    strict: bool = False,
```
falexwolf, line 62: Are we sure we want to introduce yet another parameter for the user?

If so, please craft the docstring for it and provide an example in the docs.

falexwolf: The term `strict` is also potentially overloaded and a more specific name could be adequate; see e.g. the meaning in the pandera validation API: https://pandera.readthedocs.io/en/stable/dataframe_schemas.html

lamindb/_can_curate.py
```diff
@@ -144,7 +154,9 @@ def _inspect(
     using_key = queryset.db
     if isinstance(source, Record):
         _check_source_db(source, using_key)
-        queryset = queryset.filter(source=source).all()
+        # if not strict mode, do not filter by source here because we want to inspect all records in the DB first
```
falexwolf, line 157: Can the comment be more comprehensive, please?

"if in strict mode, restrict the query to one particular ontology source"

"otherwise, search across records from all ontology sources present in the registry"

lamindb/_can_curate.py
```diff
@@ -144,7 +154,9 @@ def _inspect(
...
+        if strict:
+            queryset = queryset.filter(source=source).all()
```
falexwolf, line 159: This is for chaining, the `.all()` is not needed (I'm pretty sure).

tests/core/test_can_validate.py
```diff
@@ -19,7 +19,42 @@ def test_inspect():
     assert result.validated == ["TCF7"]
 
     # clean up
-    bt.Gene.filter().all().delete()
+    bt.Gene.filter().delete()
+
+
+# if a record was added to the DB via a different source
+# it will still be validated because it's in the DB
+# we might want to introduce a strict mode in the future
```
falexwolf, line 27: The comment here is confusing because you're introducing a strict mode in this PR and even testing it in this test?

Can you please prettify the code in this test, e.g. by not repeating `bt.Source.get(name="2dfU")` 6 times?

falexwolf (changes requested):
Please ping me when you made changes for re-review! ☺️

lamindb/_can_curate.py
sunnyosun: No, it's just for backward compatibility with downstream packages behaviors now.

sunnyosun: I'm not adding to the docstring because I don't want users to use it yet, and I hope this is never needed.

falexwolf: But isn't this a user-facing method signature? If it is, it should be documented. If it shouldn't be user-facing, how about pre-fixing it with an underscore and/or making it part of kwargs?

lamindb/models.py
```diff
@@ -239,6 +239,7 @@ def inspect(
         mute: bool = False,
         organism: str | Record | None = None,
         source: Record | None = None,
+        strict: bool = False,
```
falexwolf, line 242: How about calling it `restrict_source`?

Issue #2346, opened 2025-01-13
Python 3.13 support


Zethson:
See https://github.com/laminlabs/lamindb/pull/2366 for details and updates.

ap--:
`3.13` support would be nice, now that the minimum supported version is `3.10`

PR #2366, opened 2025-01-21
✨ Support Python 3.13

Zethson:
Fixes https://github.com/laminlabs/lamindb/issues/2346#issuecomment-2603361418

- Adds official support for Python 3.13
- Replace 3.12 with 3.13 in the test matrix
- Require `readfcs>=2.0.1`

We stick to Python 3.12 for the docs for now because they fail https://github.com/laminlabs/lamindb/pull/2366#issuecomment-2607526154

PR #2371, opened 2025-01-23
✨ Python 3.13 support

Zethson:
Fixes https://github.com/laminlabs/lamindb/issues/2346#issuecomment-2603361418

- Adds official support for Python 3.13
- Replace 3.12 with 3.13 in the test matrix
- Require `readfcs>=2.0.1`

We stick to Python 3.12 for the docs for now because they fail https://github.com/laminlabs/lamindb/pull/2366#issuecomment-2607526154

We noticed CI slowdowns because of the tiledbsoma installation. Don't merge yet


Koncopd:
Maybe just leave unit-core on 3.12 for now?

Zethson:
> Maybe just leave unit-core on 3.12 for now?

I think that these are essential tests and we should test them with 3.13 to ensure that things work smoothly... Options:

1. Have 3.12 and 3.13 in the matrix. Not a fan.
2. Have `unit-storage` and `unit-core` run with 3.12. Not a fan because these are essential for 3.13.
3. Only run 3.13 on release branch. Hmm not sure

Koncopd:
>Have unit-storage and unit-core run with 3.12. Not a fan because these are essential for 3.13.

I think it has to be 3.11 even, because we also had the same problem with 3.12. Or it is fine now for 3.12?

Zethson:
> > Have unit-storage and unit-core run with 3.12. Not a fan because these are essential for 3.13.
> 
> I think it has to be 3.11 even, because we also had the same problem with 3.12. Or it is fine now for 3.12?

unit-core currently runs with 3.12 right? I have not tested unit-storage again with 3.12

Koncopd:
Yes, you are right.

Koncopd:
I think it is important to run at least some groups on 3.13, we can switch the rest later when the problem is resolved.

falexwolf:
Why not running 3.13 on the release branch only?

falexwolf:
But I'm fine with any of the solutions proposed!

Zethson:
I went with 3.13 only on the release branch now. 

@Koncopd 3.12 also seems to work for storage now.

Koncopd (approved):

PR #2374, opened 2025-01-23
✅ Update and test replace for folder artifacts

Koncopd:


lamindb/models.py
```diff
@@ -2791,7 +2791,7 @@ def from_dir(
 
     def replace(
         self,
-        data: UPathStr,
+        data: UPathStr | pd.DataFrame | AnnData | MuData,
```
falexwolf, line 2794: I find it not very intuitive that we allow all these other types here, but I guess this is legacy. Hm.

I'd say we should deprecate this, but let's discuss.

falexwolf:
I was worried that upon `.replace()` on a folder some previous files in the folder might rename as the `sync` does not remove old files.

I don't think you're testing for this here:
image

So, I'm still worried that the folder with replacement now contains corrupted zarr fragments.

Can you please add a test for this?

falexwolf:
Small nitpicks:

1. Ideally we eliminate all notebook-based tests that don't serve documentation. So, if possible, let's not pile onto these legacy notebooks.
2. `.zrad` is not a thing and it should be replaced with `.zarr` or `.anndata.zarr`

This here is great except that the `match` thing [isn't a great way](https://laminlabs.slack.com/archives/C04A0RMA0SC/p1737095974861219) to test error messages:
image

Koncopd:
Yes, thanks, good points, i will add expand the tests.
Lets use this notebook for now, because all the replace tests are here. We can move to a script later.

falexwolf:
Good! I fear you need to delete the previous folder content first so that this corruption won't happen.

Koncopd:
You are right indeed.

PR #2377, opened 2025-01-23
🐛 Further fixes and tests for Artifact.replace

Koncopd:
Delete an existing folder in a remote destination on upload of a new folder.

Also see comments in https://github.com/laminlabs/lamindb/pull/2374

### DIFF
--- a/lamindb/core/storage/paths.py
+++ b/lamindb/core/storage/paths.py
@@ -139,7 +139,23 @@ def store_file_or_folder(
     local_path = UPath(local_path)
     if not isinstance(storage_path, LocalPathClasses):
         # this uploads files and directories
-        create_folder = False if local_path.is_dir() else None
+        if local_path.is_dir():
+            create_folder = False
+            try:
+                # if storage_path already exists we need to delete it
+                # if local_path is a directory
+                # to replace storage_path correctly
+                is_dir_storage_path = (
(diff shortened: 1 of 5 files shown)


lamindb/core/storage/paths.py
```diff
@@ -139,7 +139,20 @@ def store_file_or_folder(
...
     if not isinstance(storage_path, LocalPathClasses):
         # this uploads files and directories
-        create_folder = False if local_path.is_dir() else None
+        if local_path.is_dir():
+            create_folder = False
+            try:
+                # if storage_path already exists we need to delete it
+                # if local_path is a directory
+                # to replace storage_path correctly
+                if storage_path.stat().as_info()["type"] == "directory":
```
falexwolf, line 148: It seems odd that neither line 149 nor 151 are covered by a test. Isn't this here core functionality introduced in the PR. I would have assumed that the add-replace-cache tests capture this.

Koncopd: Yes, indeed, very strange, I will check locally what is going on.

Koncopd: Ok, i was using a different suffix, so the deletion wasn't tested indeed, as the replacement had a different path. Fix is here https://github.com/laminlabs/lamindb/pull/2381

PR #2381, opened 2025-01-24
✅ Test folder replacement correctly

Koncopd:
It wasn't testing the replacement of the folder in the same remote path because the replacement has a different suffix.


falexwolf:
Will run tests in just a bit. Sorry, broke a test for the description formatting.

falexwolf:
Triggered tests!

falexwolf:
Good that this is fixed now!

PR #2384, opened 2025-01-24
⬆️ Python 3.13 support 

Zethson:
Third time's a charm

Fixes https://github.com/laminlabs/lamindb/issues/2346#issuecomment-2603361418

- Adds official support for Python 3.13
- Replace 3.12 with 3.13 in the test matrix
- Require `readfcs>=2.0.1`
- Unpin tiledbsoma

We stick to Python 3.12 for the docs for now because they fail https://github.com/laminlabs/lamindb/pull/2366#issuecomment-2607526154


falexwolf:
Sorry about the back and forth, @Zethson!

Here is a [thread](https://laminlabs.slack.com/archives/C07DB677JF6/p1737724939349949?thread_ts=1737723874.724279&cid=C07DB677JF6) with the screenshots of the failures, which are all in tiledbsoma.

.github/workflows/build.yml
```diff
@@ -67,10 +67,7 @@ jobs:
 
       - uses: actions/setup-python@v5
         with:
-          python-version: |
-            ${{ matrix.group == 'storage' && '3.10' ||
```
Koncopd, line 71: I think we should still run something on 3.10.

Zethson:
No more slowdown for `storage` or in general. Also triggered the curator CI which passes. This should be ready now.

@Koncopd do you want a lowerbound for tiledbsoma?

.github/workflows/build.yml
Zethson: What about adding a 3.10 job for the release branch?

Zethson:
I don't understand how this PR triggers

```
Warning, treated as error:
/opt/hostedtoolcache/Python/3.13.2/x64/lib/python3.13/site-packages/bionty/base/entities/_bfxpipeline.py:docstring of bionty.base.entities._bfxpipeline.BFXPipeline:6:Unexpected indentation.
  File "/opt/hostedtoolcache/Python/3.13.2/x64/bin/lndocs", line 10, in <module>
    sys.exit(main())
             ~~~~^^
  File "/opt/hostedtoolcache/Python/3.13.2/x64/lib/python3.13/site-packages/lndocs/__main__.py", line 211, in main
    remove_lines_with_db_args(
    ~~~~~~~~~~~~~~~~~~~~~~~~~^
        Path(args.site) / generated.with_suffix(".html").name
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/opt/hostedtoolcache/Python/3.13.2/x64/lib/python3.13/site-packages/lndocs/__main__.py", line 31, in remove_lines_with_db_args
    with open(path) as f:
         ~~~~^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '_build/html/lamindb.base.doc_args.html'
```

.github/workflows/build.yml
Koncopd: Whole CI is a lot. I think just running storage on 3.10 is enough.

Koncopd:
>@Koncopd do you want a lowerbound for tiledbsoma?

No, this would require adding it as a dependency.

.github/workflows/build.yml
Zethson: I mean we only run it on the release branch again and it's not that expensive. I added the whole CI on the release branch as a lower bound sanity check. I think it makes sense.

PR #2387, opened 2025-01-24
🐛 Fix SpatialDataCurator import

Zethson:


lamindb/curators/__init__.py
```diff
@@ -1775,7 +1775,7 @@ def from_spatialdata(
                 "Please install spatialdata: pip install spatialdata"
             ) from e
 
-        from ._spatial import SpatialDataCurator
+        from lamindb.curators._spatial import SpatialDataCurator
```
falexwolf, line 1778: Why was this necessary? 🤔 

If it's just for stylistic purposes, do we want to globally disable relative imports in ruff?

Zethson: I got an import error all of the sudden using the relative import. I was not working in the lamindb repository/folder :shrug:

Issue #2394, opened 2025-01-27
More helpful error messages or UX for setting descriptions of Artifacts when using `from_x`


Zethson:
### Add a description

I already caught myself running into this trap but a few users reported the same to me. We should therefore look for solutions:

```
ln.Artifact.from_df(df, "bla").save()
```

throws

```
...

InvalidArgument: The suffix '' of the provided key is incorrect, it should be '.parquet'.
```

the correct usage would have been

```
ln.Artifact.from_df(df, description="bla").save()
```

which many users seem to not want to spell out. The error message is not helpful in this case.

PR #2401, opened 2025-01-29
✨ Require keyword parameters for `Artifact.from_x` methods

Zethson:
Fixes https://github.com/laminlabs/lamindb/issues/2394

- Adds `*` to function signatures of `from_x` methods
- Fixes many deprecated `name` parameter args for `Transform` and `Collection`


falexwolf:
Good, this was also an unfortunate miss. I did the same for `curator.save_artifact()` in the other PR.

PR #2402, opened 2025-01-29
🚸 Better user feedback during keyword validation in `Record` constructor

### LINKED ISSUE #2390: Attempt to infer keyword arguments for record creation
### Add a description

Brought up by @nik-as

```
ValueError                                Traceback (most recent call last)
Cell In[199], line 1
----> 1 bt.Organism("human")

File ~/miniforge3/envs/scp_dev/lib/python3.10/site-packages/bionty/models.py:573, in Organism.__init__(self, *args, **kwargs)
    568 def __init__(
    569     self,
    570     *args,
    571     **kwargs,
    572 ):
--> 573     super().__init__(*args, **kwargs)

File ~/miniforge3/envs/scp_dev/lib/python3.10/site-packages/bionty/models.py:224, in BioRecord.__init__(self, *args, **kwargs)
    219             raise ValueError(
    220                 "Invalid parents kwarg passed. Provide a list of ontology ids."
    221             )
    222         self._parents = parents
--> 224 super().__init__(*args, **kwargs)

File ~/miniforge3/envs/scp_dev/lib/python3.10/site-packages/lamindb/_record.py:221, in __init__(record, *args, **kwargs)
    219             raise FieldValidationError(message) from e
    220 elif len(args) != len(record._meta.concrete_fields):
--> 221     raise ValueError("please provide keyword arguments, not plain arguments")
    222 else:
    223     # object is loaded from DB (**kwargs could be omitted below, I believe)
    224     super(BasicRecord, record).__init__(*args, **kwargs)

ValueError: please provide keyword arguments, not plain arguments
```

does not list all valid arguments. Maybe we can always infer them and show them here. We'd have to be very certain that the solution generalizes well enough...

Zethson:
fixes #2390 

Before:

```
ValueError                                Traceback (most recent call last)
Cell In[199], line 1
----> 1 bt.Organism("human")

File ~/miniforge3/envs/scp_dev/lib/python3.10/site-packages/bionty/models.py:573, in Organism.__init__(self, *args, **kwargs)
    568 def __init__(
    569     self,
    570     *args,
    571     **kwargs,
    572 ):
--> 573     super().__init__(*args, **kwargs)

File ~/miniforge3/envs/scp_dev/lib/python3.10/site-packages/bionty/models.py:224, in BioRecord.__init__(self, *args, **kwargs)
    219             raise ValueError(
    220                 "Invalid parents kwarg passed. Provide a list of ontology ids."
    221             )
    222         self._parents = parents
--> 224 super().__init__(*args, **kwargs)

File ~/miniforge3/envs/scp_dev/lib/python3.10/site-packages/lamindb/_record.py:221, in __init__(record, *args, **kwargs)
    219             raise FieldValidationError(message) from e
    220 elif len(args) != len(record._meta.concrete_fields):
--> 221     raise ValueError("please provide keyword arguments, not plain arguments")
    222 else:
    223     # object is loaded from DB (**kwargs could be omitted below, I believe)
    224     super(BasicRecord, record).__init__(*args, **kwargs)

ValueError: please provide keyword arguments, not plain arguments
```

After

```
An exception has occurred, use %tb to see the full traceback.

SystemExit: Use keyword arguments instead of positional arguments, e.g.: Organism(name='...')
```

This PR further automates the inference of model parameters for specific logging messages.

## Notes

1. I originally just modified the message but I think that we can hide more of the traceback here. No one should care about these internals.
2. The "blacklist" aka the fields that are not printed may not be exhaustive. I wonder what people think about it? If we fear that maintaining such a blacklist does not work, we can just suggest the example which is a small improvement.


Zethson:
@falexwolf more of a "what do you think in general" review request (see notes)

lamindb/_record.py
```diff
@@ -168,68 +169,98 @@ def suggest_records_with_similar_names(
...
-                if "version" in kwargs:
-                    if kwargs["version"] is not None:
-                        version_comment = " and version"
-                        existing_record = record.__class__.filter(
-                            **{
-                                name_field: kwargs[name_field],
-                                "version": kwargs["version"],
-                            }
-                        ).one_or_none()
+    try:
```
falexwolf, line 172: Because the code below is already heavily indented and deserves to be refactored, can you please separate this whole new try except logic into a new function?

falexwolf:
> @falexwolf more of a "what do you think in general" review request (see notes)

Hm, but wait, I thought this would address something else. 🤔

lamindb/_record.py
```diff
@@ -168,68 +169,98 @@ def suggest_records_with_similar_names(
...
-            try:
-                if hasattr(record, "clean_fields"):
-                    record.clean_fields()
-                else:
-                    record._Model__clean_fields()
-            except DjangoValidationError as e:
-                message = _format_django_validation_error(record, e)
-                raise FieldValidationError(message) from e
-    elif len(args) != len(record._meta.concrete_fields):
-        raise ValueError("please provide keyword arguments, not plain arguments")
```
falexwolf, line 227: I see that this was ugly and it needs to be improved.

lamindb/_record.py
```diff
@@ -168,68 +169,98 @@ def suggest_records_with_similar_names(
...
+                try:
+                    if hasattr(record, "clean_fields"):
+                        record.clean_fields()
+                    else:
+                        record._Model__clean_fields()
+                except DjangoValidationError as e:
+                    message = _format_django_validation_error(record, e)
+                    raise FieldValidationError(message) from e
+        elif len(args) != len(record._meta.concrete_fields):
+            valid_field_names = [
```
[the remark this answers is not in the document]
falexwolf: If you wanna do something like this, I'd get the valid keyword arguments from the constructor signature.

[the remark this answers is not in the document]
falexwolf: The constructor signature is the only reliable source for this information: https://github.com/laminlabs/lamindb/blob/cf1580f0538d1f58a163eed3c54ff06bcf710b23/lamindb/models.py#L2575-L2591

The fields are not.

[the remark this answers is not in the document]
Zethson: Yeah I was worried about that. Niklas wanted the valid fields to be printed out but I think that's messy and not that easy to generalize.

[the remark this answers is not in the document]
falexwolf: Now, because Django model are very unconventional Python classes we're doing all this work with overloading and manually validating dynamic `**kwargs`.

The user however should get the typed experience.

This will break down for custom registries if people defining these custom registries don't define a constructor signature.

So, it's really hard to get the valid keyword arguments.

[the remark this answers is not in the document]
falexwolf: TLDR: I'd not attempt this but rather throw a prettier error that refers the user to the definition of the class. For everything we do, stuff is spelled out and typed and this problem shouldn't occur anyway (even though I'd need to check bionty).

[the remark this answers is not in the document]
falexwolf: Yeah, so bionty also has the proper signatures where intellisense etc. should avoid the problem already:

- https://github.com/laminlabs/bionty/blob/f53df550bbef9819d4dcf25f7920d62046f5e62d/bionty/models.py#L102-L113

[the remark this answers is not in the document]
Zethson: It's not in Bionty but I'll take us there. See e.g. https://github.com/laminlabs/bionty/pull/179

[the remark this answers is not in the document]
falexwolf: I don't know of any other python package that prints out a list of valid keyword arguments upon failure. So, now I'm pretty convinced we shouldn't do this but rather ensure that we provide the proper typed & correct signatures everywhere.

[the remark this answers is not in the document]
falexwolf: > Niklas wanted the valid fields to be printed out but I think that's messy and not that easy to generalize.

Yeah, he is wrong and should switch on intellisense.

lamindb/_record.py
```diff
@@ -168,68 +169,98 @@ def suggest_records_with_similar_names(
...
+            sys.exit(
+                f"Don't pass a string directly. Use: {example}(name='...')\nValid fields: {valid_field_names}"
+            )
+        else:
+            # object is loaded from DB (**kwargs could be omitted below, I believe)
+            super(BasicRecord, record).__init__(*args, **kwargs)
+            _store_record_old_name(record)
+            _store_record_old_key(record)
+    except (TypeError, AttributeError) as e:
+        if "missing required argument" in str(
```
falexwolf, line 259: I think this is too layered. The "missing required argument" error message is already fine-tuned to the user and exits with a ValidationError (I'm pretty sure). So, I don't even get why this would occur under TypeError, AttributeError 🤔

falexwolf: Btw, every core lamindb Record class already prints the valid keyword arguments.

For instance here:

image

But same for artifact, collection, transform etc. 

So this whole effort here is double.

This being said: the validation in lamindb constructors is done manually rather than based on the overloaded signature; so I imagine this could be simplified.

lamindb/_feature.py
```diff
@@ -154,7 +155,8 @@ def __init__(self, *args, **kwargs):
     is_type: bool = kwargs.pop("is_type") if "is_type" in kwargs else False
     type_: Feature | str | None = kwargs.pop("type") if "type" in kwargs else None
     if kwargs:
-        raise ValidationError("Only name, dtype, is_type are valid keyword arguments")
+        valid_keywords = ", ".join([val[0] for val in _get_model_params(Feature)])
```
falexwolf, line 158: This seems nice! 😄

lamindb/_record.py
```diff
@@ -224,14 +229,25 @@ def __init__(record: Record, *args, **kwargs):
                 message = _format_django_validation_error(record, e)
                 raise FieldValidationError(message) from e
     elif len(args) != len(record._meta.concrete_fields):
-        raise ValueError("please provide keyword arguments, not plain arguments")
+        sys.exit(
```
falexwolf, line 232: Our convention is `raise SystemExit` or ideally one of the errors that inherit from it.

My hope is still that we find another way of surpressing the traceback (I have a thread somewhere experimenting with different Exception classes) and then it'll be hard to look for all occurances of `sys.exit()` in the code base and replace it with that solution.

lamindb/_record.py
```diff
@@ -224,14 +229,25 @@ def __init__(record: Record, *args, **kwargs):
...
         _store_record_old_name(record)
         _store_record_old_key(record)
 
 
+def __init__(record: Record, *args, **kwargs):
+    try:
+        _init_with_potential_errors(record, *args, **kwargs)
+    except ValueError as e:
+        if "missing required argument" in str(e):
+            sys.exit(str(e))
```
falexwolf, line 247: raise SystemExit or one of the custom exceptions

falexwolf: Wait, no. Instead of doing this we should update the ValueError that is raised this in the first place with a `FieldValidationError` (and move this into the exceptions: https://github.com/laminlabs/lamindb/blob/bf44bbf579ca403a1c40735f4d3174ed58240dac/lamindb/base/validation.py#L9; it's currently here because of lnschema-core migration).

I see that according to our code it should be a `TypeError` anyway. But yeah, please simply replace with FieldValidationError and then the __init__ can stay as it is: https://github.com/laminlabs/lamindb/blob/bf44bbf579ca403a1c40735f4d3174ed58240dac/lamindb/_record.py#L108

lamindb/_record.py
```diff
@@ -263,6 +279,48 @@ def _format_django_validation_error(record: Record, e: DjangoValidationError):
         return message
 
 
+def _get_model_params(record_class) -> list[tuple[str, str]]:
```
falexwolf, line 282: Can you please rename this to `_get_record_kwargs()` or `_get_record_keyword_arguments()` or `_get_record_constructor_arguments()` or something like this?

We want to clearly separate the Django Model layer also in naming from what we add on top with `Record` / `BasicRecord`.

And `params` has a specific meaning in lamindb (run params, model params as in ML model params), that's one reason for why we have Google style and not NumpyDoc style.

lamindb/_record.py
```diff
@@ -263,6 +279,48 @@ def _format_django_validation_error(record: Record, e: DjangoValidationError):
...
+    source = inspect.getsource(record_class)
+
+    # Find first overload that's not *db_args
+    pattern = r"@overload\s+def __init__\s*\(([\s\S]*?)\):\s*\.{3}"
+    overloads = re.finditer(pattern, source)
+
+    for overload in overloads:
+        params_block = overload.group(1)
+        if "*db_args" in params_block:
+            continue
```
falexwolf, line 298: That's weird that this is never called in tests. Is there something wrong here?

lamindb/_record.py
```diff
@@ -263,6 +279,48 @@ def _format_django_validation_error(record: Record, e: DjangoValidationError):
...
+        params = []
+        for line in params_block.split("\n"):
+            line = line.strip()
+            if not line or "self" in line:
+                continue
+
+            # Extract name and type annotation
+            param_pattern = (
+                r"(\w+)\s*:\s*((?:[^=,]|(?<=\[)[^[\]]*(?=\]))+)(?:\s*=\s*([^,]+))?"
+            )
```
falexwolf, line 309: This looks a bit scary. Can this break?

lamindb/_record.py
```diff
@@ -263,6 +279,48 @@ def _format_django_validation_error(record: Record, e: DjangoValidationError):
...
+                continue
+
+            name, type_str = match.group(1), match.group(2).strip()
+
+            # Keep type as string instead of evaluating
+            params.append((name, type_str))
+
+        return params
+
+    return []
```
falexwolf, line 321: Why is this not called in tests?

falexwolf:
Given you're already doing this, can you please do it for all Record constructors?

- https://github.com/laminlabs/lamindb/blob/bf44bbf579ca403a1c40735f4d3174ed58240dac/lamindb/_artifact.py#L564-L567
- https://github.com/laminlabs/lamindb/blob/bf44bbf579ca403a1c40735f4d3174ed58240dac/lamindb/_collection.py#L121

Etc.

And at least manually check that the result is as expected (matches what's currently hard-coded)?

falexwolf (changes requested):
This is potentially a cool UX improvement and consolidation of the keyword argument check if what's currently called `_get_model_params()` works robustly.

I recently had a case where @Koncopd didn't see a PR review due to what seems to be a bug on GitHub; so I hope you see all comments I left.

lamindb/_record.py
Zethson: Agree, done.

lamindb/_record.py
Zethson: No we just want to ensure that we don't accidentally look at the signature that has `*db_args` in the args. I would have to find a Record class where this overload signature comes first. We currently don't have that but this makes the code safer if someone accidentally does that.

falexwolf: Ok, understood! Can you please add this is as a comment in the code?

lamindb/_record.py
Zethson: I'm not sure whether I got what you're saying here correctly despite me reading it a couple of times.

1. I moved the `FieldValidationError` to `exceptions.py`.
2. We are now raising a `FieldValidationError` instead of a `ValueError`.
3. The `__init__` now stays as it was.

This gives us this experience:

```
An exception has occurred, use %tb to see the full traceback.

FieldValidationError: Use keyword arguments instead of positional arguments, e.g.: Organism(name='...'). See model documentation for valid fields.
```

(for `bt.Organism("bla")`)

Zethson:
Added the automatic inference to `Artifact` and `Collection. Concerning `Collection`, it used to be:

> 'artifacts, key, run, description, reference, reference_type'

which has now become 

> 'artifacts, name, description, meta, reference, reference_type, run, revises'

The new behavior is more correct but a tiny bit more verbose.

lamindb/_record.py
falexwolf: This reads correct. I'm looking at the code now.

lamindb/_record.py
```diff
@@ -105,7 +106,7 @@ def validate_fields(record: Record, kwargs):
         k for k, v in kwargs.items() if v is None and k in required_fields
     ]
     if missing_fields:
-        raise TypeError(f"{missing_fields} are required.")
+        raise FieldValidationError(f"{missing_fields} are required.")
```
falexwolf, line 109: Good!

lamindb/_record.py
```diff
@@ -224,7 +224,9 @@ def __init__(record: Record, *args, **kwargs):
                 message = _format_django_validation_error(record, e)
                 raise FieldValidationError(message) from e
     elif len(args) != len(record._meta.concrete_fields):
-        raise ValueError("please provide keyword arguments, not plain arguments")
+        raise FieldValidationError(
```
falexwolf, line 227: Good!

lamindb/_record.py
```diff
@@ -224,7 +224,9 @@ def __init__(record: Record, *args, **kwargs):
...
+            f"Use keyword arguments instead of positional arguments, e.g.: {record.__class__.__name__}(name='...'). See model documentation for valid fields."
```
falexwolf, line 228: > See model documentation for valid fields.

Simply remove. There are no "models" in lamindb (it'd be called registry) but it's also evident. And typically they should just switch on intellisense rather than looking into the docs.

lamindb/_ulabel.py
```diff
@@ -31,9 +32,8 @@ def __init__(self, *args, **kwargs):
         kwargs.pop("reference_type") if "reference_type" in kwargs else None
     )
     if len(kwargs) > 0:
-        raise ValueError(
-            "Only name, description, reference, reference_type are valid keyword arguments"
-        )
+        valid_keywords = ", ".join([val[0] for val in _get_record_kwargs(ULabel)])
+        raise ValueError(f"Only {valid_keywords} are valid keyword arguments")
```
falexwolf, line 36: Why is this a ValueError and not a FieldValidationError or ValidationError?

lamindb/_record.py
Zethson: Done.

lamindb/_ulabel.py
Zethson: Now all FieldValidationError. You were looking at my PR before I rerequested a review ^_^

Zethson:
Why does the migration check fail? I cannot reproduce that locally

PR #2403, opened 2025-01-29
🚚 Move CellXGene `Curator` from `cellxgene-lamin` here

falexwolf:
This PR merely moves code without yet refactoring it.

It moves the `CellxGeneCurator` from `cellxgene_lamin` here and it moves the `SpatialDataCurator` from the separate file `curators._spatial.py` into `curators/__init__.py`.

This is to prepare a joint refactor of all `Curator` classes.

- https://github.com/laminlabs/cellxgene-lamin/pull/120


Koncopd:
Why don't we split the gigantic __init__.py for curators?

falexwolf:
> Why don't we split the gigantic __init__.py for curators?

To prepare a refactor.

PR #2406, opened 2025-01-30
🚸 Fix warning about artifacts in trash

ap--:
Fixes the warning message for artifacts in trash.

Though it did sound like a cool dis before 😄


falexwolf:
:laughing:

PR #2408, opened 2025-01-30
🚚 Move `PertCurator` from `wetlab` here and add `CellxGene` `Curator` test

falexwolf:
Is part of a sequence of PRs that refactors the curators:

- https://github.com/laminlabs/lamindb/pull/2403
- https://github.com/laminlabs/lamindb/pull/2388

Made a "micro-farland2020" dataset to speed up the tests that were running in the `wetlab` repository: https://lamin.ai/laminlabs/lamindata/transform/JreAW9tkAHrc

- https://github.com/laminlabs/lamindb/pull/2408/

Also added a test for the CxG curator based on the notebook in `cellxgene-lamin`.

- https://github.com/laminlabs/cellxgene-lamin/pull/120

PR #2412, opened 2025-02-01
🎨 Self-contained definition of `CellxGene` schema / validation constraints

falexwolf:
Is part of a sequence of PRs that refactors the curators:

- https://github.com/laminlabs/lamindb/pull/2408
- https://github.com/laminlabs/lamindb/pull/2403
- https://github.com/laminlabs/lamindb/pull/2388

---

This PR simplifies the definition of constraints to validate against the CxG schema:

Before | After
--- | ---
It was implicitly defined by the state of another database instance passed via the `using_key` parameter to `AnnDataCurator`. |  It is now self-contained in the `CellxGeneAnnDataCurator` class.

It also brings down the run time of the CxG test further down [from 50s](https://github.com/laminlabs/lamindb/pull/2408#issuecomment-2628796391) to 20s:

```
============================= slowest 50 durations =============================
37.28s call     tests/curators/test_pert_curator.py::test_pert_curator
34.08s call     tests/curators/test_cat_curators.py::test_spatialdata_curator
19.95s call     tests/curators/test_cxg_curator.py::test_cxg_curator
6.86s call     tests/curators/test_cat_curators.py::test_df_curator
6.01s call     tests/curators/test_cat_curators.py::test_soma_curator
5.71s call     tests/curators/test_cat_curators.py::test_soma_curator_genes_columns
```

Collateral:

- I'm removing the `_check_valid_keys()` utils function because throwing an error upon Curator init upon missing columns/features is problematic: we do want to be able to validate datasets with missing columns and then receive this as a SchemaError during `.validate()`. This is a common type of error.
- Fixed the signatures for `Feature`, `ULabel` & `Schema` to now include `type` and `is_type`.

### DIFF
--- a/lamindb/curators/__init__.py
+++ b/lamindb/curators/__init__.py
@@ -380,13 +380,13 @@ class DataFrameCatCurator(BaseCurator):
                 raise TypeError(f"{name} must be a dictionary!")
             valid_keys = set(self._df.columns) | {"columns"} | extra
             nonval_keys = [key for key in d.keys() if key not in valid_keys]
-            n = len(nonval_keys)
-            s = "s" if n > 1 else ""
-            are = "are" if n > 1 else "is"
-            if len(nonval_keys) > 0:
-                raise ValidationError(
-                    f"key{s} passed to {name} {are} not present in columns: {colors.yellow(_format_values(nonval_keys))}"
-                )
+            len(nonval_keys)
+            # throwing an error here is problematic because my dataset might have missing columns
+            # and I still want to define sources mappings on the Curator level
+            # if len(nonval_keys) > 0:
+            #     raise ValidationError(
+            #         f"key{s} passed to {name} {are} not present in columns: {colors.yellow(_format_values(nonval_keys))}"
+            #     )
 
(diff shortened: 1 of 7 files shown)


lamindb/curators/__init__.py
```diff
@@ -380,13 +380,13 @@ def _check_valid_keys(self, extra: set | None = None) -> None:
...
-            n = len(nonval_keys)
-            s = "s" if n > 1 else ""
-            are = "are" if n > 1 else "is"
-            if len(nonval_keys) > 0:
-                raise ValidationError(
-                    f"key{s} passed to {name} {are} not present in columns: {colors.yellow(_format_values(nonval_keys))}"
-                )
+            len(nonval_keys)
+            # throwing an error here is problematic because my dataset might have missing columns
+            # and I still want to define sources mappings on the Curator level
```
falexwolf, line 385: I don't know who wrote this, so FYI @Zethson @sunnyosun -- I don't think this makes sense for the reason defined in the comment

We definitely want to be able to pass datasets with missing columns to the Curator constructor without raising an error

PR #2432, opened 2025-02-06
🚸 Full guarantee of avoiding duplicating `Transform`, `Artifact` & `Collection` in concurrent runs

falexwolf:
# 🚸 Guarantee of no duplication

**Before**

It was possible to create transforms, artifacts, and collections with the same hash when _creating them at the same time_ in concurrent logic. The exact concurrency enabled bypassing the built-in hash-based existence check in the `Transform`, `Artifact` & `Collection` constructors.

**After**

There now is a hash uniqueness constraint on the database level. This hard-codes the impossibility of storing multiple transforms, artifacts, and collections with the same hash.

In case hash uniqueness _would_ be violated, e.g. by attempting to create the same objects at the same time in concurrent logic, `.save()` does **not** error but instead returns the matching transform from the database:
```python
transform = Transform(...).save()
```

This is particularly useful for writing `get_or_create`-like logic and was needed to make `ln.tracked()` thread-safe.

# 🎨 Enabling the creation of non-validated records

In several internal parts of the code, the non-thread-safe global variable `ln.settings.creation.search_names` is used.

Instead of using a global setting, in internal parts of the code, one can now pass `_skip_validation` to any `Record` constructor. This allowed to eliminate the pattern for `CanCurate.from_values()`.

# 👷 Testing for logging messages

One can now easily test for logging messages using the `ccaplog` fixture, e.g.

```python
def test_ulabel_plural_type_warning(ccaplog):
    ln.ULabel(name="MyThings", is_type=True)
    assert (
        "name 'MyThings' for type ends with 's', in case you're naming with plural, consider the singular for a type name"
        in ccaplog.text
    )
```

# Others

* If an artifact is found in trash during hash-lookup, rather than throwing an error, it is now being restored (addresses https://github.com/laminlabs/pfizer-lamin-usage/issues/138)
* The unused settings and documentation about the behavior upon hash collisions for artifacts were removed. After 2.5 years, there hasn't been a single use case in which a user wanted to trigger an error or duplicate an artifact. Replication across storage regions is going to be handled differently.
* Creating params & features & ulabels with `is_type` is True has been improved.
* Re-vive the `Migration` registry under `.models`
* Add a few migrations with small non-breaking fixes; these will be re-squashed in the next PR.

# Materials

Internal [Slack thread](https://laminlabs.slack.com/archives/C04FPE8V01W/p1738826122789549).


lamindb/_from_values.py
```diff
@@ -110,7 +109,7 @@ def get_or_create_records(
         #         logger.debug(f"added default feature '{feature_name}'")
         return RecordList(records)
     finally:
-        settings.creation.search_names = creation_search_names
+        settings.creation.search_names = True  # need to refactor the whole function because can no longer use settings; not thread safe
```
falexwolf, line 112: @sunnyosun, we gotta eliminate the use of settings in `from_values()` entirely.

This is a big refactor in which we can also look into the dangers of creating duplicated records and similar.

I don't want to make this refactor here in this PR as it'd get too complicated (and I'm not so familiar with the `_from_values.py` file.

falexwolf: Slack thread: https://laminlabs.slack.com/archives/C04FPE8V01W/p1739365084531499

lamindb/_from_values.py
```diff
@@ -276,7 +276,7 @@ def create_records_from_source(
             else {"source": source}
         )
         for bk in bionty_kwargs:
-            records.append(model(**bk, **create_kwargs))
+            records.append(model(**bk, **create_kwargs, _skip_validation=True))
```
falexwolf, line 279: This modification is the main solution to the search_names problem.

falexwolf: Slack thread: https://laminlabs.slack.com/archives/C04FPE8V01W/p1739365084531499

lamindb/_record.py
```diff
@@ -76,6 +77,30 @@
 IPYTHON = getattr(builtins, "__IPYTHON__", False)
 
 
+def is_approx_pascal_case(s):
+    """Check if the last component of a dotted string is in PascalCase.
+
+    Args:
+        s (str): The string to check
```
Zethson, line 84: No typehints in docstrings, please. LLMs love it IK

tests/core/test_tracked.py
```diff
@@ -0,0 +1,79 @@
...
+        {"id": range(chunk_id * 10, (chunk_id + 1) * 10), "value": range(10)}
+    )
+
+    # Save it as an artifact
+    key = f"chunk_{chunk_id}.parquet"
+    artifact = ln.Artifact.from_df(df, key=key).save()
+    return artifact.key
+
+
+def test_tracked_parallel():
```
Zethson, line 21: There's a lot of prints here that clutter when we run `pytest -s`. Can we remove them again?

tests/core/test_tracked.py
```diff
@@ -0,0 +1,79 @@
...
+    for artifact in artifacts:
+        artifact.delete(permanent=True)
+
+    ln.context._uid = None
+    ln.context._run = None
+    ln.context._transform = None
+    ln.context._path = None
+
+
+if __name__ == "__main__":
```
Zethson, line 78: This should not be necessary with pytest

lamindb/_record.py
falexwolf: As this is _not_ rendered on the docs, isn't it better to have it? We also need it for the `__init__` because of sphinx type hints can't deal with overloads.

But I don't have a strong opinion, so also happy to remove them next time.

tests/core/test_tracked.py
falexwolf: Don't you want to see the prints when running `pytest -s`? 😅

lamindb/_record.py
Zethson: I'm confused. Why not just use a proper typehint for the parameter?

tests/core/test_tracked.py
Zethson: Yes but only the prints that I care about and not prints in many tests haha. Sure - fair game if I only run this test, but if we start adding prints to many tests, it's not the nicest UX.

tests/core/test_tracked.py
falexwolf: It's absolutely not but it's really helping understanding this file because it's a script that makes sense standalone.

I haven't yet gotten around to formalizing this more but when we want to given an example on the docs about a multi-threaded script, this file is what I'd like to show and what people should run.

I'm not sooooo happy with the strategy of having code in the `docs` folder like here https://github.com/laminlabs/lamindb/blob/main/docs/scripts/run-workflow.py

So, I left the `__main__` in there just as a reminder and first example to experiment with how to better use the content of the `tests` folder in the docs upon the next docs. tiangolo does this super nicely in all his projects; all code that's shown in the docs actually lives in the tests. It's initially much more work than what we're doing but leads to much less code & examples to maintain leading overall to better quality of the material.

lamindb/_record.py
falexwolf: Oh, sure, that's an omission. There should be a `str` there.

PR #2437, opened 2025-02-07
⬇️ Pin mistune to <3.1.0

Zethson:
3.1.1 got released which also doesn't work. https://pypi.org/project/mistune/


Koncopd:
What is the problem actually?

Zethson:
I am confused, I cannot reproduce it anymore. I had several errors for `ln.finish()` that went immediately away when I downgraded mistune. I'll keep observing...

Zethson:
Have not observed this again. No idea what happened back then

PR #2441, opened 2025-02-10
✨ Allow passing `artifact` to `Curator`

sunnyosun:
This PR enables passing an `Artifact` object to `Curator`.

If an `artifact` is passed, `.save_artifact` will not re-upload the artifact, but merely link all metadata records to it.

Calling `standardize()` errors when an artifact is passed, because this will mutate the dataset and therefore no longer matches the saved artifact.

(Note: passing `artifact` was already possible for `TiledbsomaCatCurator` before this PR).

### DIFF
--- a/lamindb/curators/__init__.py
+++ b/lamindb/curators/__init__.py
@@ -175,14 +175,6 @@ class Curator:
     def __init__(self, dataset: Any, schema: Schema | None = None):
         self._artifact: Artifact = None  # pass the dataset as a non-curated artifact
         self._dataset: Any = dataset  # pass the dataset as a UPathStr or data object
-        if isinstance(dataset, Artifact):
-            self._artifact = dataset
-            # tiledbsoma logic is dealt in TiledbsomaCatCurator
-            if (
-                self._artifact.otype != "tiledbsoma"
-                or self._artifact.suffix != ".tiledbsoma"
-            ):
-                self._dataset = self._artifact.load()
         self._schema: Schema | None = schema
         self._is_validated: bool = False
         self._cat_curator: CatCurator = None  # is None for CatCurator curators
@@ -205,6 +197,12 @@ class Curator:
         # Note that this docstring has to be consistent with the Artifact()
         # constructor signature
         pass
+
+    def _set_dataset_and_artifact(self):
+        if isinstance(self._dataset, Artifact):
+            self._artifact = self._dataset
+            if self._artifact.otype in {"DataFrame", "AnnData"}:
+                self._dataset = self._dataset.load()
 
 
 class DataFrameCurator(Curator):
@@ -254,6 +252,7 @@ class DataFrameCurator(Curator):
         schema: Schema,
     ) -> None:
         super().__init__(dataset=dataset, schema=schema)
+        self._set_dataset_and_artifact()
         if schema.n > 0:
             # populate features
             non_categoricals = {}
@@ -406,6 +405,7 @@ class AnnDataCurator(Curator):
         schema: Schema,
     ) -> None:
         super().__init__(dataset=dataset, schema=schema)
+        self._set_dataset_and_artifact()
         if not data_is_anndata(self._dataset):
             raise InvalidArgument("dataset must be AnnData-like.")
         if schema.otype != "AnnData":
@@ -472,6 +472,7 @@ class CatCurator(Curator):
         self, *, dataset, categoricals, sources, organism, exclude, columns_field=None
     ):
         super().__init__(dataset=dataset)
+        self._set_dataset_and_artifact()
         self._categoricals = categoricals or {}
         self._non_validated = None
         self._organism = organism
@@ -839,10 +840,6 @@ class AnnDataCatCurator(CatCurator):
 
         if sources is None:
             sources = {}
-        if not data_is_anndata(data) and not isinstance(data, Artifact):
-            raise TypeError(
-                "data has to be an AnnData object or a path to AnnData-like"
-            )
 
         if "symbol" in str(var_index):
             logger.warning(
@@ -1042,12 +1039,7 @@ class MuDataCatCurator(CatCurator):
         self._organism = organism
         self._var_fields = var_index
         self._columns_field = var_index  # this is for consistency with BaseCatCurator
-        if isinstance(mdata, Artifact):
-            self._artifact = mdata
-            self._dataset = mdata.load()
-        else:
-            self._artifact = None
-            self._dataset = mdata
+        self._set_dataset_and_artifact()
         self._verify_modality(self._var_fields.keys())
         self._obs_fields = self._parse_categoricals(categoricals)
         self._modalities = set(self._var_fields.keys()) | set(self._obs_fields.keys())
--- a/lamindb/curators/__init__.py
+++ b/lamindb/curators/__init__.py
@@ -1790,6 +1790,8 @@ class SpatialDataCatCurator(CatCurator):
         if isinstance(sdata, Artifact):
             # TODO: load() doesn't yet work
             self._sdata = sdata.load()
+        else:
+            self._sdata = self._dataset
         self._sample_metadata_key = sample_metadata_key
         self._var_fields = var_index
         self._verify_accessor_exists(self._var_fields.keys())
--- a/lamindb/curators/__init__.py
+++ b/lamindb/curators/__init__.py
@@ -482,7 +482,7 @@ class CatCurator(Curator):
         self._validate_category_error_messages: str = ""
 
     @property
-    def non_validated(self) -> dict[str, list[str]]:
+    def non_validated(self) -> dict[str, list[str]] | dict[str, dict[str, list[str]]]:
         """Return the non-validated features and labels."""
         if self._non_validated is None:
             raise ValidationError("Please run validate() first!")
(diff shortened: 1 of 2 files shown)


lamindb/curators/__init__.py
```diff
@@ -173,8 +173,15 @@ class Curator:
...
 
     def __init__(self, dataset: Any, schema: Schema | None = None):
-        self._dataset: Any = dataset  # pass the dataset as a UPathStr or data object
         self._artifact: Artifact = None  # pass the dataset as a non-curated artifact
+        self._dataset: Any = dataset  # pass the dataset as a UPathStr or data object
+        if isinstance(dataset, Artifact):
+            self._artifact = dataset
+            if (
+                self._artifact.otype != "tiledbsoma"
+                and self._artifact.suffix != ".tiledbsoma"
```
sunnyosun, line 182: @Koncopd is this the best way to check if an artifact is a soma store?

lamindb/curators/__init__.py
```diff
@@ -173,8 +173,16 @@ class Curator:
...
-        self._dataset: Any = dataset  # pass the dataset as a UPathStr or data object
         self._artifact: Artifact = None  # pass the dataset as a non-curated artifact
+        self._dataset: Any = dataset  # pass the dataset as a UPathStr or data object
+        if isinstance(dataset, Artifact):
+            self._artifact = dataset
+            # tiledbsoma logic is dealt in TiledbsomaCatCurator
+            if (
+                self._artifact.otype != "tiledbsoma"
+                or self._artifact.suffix != ".tiledbsoma"
+            ):
```
Koncopd, line 184: Maybe this should actually check that this is `anndata` or `dataframe` in addition to checking that it is `tiledbsoma`?

sunnyosun: I agree, but I think that's even harder, isn't it? For example, how to check if an artifact is anndata? `otype="anndata"` or `suffix=".h5ad"` or `suffix=".h5"? dataframe also has different suffixes.

falexwolf: I think that this `otype` checking should _solely_ be done based on the `otype`, not taking into account `.suffix`.

Assuring integrity of the artifact state is another concern that we can debate maintaining in another function, but not with custom "`or`" logic that's only going to bite us with unexpected behavior and hard-to-maintain code.

lamindb/curators/__init__.py
```diff
@@ -1034,10 +1039,15 @@ def __init__(
         if exclude is None:
             exclude = {}
         self._exclude = exclude
-        self._dataset = mdata
         self._organism = organism
         self._var_fields = var_index
         self._columns_field = var_index  # this is for consistency with BaseCatCurator
+        if isinstance(mdata, Artifact):
+            self._artifact = mdata
+            self._dataset = mdata.load()
```
falexwolf, line 1047: This is a bit dangerous given this thing might be huge and it might lead to an unexpected download, but I don't see another way for now.

falexwolf (changes requested):
Looks great! Only mini changes, please!

tests/curators/test_cat_curators.py
```diff
@@ -384,6 +399,7 @@ def test_mudata_curator(mdata):
     }
 
     try:
+        artifact = None
```
falexwolf, line 402: Why is this needed?

tests/curators/test_cat_curators.py
```diff
@@ -192,6 +192,21 @@ def test_df_curator(df, categoricals):
         ln.Schema.filter().delete()
 
 
+def test_pass_artifact(df):
```
falexwolf, line 195: One could consider parametrizing all tests with passing artifact vs. not passing it but given the minimal layer of logic in every constructor it's presumably overkill.

Another elegant way of sharing the logic across all Curator classes could be a private method

```
def _assign_dataset_and_artifact(data: Any):
            if isinstance(sdata, Artifact):
            self._artifact = sdata
            self._dataset = sdata.load()
        else:
            self._artifact = None
            self._dataset = sdata
```

Then we'd not even need to test it outside of testing this mini function.

falexwolf: Then, all curator child classes could simply call this method, @sunnyosun?

tests/curators/test_cat_curators.py
sunnyosun: So the variable exists.

lamindb/curators/__init__.py
sunnyosun: The danger is old artifacts might not have `otype` properly populated. Also for `.csv` files?

tests/curators/test_cat_curators.py
sunnyosun: sure

lamindb/curators/__init__.py
Koncopd: I agree.

lamindb/curators/__init__.py
```diff
@@ -821,30 +834,21 @@ def __init__(
...
     ) -> None:
-        from lamindb_setup.core import upath
-
         if isinstance(var_index, str):
             raise TypeError("var_index parameter has to be a bionty field")
 
         if sources is None:
             sources = {}
-        if not data_is_anndata(data):
+        if not data_is_anndata(data) and not isinstance(data, Artifact):
```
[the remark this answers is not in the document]
sunnyosun: I'm not sure, because data_is_anndata is used when registering data from path or in-memory object in Artifact init. Passing an Artifact doesn't make sense and will require changing other related logic in Artifact init.

lamindb/curators/__init__.py
falexwolf: An `or` check is unpredictable. If `otype` isn't populated this should throw an error.

The `Curator` is a very complicated thing; you shouldn't be able to feed in a `.csv` file. Only artifacts with `otype` `DataFrame`.

We have to get much more rigorous with enforcing `otype`-integrity across the codebase. So if you keep a `.suffix` check here it'll have to be refactored sooner or later.

falexwolf: (The `.suffix` is a wild-west filesystem thing to a large degree; the `otype` we can control and validate.)

lamindb/curators/__init__.py
```diff
@@ -198,6 +198,12 @@ def save_artifact(
         # constructor signature
         pass
 
+    def _set_dataset_and_artifact(self):
+        if isinstance(self._dataset, Artifact):
+            self._artifact = self._dataset
+            if self._artifact.otype in {"DataFrame", "AnnData"}:
```
falexwolf, line 204: What do we do in case this evaluates to `False`? We need to raise an error, right?

lamindb/curators/__init__.py
```diff
@@ -198,6 +198,12 @@ def save_artifact(
         # constructor signature
         pass
 
+    def _set_dataset_and_artifact(self):
```
falexwolf, line 201: The function is misnamed now because it doesn't accept `dataset` as an argument (as I thought it would).

Given it operates on the `self._dataset` attribute, it should be called `_set_artifact()` or something similar.

Overall, there is less logic here than I thought, will look further down the code.

lamindb/curators/__init__.py
```diff
@@ -1725,7 +1728,7 @@ def save_artifact(
         return artifact.save()
 
 
-class SpatialDataCatCurator:
+class SpatialDataCatCurator(CatCurator):
```
falexwolf, line 1731: That is very nice that it starts to inherit now! 😄

lamindb/curators/__init__.py
```diff
@@ -1777,15 +1780,19 @@ def __init__(
...
+        super().__init__(
+            dataset=sdata,
+            categoricals={},
+            sources=sources,
+            organism=organism,
+            exclude=exclude,
+        )
+        if isinstance(sdata, Artifact):
+            # TODO: load() doesn't yet work
+            self._sdata = sdata.load()
```
falexwolf, line 1792: This is also problematic because this spatial data object might be huge.

lamindb/curators/__init__.py
```diff
@@ -1875,7 +1882,7 @@ def categoricals(self) -> dict[str, dict[str, FieldAttr]]:
         return self._categoricals
 
     @property
-    def non_validated(self) -> dict[str, dict[str, list[str]]]:
+    def non_validated(self) -> dict[str, dict[str, list[str]]]:  # type: ignore
```
falexwolf, line 1885: Rather than adding type ignore, we should think about defining the schema for `non_validated`. It's a bit of a complicated, vague thing right now.

falexwolf (approved):
Good! I'm OK merging now; but it'd still be good to address the small questions I have.

falexwolf:
What if the dataset gets mutated? Then `save_artifact()` has to update it, right?

lamindb/curators/__init__.py
sunnyosun: I know, but there isn't really another way get the in-memory object. If we allow user to pass both the object and artifact, that will be too dangerous because they might not match.

lamindb/curators/__init__.py
sunnyosun: No, we can't here because this will be dealt with in TiledbCurator, SpatialDataCurator etc.

falexwolf (approved):

PR #2448, opened 2025-02-12
🐛 Fix dangling folders on upload failures

Koncopd:


falexwolf:
@Koncopd, can you please add a detailed PR description? ☺️

Issue #2456, opened 2025-02-14
Refactor `SpatialDataCatCurator.save_artifact()`


falexwolf:
We have an urgent item here related to the writing to a random local file location (e.g. raised by @sophiamaedler):

https://github.com/laminlabs/lamindb/blob/377b75ddaa873c3829e6367be83dd7e2ff9ba57a/lamindb/curators/__init__.py#L2060-L2230

Then there is a broader question of whether it can leverage the `save_artifact()` logic that DataFrame, AnnData & MuData curators leverage. But this could be a later refactor.

falexwolf:
@Zethson, my goal is polish the new-generation DataFrameCurator & AnnDataCurator today and make lamindb 1.1 soon.

Ideally, we'd adopt the spatial data curator, too. I currently can't judge how much work that will be. It could be just a few lines of code.

In any case, the `SpatialDataCatCurator` should be production-ready in 1.1 and the random file writing appears to me a major issue in the code that should be resolved even if more refactoring and adapting to the new pattern happens later.

Zethson:
Yeah sure. Then I'll bump https://github.com/laminlabs/lamindb/issues/2455 in priority today. I think this might warrant a Slack thread. Update to come.

Zethson:
So the fundamental question is:

How to create the final `UPath` for the `Artifact` constructor. Current code:

```python
write_path = (
    f"{settings.cache_dir}/{random.randint(10**7, 10**8 - 1)}.zarr"
)
self._sdata.write(write_path)

# Create the Artifact and associate Artifact metadata
self._artifact = Artifact(
    write_path,
    description=description,
    key=key,
    revises=revises,
    run=run,
)
```

@Koncopd had suggested a solution like this to me back then. "Just write it somewhere in the cache folder."

Is there a better way to do this? I had thought of adapting more code of https://github.com/laminlabs/lamindb/blob/main/lamindb/_artifact.py#L147 for this. If you already know a better solution please let me know. Else, I'll look into this piece of code again and see whether I can extract parts of it do not have this issue.

sunnyosun:
Here's my suggestion: 

```python

self._write_path = (
    f"{settings.cache_dir}/{random.randint(10**7, 10**8 - 1)}.zarr"
)
if self._write_path:
    self._sdata.write(write_path, overwrite=True)

self._artifact = Artifact(
    self._write_path,
    description=description,
    key=key,
    revises=revises,
    run=run,
)
```

[...]

Issue #2455, opened 2025-02-13
Investigate why the SDC might re-ingest data during curation


Zethson:
### Add a description

https://github.com/laminlabs/lamindb/blob/main/lamindb/curators/__init__.py#L2080

This was never a pretty solution but it's not clear why it creates a new one because there should be a hash as a folder.

Issue #2400, opened 2025-01-29
Add `Artifact.from_spatialdata`


Zethson:
### Add a description

Currently does not exist and there's quite some custom code in the `SpatialDataCurator`. Also for consistency

PR #2461, opened 2025-02-14
✨ Introduce `Artifact.from_spatialdata()` and re-factor validation of `spatialdata` stores

Zethson:
Partially fixes

- https://github.com/laminlabs/lamindb/issues/2456

Fixes

- https://github.com/laminlabs/lamindb/issues/2455
- https://github.com/laminlabs/lamindb/issues/2400

Here, we:
- Add `Artifact.from_spatialdata` which is now also used for `SpatialdataCurator.save_artifact()`
- Add more tests for the `Artifact.from_x` tests. This requires adding `mudata` and `spatialdata` to our `unit-core` test matrix. However, given that this is core functionality, I think it is a tradeoff that we should make.
- Make some handling around identifying zarr paths more general
- Adds more fixtures in preparation for `load()` for SpatialData objects


lamindb/core/storage/objects.py
```diff
@@ -1,62 +1,95 @@
...
+SpatialData = TypeVar("SpatialData")
+MuData = TypeVar("MuData")
 
-def _mudata_is_installed():
-    try:
-        import mudata  # noqa: F401c
-    except ImportError:
-        return False
-    return True
+SupportedDataTypes: TypeAlias = AnnData | DataFrame | MuData | SpatialData
```
falexwolf, line 15: Nice! But let's please call these `DataObjectTypes` or `OTypes` or `ObjectTypes` or `SupportedObjectTypes` to match the convention `Artifact.otype`, which holds the string serialization of the class name of the object.

lamindb/curators/__init__.py
```diff
@@ -2171,7 +2159,7 @@ def _add_set_from_spatialdata(
                 """Add Schemas from SpatialData."""
                 if obs_fields is None:
                     obs_fields = {}
-                assert host.otype == "spatialdata"  # noqa: S101
+                assert host.otype == "SpatialData"  # noqa: S101
```
falexwolf, line 2162: If we're changing the convention here we need to run a data migration, too. Otherwise this here will be a mess:

image

Can of course deal with this manually, but that's not nice.

Hmmm. I'm running another data migration anyway this weekend (PR that got merged to main), and could take care of this.

But this means we have to merge this PR and have it be part of lamindb 1.2.

The PR looks great up to the uncovered lines. I assume the only thing that misses are some tests? Is that true?

falexwolf:
I haven't studied in detail whether the "identify store type" is always ran in the `Artifact()` constructor. If it is, there are likely performance considerations.

Given `from_spatialdata()` gets introduced, we don't have to run these checks. If we were going through `Artifact()` then that'd be different unless we pass a schema (see our Slack thread on the whole topic).

falexwolf:
I'll just quickly add the path-based tests and see what this does to coverage.

falexwolf:
Ah, this re-organization only covered two additional lines.

I'm pretty sure that before this last change, I saw 12 uncovered lines. Now it's 10:

image

falexwolf:
The remaining misses are trivial & quick to fix:

- pass a non-mudata to mudata
- pass a non-spatialata to spatialdata
- try using spatialdata when spatialdata is not installed

The one interesting piece is this, which gets us to the question when "check otype" is actually being run.

image

falexwolf:
And yeah, it's consciously not run if using `from_spatialdata()`. Great!

image

This also means we can simply call `Artifact(sdata, ...)` and then it will be run.

falexwolf:
Now it should be good, I think. Let's see what remains!

If things are good, I'll merge this PR to enable refactoring the entire library away from monkey-patching. ☺️

lamindb/core/storage/objects.py
Zethson: Yah, I was actually thinking of adding a `ScverseDataStructures` type to a `types.py` module at the root level of lamindb. This would help us type a bit better across a lot more functions.

lamindb/curators/__init__.py
Zethson: Yup, it's just super minor things. I wanted to get this PR ready Monday morning so let me do that and then we can merge it with 1.2

lamindb/core/storage/objects.py
falexwolf: Sounds great!

falexwolf:
While coverage checks are now green:

image

We still have 6 remaining uncovered lines:

image

falexwolf (approved):
I'm merging this now so that I can proceed with the library refactor!

Thanks, Lukas! It's a good PR!

lamindb/core/storage/objects.py
```diff
@@ -1,62 +1,95 @@
...
+
+        if isinstance(dmem, MuData):
+            return ".h5mu"
+
+    if is_package_installed("spatialdata"):
+        from spatialdata import SpatialData
+
+        if isinstance(dmem, SpatialData):
+            if format is not None:
+                if format not in {"spatialdata.zarr"}:
```
Zethson, line 52: Why is there not `zarr` here? I thought I had it before and it would be consistent with the error message and the behavior for anndata just above.

falexwolf: I need to dig through changes again; will investigate whether this was conscious or not.

Please no changes currently to main however; I spent >14 hours already on the re-structure and am pretty close to merging it.

Then one can polish again without causing conflicts.

PR #2473, opened 2025-02-17
🎨 A `ManyToMany` between `Schema.components` and `.composites`

falexwolf:
This replaces the foreign key `Schema.composite` with the ManyToMany `Schema.components`.

The latter wasn't part of any released functionality, so we can do this in a non-breaking fashion.

The user-facing API remains unchanged:
```python
anndata_schema = ln.Schema(
    name="small_dataset1_anndata_schema",
    otype="AnnData",
    components={"obs": obs_schema, "var": var_schema},
).save()
```


lamindb/_schema.py
```diff
@@ -106,11 +105,11 @@ def __init__(self, *args, **kwargs):
         dtype = None if itype is not None and itype == "Feature" else NUMBER_TYPE
     else:
         dtype = get_type_str(dtype)
-    if composite is not None and composite._state.adding:
-        raise InvalidArgument(f"composite schema {composite} must be saved before use")
     components: dict[str, Schema]
     if components:
         itype = "Composite"
+        if otype is None:
+            raise InvalidArgument("Please pass otype != None for composite schemas")
```
falexwolf, line 112: This is a fluke of codecov because it's actually tested.

Issue #2488, opened 2025-02-19
🚸 `ln.Artifact()` accepts inconsistent suffix in `key`


sunnyosun:
### Add a description

`ln.Artifact(filename, key="filename.suffix").save()` result in `artifact.suffix=None` without raising an error.

PR #2490, opened 2025-02-19
🚸 Raise consistency error if a source path suffix doesn't match the artifact `key` suffix

falexwolf:
We've so far supported the case in which source path and target key in `Artifact()` have different suffixes.

While all functionality worked as expected there are two risks:

1. A user might provide an erroneous suffix either through the passed filepath or the passed key and because the suffix gets persisted as part of the S3 key it can't be easily changed
2. The user might get confused because the storage suffix and the suffix of the virtual path don't agree

Hence we're now throwing this error:
```python
try:
    artifact = ln.Artifact("my_source_file.suffix1", key="my_target_file.suffix2")
except InvalidArgument as error:
    assert str(error).startswith(
        "The suffix '.inconsistent' of the provided key is incorrect, it should be '.md'."
    )
```

Before this PR, the above silently passed.

What remains possible is to call `artifact.replace("my_source_file.suffix2")` if `artifact.suffix == ".suffix1", which will lead to `artifact` having the suffix `.suffix2`.

Resolves:

- https://github.com/laminlabs/lamindb/issues/2488

### DIFF
--- a/docs/registries.ipynb
+++ b/docs/registries.ipynb
@@ -56,7 +56,7 @@
     "\n",
     "# Create non-curated datasets\n",
     "ln.Artifact(datasets.file_jpg_paradisi05(), key=\"images/my_image.jpg\").save()\n",
-    "ln.Artifact(datasets.file_fastq(), key=\"raw/my_fastq.fastq\").save()\n",
+    "ln.Artifact(datasets.file_fastq(), key=\"raw/my_fastq.fastq.gz\").save()\n",
     "ln.Artifact.from_df(datasets.df_iris(), key=\"iris/iris_collection.parquet\").save()\n",
     "\n",
     "# Create a more complex case\n",
--- a/docs/storage/add-replace-cache.ipynb
+++ b/docs/storage/add-replace-cache.ipynb
@@ -89,23 +89,13 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "artifact = ln.Artifact(\"./test-files/iris.csv\", description=\"iris.csv\")"
+    "artifact = ln.Artifact(\"./test-files/iris.csv\", description=\"iris.csv\").save()"
    ]
   },
   {
    "cell_type": "code",
    "execution_count": null,
    "id": "8",
-   "metadata": {},
-   "outputs": [],
-   "source": [
-    "artifact.save()"
-   ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "id": "9",
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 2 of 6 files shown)


falexwolf:
We actually don't necessarily want this hard error behavior.

All functionality works correctly.

I was just confused by @sunnyosun's issue:

- https://github.com/laminlabs/lamindb/issues/2488

I think what we want instead is using the user-provided suffix instead of the original one. 🤔

falexwolf:
The issue is that the user might make a mistake and the one thing that's annoying to fix is a suffix inconsistency because paths on S3 can't be easily changed.

So, we might in fact want the hard error to de-risk user mistakes in the suffixes.

falexwolf:
The last test I'm struggling with is this one:

image

falexwolf:
I don't manage to trigger this `ValueError`: `"ValueError: Can only replace with a local path not in any Storage."`

The case is still raising an error but it's different:

image

falexwolf:
So, @Koncopd, I'm waiting for your review before I merge this.

Koncopd:
>The case is still raising an error but it's different:

Yes, it is because it is replaced by the path in an existing storage with a different suffix, and it raises the suffix error. I assume that now it is not really possible to replace a file by a file with a different suffix, right? If so, then it will raise the correct error when `path_in_storage` is a cloud path with the same suffix.

Koncopd:
Uhh, no, i am not right.

Koncopd:
https://github.com/laminlabs/lamindb/blob/ca681ea77c93bf6f9c60489f5b5c2cd386852ac7/lamindb/_artifact.py#L917
The error is here and should be raised, so something is incorrect either in the code or in the test.

I will check tommorrow.

Koncopd:
Ok, i see, you get a different error because the artifact you are trying to replace has a key that is different from the replacement path.

Koncopd:
Added back triggering `ValueError`.

falexwolf:
> I am not sure, do we want to forbid .replace with different file extensions?

Yes, I think so. For consistency, simplicity of our code, and avoiding user errors (same rationale as at the top of the PR applies).

I think you can probably remove quite a bit of logic if we don't support this, can't you?

Koncopd (approved):

falexwolf:
I'm finally back to this and will now try to restore the previous behavior for `.replace()` as discussed Thursday night.

falexwolf:
FYI, @Koncopd, it was _very_ easy to maintain the existing tests and behavior upon `.replace()`.

I think this is now a much better solution than changing the behavior for `.replace()`.

Issue #2496, opened 2025-02-20
→ fix typos, remove non-existent values, or save terms via .add_new_from("cell_type")


Zethson:
### Add a description

`add_new_from` suggested but it does not exist anymore in this way. It's only accessible through `cat`.

Zethson:
Will be fixed by only enforcing new style curators.

PR #2497, opened 2025-02-20
🐛 Fix `.add_new_from` message

Zethson:
Fixes https://github.com/laminlabs/lamindb/issues/2496

### DIFF
--- a/lamindb/curators/__init__.py
+++ b/lamindb/curators/__init__.py
@@ -3258,7 +3258,7 @@ def validate_categories(
             settings.verbosity = verbosity
 
     # logging messages
-    non_validated_hint_print = hint_print or f'.add_new_from("{key}")'
+    non_validated_hint_print = hint_print or f'.cat.add_new_from("{key}")'
     non_validated = [i for i in non_validated if i not in values_validated]
     n_non_validated = len(non_validated)
     if n_non_validated == 0:
--- a/tests/curators/test_dataframe_curators_accounting_example.py
+++ b/tests/curators/test_dataframe_curators_accounting_example.py
@@ -131,5 +131,5 @@ def test_invalid_label(transactions_schema):
     with pytest.raises(ln.errors.ValidationError) as err:
         curator.validate()
     exconly = """lamindb.errors.ValidationError: 1 term is not validated: 'GBP'
-    → fix typos, remove non-existent values, or save terms via .add_new_from("currency_name")"""
+    → fix typos, remove non-existent values, or save terms via .cat.add_new_from("currency_name")"""
     assert err.exconly() == exconly


lamindb/curators/__init__.py
```diff
@@ -3258,7 +3258,7 @@ def _log_mapping_info():
             settings.verbosity = verbosity
 
     # logging messages
-    non_validated_hint_print = hint_print or f'.add_new_from("{key}")'
+    non_validated_hint_print = hint_print or f'.cat.add_new_from("{key}")'
```
falexwolf, line 3261: Because this is used across the different Curator classes and MuData, Tiledbsoma & SpatialData don't yet exist, I had consciously not touched this: he `.cat` accessor doesn't exist when you run e.g. the `SpatialDataCatManager()` flow.

What seems particularly problematic with this PR is that it's an ad hoc fix that disregards all other design considerations that went into the curator logging. So, some statements will now come with `.cat` prefix and others won't, which will be confusing.

I had hoped that one of us (@sunnyosun, @Zethson or I) does a proper refactor of the logging module where we make decisions on what we want to recommend and then add the decide on whether we want to add the `.cat` prefix or somehow drive users into a direction where they do not ad-hoc update registries.

So, I might be wrong, but I fear this PR does more harm than good. ☺️

sunnyosun: Yes, this is wrong, let's revert it, please!

falexwolf: Given you agree, I'll remove it from the main branch before we have any additional merges on top.

Zethson: I'm surprised that both of you think that it's so wrong but okay.

falexwolf: > I would have thought that the DF and AnnData curators are the most widely used and should therefore not have broken suggestions.

They're not _broken_. ☺️ 

Given `.add_new_from()` doesn't indicate the class and the `Curator` has a very simple API independent of categoricals, it's legit to assume that a suggestion to fix categorical validation concerns the `.cat` manager and not the Curator, isn't it? 
image

Zethson: > What seems particularly problematic with this PR is that it's an ad hoc fix that disregards all other design considerations that went into the curator logging. So, some statements will now come with .cat prefix and others won't, which will be confusing.

Like I think this is currently already problematic because when it is needed it doesn't come either. For our most widely used curators.

falexwolf: Customers e.g. use the `MuDataCurator`.

And the inconsistency that this induces for `AnnData` and `DataFrame` curator because all of a sudden this single statement has a `.cat` prefix but all other comparable statements don't have it will destroy the user mental model that "categoricals are dealt with in the cat manager".

But exactly this is a main design idea and separation of concern (mostly because, as discussed, one requires updating the database whereas the other is mere dataset curation).

Zethson: > Given .add_new_from() doesn't indicate the class and the Curator has a very simple API independent of categoricals, it's legit to assume that a suggestion to fix categorical validation concerns the .cat manager and not the Curator, isn't it?

No, that was not my intuition as someone that has used the old Curators. You might be right that the docs hold some truth here but it violated my expectation.

falexwolf: So, before we don't review the statements in general I'd not rush modifying a single message. I don't even think it'd be sooo much work to review the logging statements, maybe 1 hour.

falexwolf: All this being said: I get why you and others interpret the messages as _broken_ coming from your usage of the old-style flow. But given all the other mental switches people have to do when doing `ln.curators.DataFrameCurator()` instead of `ln.Curator.from_df()` that realization about the new `.cat` manager should be minor. Am I off?

falexwolf: > I didn't realize it

That's why I'm saying _I get why you..._ ☺️

PR #2510, opened 2025-02-26
🩹 Force no compression in `Collection.mapped` even for files with `.gz` extension

Koncopd:
Not nice, but i don't see any other way now.
We don't have any properly (externally) compressed `.gz` files anyways. But this is a mess.

### DIFF
--- a/lamindb/core/_mapped_collection.py
+++ b/lamindb/core/_mapped_collection.py
@@ -27,7 +27,8 @@ if TYPE_CHECKING:
 class _Connect:
     def __init__(self, storage):
         if isinstance(storage, UPath):
-            self.conn, self.store = registry.open("h5py", storage)
+            # force no external compression even for files with .gz extension. REMOVE LATER
+            self.conn, self.store = registry.open("h5py", storage, compression=None)
             self.to_close = True
         else:
             self.conn, self.store = None, storage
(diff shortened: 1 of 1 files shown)


falexwolf (approved):
Maybe we should simply ignore attributing meaning to `.gz`? On the other Claude agrees that many applications do.

https://claude.ai/share/d8d22a9b-d972-468b-a4c2-a4cdb2c8444c

Koncopd:
`fsspec` has the explicit convention of using suffix for the external compression. https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.open
I am also sure virtually everybody uses suffixes to indicate external conversion.

ap--:
@Koncopd this comment is relevant and points to a better solution relying on the magic byte headers. Worth taking a look at intake.

Koncopd:
@ap-- this is in general reasonable, the problem is that we have a very specific edge case in mind here.
Also due to performance reasons we can't just do opening first without gzip, reading 2 bytes, checking and opening with gzip because this operation is performed on every index retrieval in multi-worker mode. Of course we can devise a complicated caching system for this, but as i said this is a very specific edge case.
We are tracking it here https://github.com/ArcInstitute/arc-virtual-cell-atlas/issues/4

PR #2512, opened 2025-02-27
✅ Test fine-grained permissions

Koncopd:
Test basic functionality for fine-grained permissions.

### DIFF
--- a/tests/permissions/conftest.py
+++ b/tests/permissions/conftest.py
@@ -1,0 +1,65 @@
+import json
+import shutil
+from subprocess import DEVNULL, run
+from time import perf_counter
+
+import lamindb_setup as ln_setup
+import psycopg2
+import pytest
+from lamin_utils import logger
+from laminci.db import setup_local_test_postgres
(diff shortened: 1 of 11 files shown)


tests/permissions/conftest.py
```diff
@@ -0,0 +1,37 @@
...
+import lamindb_setup as ln_setup
+import pytest
+from lamin_utils import logger
+
+
+def pytest_sessionstart():
+    t_execute_start = perf_counter()
+
+    ln_setup.settings.auto_connect = True
+    # these are called in separate scrits because can't change connection
```
falexwolf, line 14: Got it, you need to both access as an admin and then as a user to test.

falexwolf (approved):
Great! 🙏

tests/permissions/conftest.py
Koncopd: Yes, exactly.

falexwolf: Typo in "scripts"

Issue #2504, opened 2025-02-24
`.cache()` returns a Path but `.load().path` returns a str


Zethson:
### Add a description

This just caused confusion in the scportrait use case. @Koncopd should both return a Path object?

Koncopd:
What is the format of `.load()` here? If it is just `UPath`, then `UPath.path` is a string indeed.

Zethson:
Never claimed it's a bug. It's just about UX.

Koncopd:
I am not sure i understand then. `.load()` just gives `UPath` in case no loader is registered for this format.

Zethson:
Yeah and that makes sense, but what is potentially off then is that `.path` returns a str and not a `Path` object like all others for which a loader is registered.

It's slightly inconsistent.

Koncopd:
Yeah, but it is a property of `UPath`, it is not our code.

ap--:
`universal-pathlib` did chose `.path` to align with naming in fsspec. (Simplified) It's the path section of a protocol://path fsspec urlpath, or in other words a stripped urlpath.

In my mind the slightly confusing behavior here comes from .load() defaulting to returning a UPath if no loader is specified.

To me it would be more intuitive if load() would raise some error if no loader can be found.

`.open()` would return a (potentially remote) UPath
`.cache()` would return a local UPath

(It might be worth thinking about introducing `open_local()` as an alias for cache?)

falexwolf:
> I agree. @falexwolf @Zethson @sunnyosun , what do you think?

Agreed!

sunnyosun:
Also agreed! `load() can raise a `NotImplementedError` and point user to `.cache()` the file and use their own reader.

PR #2514, opened 2025-02-28
🚸 Raise `NotImplementedError` in `Artifact.load()` if there is no loader

Koncopd:
Fix https://github.com/laminlabs/lamindb/issues/2504

Raise `NotImplementedError` on `artifact.load()` call when no loader is registered for the suffix.


falexwolf (approved):
Great!

PR #2515, opened 2025-02-28
🎨 When re-creating an `Artifact`, link subsequent runs instead of updating `.run` and linking previous runs

falexwolf:
This PR introduces an important change of behavior upon instantiating `Artifact` on the same data, which triggers a hash-based look-up and return of an existing artifact.

Before | After
--- | ---
the `artifact.run` field was updated to the current run and the previous `run` was saved via the `ManyToMany` `artifact._previous_runs`. | the `artifact.run` field was updated to the current run and the previous `run` was saved via the `ManyToMany` `artifact._previous_runs`.

This might seem like an edge case but occurs rather frequently upon re-running the same transforms.

The new behavior is now consistent with all other entities who remain the association with the originally creating run. It also leads to more intuitive lineage graphs: artifacts that were created a long time ago now keep their _primary_ association with the original run, rather than with the latest run.

Note: The same change holds for `Collection` in complete analogy to `Artifact`.


chaichontat:
@falexwolf seems like the before/after text is exactly the same?

falexwolf:
Thanks, @chaichontat! Fixed. Must have been an accidental copy & paste while polishing the message.

image

PR #2516, opened 2025-03-02
🏗️ Re-structure internal API away from monkey-patching Django models

falexwolf:
Objective: Eliminate monkey-patching of Django models.

This PR re-structures the package by integrating all models-related logic into the `models` submodule. Previously, `models.py` only defined abstract base classes for managing the database schema via Django models.

While we lose the separation of concerns between database schema management and downstream logic this no longer seems an important issue: the database schema is essentially converged, and we now want to treat the Django models as regular Python classes with a nice experience in an IDE, and hence, no monkey patching.

What naturally became much harder through this PR is to organize imports of models without inducing circular imports. The hierarchy of models is now reflected in the code organization: models like `Record` or `Storage` or `Run` that need to be defined and imported first are at the top of `models/__init__.py`. Dependent records are defined and imported after.

## Background

In the early design phase of lamindb two design principles were considered very important:

- separation of the database schema and Python layer
- extendability of the database schema through schema modules

So, we separated out the core schema of LaminDB into another package `lnschema_core` early on:

https://docs.lamin.ai/changelog/2022#lnschema-core

Ever since migrating LaminDB from SQLAlchemy/SQLModel to Django, schema management became much simpler. Both because migration management is more robust & automated in Django and because Django comes with a well-established plug-in system that we started to leverage.

The existence of `lnschema_core` lost its practical value and was hence a relict that made the development process more complicated. Hence, it got integrated into lamindb 2 months ago:

- https://github.com/laminlabs/lamindb/pull/2309

## Materials

- https://github.com/laminlabs/lamindb/pull/2129
- https://github.com/laminlabs/pfizer-lamin-usage/issues/75
- https://github.com/laminlabs/lndocs/pull/116


lamindb/models/base.py
```diff
@@ -243,8 +247,8 @@ def inspect(
         field: str | StrField | None = None,
         *,
         mute: bool = False,
-        organism: str | Record | None = None,
-        source: Record | None = None,
+        organism: Union[str, "Record", None] = None,
```
Koncopd, line 250: Hm, why is it needed? I mean changing the typing approach.

falexwolf: I was wondering the same thing.

ruff marked it as an error.

I'll first do the whole refactor and try to get things to work including linting. I'll also put proper types for M2M which we never had.

If one can then still optimize the annotations, I'll do so.

falexwolf: This is what I started to get all of a sudden everywhere

```
lamindb/models/record.py:1337:43: TC010 Invalid string member in `X | Y`-style union type
     |
1337 | def check_key_change(record: "Artifact" | "Transform"):
     |                                           ^^^^^^^^^^^ TC010
1338 |     """Errors if a record's key has falsely changed."""
1339 |     from .artifact import Artifact
```

falexwolf: I realized that these changes seem to be all to due to removing

```
from __future__ import annotations
```

I thought this isn't needed for >3.10 but it clearly changes the behavior.

So, I'm adding it back.

Koncopd: Interesting. So our normal type annotation might be outdated then?

falexwolf: That's what I initially thought but now I don't trust anything anymore so I'll restore things as they were.

Zethson: There's a bit more to it. https://peps.python.org/pep-0563/ was rolled back because of concerns of the pydantic developers. It was put on hold indefinitely https://docs.python.org/3/whatsnew/3.11.html#pep-563-may-not-be-the-future  and there's a new proposal for Python 3.14 https://peps.python.org/pep-0649/ that got accepted. Note that our lower bound is 3.10 so Ruff will always stick to that and not enable features that require later versions.

falexwolf:
I'm currently losing the hyperref links from the relational type annotations 🤔 

Before | After
--- | ---
image | image

falexwolf:
I just noticed now and already merged:

- https://github.com/laminlabs/lndocs/pull/116

falexwolf:
I think the hyperref in the docs is the only problem.

I can't spend more time on this right now, so the links need to be restored in the next iteration, presumably it's an `lndocs` thing.

falexwolf:
Tried to fix via the absence of hyperlinks by importing explicitly in the root level init, but it didn't help.

```
    from .models.artifact import Artifact
    from .models.collection import Collection
    from .models.feature import Feature
    from .models.run import Run, Param
    from .models.project import Person, Project, Reference
    from .models.schema import Schema
    from .models.core import Storage
    from .models.transform import Transform
    from .models.ulabel import ULabel
    from .models.run import User
```

lamindb/models/base.py
falexwolf: I think the key difference thing is (according to Claude): "When you use from __future__ import annotations, type annotations are preserved as strings rather than being evaluated."

So, dropping the future import seems to lead to an "eager evaluation" of the type annotations (I'm not sure whether partial or complete).

I get the difference between the two PEPs in doing lazy/deferred/postponed evaluation. But does this say something about the correct syntax and the need of adding quotes in one case vs. no such need in another?

It clearly "looks prettier" to me without the quotes (and the future import).

falexwolf:
I've created a minimal setup for investigating the interplay of 

```
from __future__ import annotations
```

Type annotations that come from
```
if TYPE_CHECKING:
```

What I'm observing now is that I get the correct for `Space` on the `Record` class:

image

Presumably because they're defined in the same file.

falexwolf:
Even before this PR it turns out that we never had the resolved type other than for the core models.

So, the problem has nothing to do with the refactor in this PR, but with the brittle resolution of actual types that Sphinx does.

I think I've fixed it in `lndocs`.

falexwolf:
But the bigger issue is this behavior:
```python
>>> import lamindb as ln
→ connected lamindb: falexwolf/docsbuild
>>> from typing import get_type_hints
>>> get_type_hints(ln.Run)
  File "/Users/falexwolf/miniconda3/envs/py312/lib/python3.12/typing.py", line 2273, in get_type_hints
    value = _eval_type(value, base_globals, base_locals, base.__type_params__)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/falexwolf/miniconda3/envs/py312/lib/python3.12/typing.py", line 415, in _eval_type
    return t._evaluate(globalns, localns, type_params, recursive_guard=recursive_guard)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/falexwolf/miniconda3/envs/py312/lib/python3.12/typing.py", line 947, in _evaluate
    eval(self.__forward_code__, globalns, localns),
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<string>", line 1, in <module>
NameError: name 'Transform' is not defined. Did you mean: 'transform'?
```

I'll see whether I can resolve this within lamindb in a similar way.

falexwolf:
No, `get_type_hints()` does more than accessing `__annotations__` -- it doesn't care whether this is dynamically updated or not.

falexwolf:
Docs should be fixed:

- https://github.com/laminlabs/lndocs/pull/117

Literal field validation is fixed through injecting the correct literal type for known literal types.

falexwolf:
I had a feeling that the import became slower, but it gotta be a fluke unless these simple checks here are wrong

```
(py312) falexwolf@mbpalex lamindb % git switch restructure
Switched to branch 'restructure'
Your branch is up to date with 'origin/restructure'.
(py312) falexwolf@mbpalex lamindb % time python import_lamindb.py
→ connected lamindb: falexwolf/docsbuild
python import_lamindb.py  1.02s user 0.18s system 98% cpu 1.211 total
(py312) falexwolf@mbpalex lamindb % git switch main
Switched to branch 'main'
Your branch is up to date with 'origin/main'.
(py312) falexwolf@mbpalex lamindb % time python import_lamindb.py
→ connected lamindb: falexwolf/docsbuild
python import_lamindb.py  1.21s user 0.25s system 97% cpu 1.492 total
```

I repeated them and they yielded the same numbers.

The `import_lamindb.py` script only has one line `import lamindb`. Auto-connect is switched on.

lamindb/core/storage/_tiledbsoma.py
```diff
@@ -134,9 +135,10 @@ def save_tiledbsoma_experiment(
     except ImportError as e:
         raise ImportError("Please install tiledbsoma: pip install tiledbsoma") from e
 
-    from lamindb.core._data import get_run
     from lamindb.core.storage.paths import auto_storage_key_from_artifact_uid
-    from lamindb.core.versioning import create_uid
+    from lamindb.models import Artifact
```
Koncopd, line 139: I don't like that we can't really import the basic things directly anymore here, but i guess nothing can be done.

Koncopd:
But i also don't see any difference.

Koncopd (approved):

PR #2521, opened 2025-03-03
✨ Support CELLxGENE schema 5.2.0

Zethson:
Tested with the old `cxg.Curator`.

This will become one big PR also with #2522


sunnyosun:
See https://github.com/laminlabs/lamindb/pull/2556

PR #2529, opened 2025-03-05
✨ Enable `ln.track(project="My project")` to auto-link entities to projects

falexwolf:
image


sunnyosun:
It's also confusing to have `Artifact.references` and `Artifact.reference`, maybe we should somehow rename.

lamindb/core/_context.py
```diff
@@ -256,10 +264,10 @@ def track(
         self,
         transform: str | Transform | None = None,
         *,
+        project: str | None = None,
```
sunnyosun, line 267: why can't pass a Project record?

sunnyosun: Also Project.name doesn't have a unique constraint: https://github.com/laminlabs/lamindb/blob/main/lamindb/models/project.py#L168

falexwolf: > why can't one passes a Project record?

To keep the interface & code simple. If you have to write a query before `ln.track()` it starts to get complicated; the goal is a light-weight UX for this.

sunnyosun (approved):

lamindb/core/_context.py
sunnyosun: What about passing `abbr` if the project name is too long?

falexwolf: I think we can allow this down the road, but even the explanation that this is also possible will read complicated. Most people will only discover `abbr` very far down the road.

PR #2530, opened 2025-03-05
♻️ Introduce slots in Schema, Curator and FeatureManager

sunnyosun:


lamindb/models/_feature_manager.py
```diff
@@ -739,11 +741,11 @@ def get(cls, **expression) -> Record:
 
 
 @property  # type: ignore
-def _schema_by_slot(self):
+def slots(self):
```
falexwolf, line 744: A return type is missing here.

lamindb/models/schema.py
```diff
@@ -626,11 +627,27 @@ def registry(self) -> str:
     def registry(self, value) -> None:
         self.itype = value
 
+    @property
+    def slots(self) -> dict[str, Schema]:
+        """Slots."""
+        if hasattr(self, "_slots"):
+            return self._slots
+        if self.itype == "Composite":
+            self._slots = {
```
falexwolf, line 636: This isn't tested.

noxfile.py
```diff
@@ -118,10 +118,10 @@ def install_ci(session, group):
         )
         run(session, "uv pip install --system vitessce")
     elif group == "curator":
-        extras += "zarr,jupyter"
+        extras += "zarr,jupyter,bionty"
         run(
             session,
-            "uv pip install --system ./sub/bionty ./sub/wetlab",
+            "uv pip install --system ./sub/wetlab",
```
falexwolf, line 124: The `./sub/bionty` thing has to be restored. If it causes an issue, it has to be fixed in bionty.

lamindb/models/schema.py
falexwolf: It should also replace the `_get_component(slot: str)` method. If you refactor this it's also going to be tested.

lamindb/models/schema.py
```diff
@@ -626,11 +627,27 @@ def registry(self) -> str:
...
+                link.slot: link.component
+                for link in self.components.through.filter(composite_id=self.id).all()
```
falexwolf, line 638: This implementation is ultra-slow because it makes one db-request for `link.component`.

It'd be better to make a bulk-request that fetches it at all in one join.

falexwolf (approved):
Please document this in the respective guides and add an example to the doc strings. 🙏 

Then you can screenshot these and add it to the PR description.

This is very important consolidation of the code-base. 

We should also call out in the curate-df guide how `schema`, `artifact.features` and `curator` are all parallel.

sunnyosun:
It'd also be really nice to do this for the mutable `artifact.features`:

```python
artifact.features.slots["obs"] = obs_schema
```

But for `schema.slots` and `curator.slots` we want it to be immutable. And also for artifact it shouldn't be mutable in the public API I think, because this concerns the dataset features that should match the dataset in the artifact and that's a one-time thing.

So, in fact that an immutable design is good.

What I don't like then though is that we still have `artifact.features.add_schema()` and `artifact.features.add_feature_set()` in the public API. Both should be deprecated and be private/internal.

noxfile.py
sunnyosun: I think it's the pinning of lamindb in bionty

falexwolf: But bionty should only lower bound lamindb, not pin it 🤔 -- otherwise there will be issues

falexwolf: This is compatible, everything should pass: https://github.com/laminlabs/bionty/blob/559cc9a301857aade2f9c739ce935f6e24121b5d/pyproject.toml#L19

If it doesn't that might be dangerous.

PR #2532, opened 2025-03-05
✨ Add `.load()` for artifacts holding `SpatialData` objects

### LINKED ISSUE #2513: Implement SpatialData .load()
### Add a description

https://github.com/laminlabs/lamindb/blob/28c19d67d9678780ff72bb22726d53e6504234f7/lamindb/core/loaders.py#L159

will need to be adapted. We cannot assume that a `.zarr`  is an AnnData object

Zethson:
Fixes #2513 

- Adds support for loading `SpatialData` `Artifacts` into memory
- Adds a new `ScverseDataStructures` type
- Adds a new `_compat.py` module which has utility functions to execute code from functions that require modules to be installed and available
- Introduces a new `data_fixtures` module that will eventually consolidate data fixtures across tests as much as reasonable
- Adds more tests for various `.load()` functions

### DIFF
--- a/lamindb/core/_compat.py
+++ b/lamindb/core/_compat.py
@@ -1,5 +1,3 @@
-# pragma: no cover
-
 import importlib.util
 from typing import Any, Callable, TypeVar
 
--- a/lamindb/core/_compat.py
+++ b/lamindb/core/_compat.py
@@ -59,10 +59,9 @@ def apply_class_func(
             result = operation(obj)
             return True, result
         except ImportError:
-            print(
+            raise ImportError(
                 f"Object appears to be {class_name} but '{package_name}' package is not installed. "
                 f"Please install with: pip install {package_name}"
-            )
-            return True, None
+            ) from None
 
     return False, None
--- a/lamindb/core/_compat.py
+++ b/lamindb/core/_compat.py
@@ -1,3 +1,5 @@
+# pragma: no cover
+
 import importlib.util
 from typing import Any, Callable, TypeVar
 
--- a/lamindb/core/_compat.py
+++ b/lamindb/core/_compat.py
@@ -16,9 +16,6 @@ def with_package(package_name: str, operation: Callable[[Any], T]) -> T:
         package_name: Package name (e.g., "mudata")
         operation: Function that takes the imported module and returns a result
 
-    Returns:
-        Result of operation if package is available, None otherwise
-
     Examples:
         # For direct package functions
         result = with_package("mudata", lambda mod: mod.read_zarr(path))
@@ -42,15 +39,11 @@ def with_package_obj(
         obj: The object to operate on
         class_name: Expected class name (e.g., "MuData")
         package_name: Package that provides the class (e.g., "mudata")
-        operation: Function to call with the object if package is available
-
-    Returns:
-        Tuple with boolean indicating if object was handled
-        and the result of the operation (if handled)
+        operation: Function to call with the object if package is available.
 
     Examples:
         # For instance methods
-        handled, _ = apply_class_func(dmem, "MuData", "mudata",
+        handled, res = apply_class_func(dmem, "MuData", "mudata",
                                       lambda obj: obj.write(filepath))
     """
     if obj.__class__.__name__ == class_name:
--- a/lamindb/core/_compat.py
+++ b/lamindb/core/_compat.py
@@ -1,0 +1,68 @@
+import importlib.util
+from typing import Any, Callable, TypeVar
+
+T = TypeVar("T")
+
+
+def is_package_installed(package_name: str) -> bool:
+    spec = importlib.util.find_spec(package_name)
+    return spec is not None
+
+
+def with_package(package_name: str, operation: Callable[[Any], T]) -> T:
+    """Execute an operation that requires a specific package.
+
+    Args:
+        package_name: Package name (e.g., "mudata")
+        operation: Function that takes the imported module and returns a result
+
+    Returns:
+        Result of operation if package is available, None otherwise
+
+    Examples:
+        # For direct package functions
+        result = with_package("mudata", lambda mod: mod.read_zarr(path))
+    """
+    try:
+        module = importlib.import_module(package_name)
+        return operation(module)
+    except ImportError:
+        raise ImportError(
+            f"Package '{package_name}' is required but not installed. "
+            f"Please install with: pip install {package_name}"
+        ) from None
+
+
+def apply_class_func(
+    obj: Any, class_name: str, package_name: str, operation: Callable[[Any], T]
+) -> tuple[bool, T | None]:
+    """Handle operations on objects that require specific packages.
+
+    Args:
+        obj: The object to operate on
+        class_name: Expected class name (e.g., "MuData")
+        package_name: Package that provides the class (e.g., "mudata")
+        operation: Function to call with the object if package is available
+
+    Returns:
+        Tuple with boolean indicating if object was handled
+        and the result of the operation (if handled)
+
+    Examples:
+        # For instance methods
+        handled, _ = apply_class_func(dmem, "MuData", "mudata",
+                                      lambda obj: obj.write(filepath))
+    """
+    if obj.__class__.__name__ == class_name:
+        try:
+            importlib.import_module(package_name)
+            result = operation(obj)
+            return True, result
+        except ImportError:
+            print(
+                f"Object appears to be {class_name} but '{package_name}' package is not installed. "
+                f"Please install with: pip install {package_name}"
+            )
+            return True, None
+
+    return False, None
--- a/lamindb/core/_compat.py
+++ b/lamindb/core/_compat.py
@@ -33,7 +33,7 @@ def with_package(package_name: str, operation: Callable[[Any], T]) -> T:
         ) from None
 
 
-def apply_class_func(
+def with_package_obj(
     obj: Any, class_name: str, package_name: str, operation: Callable[[Any], T]
 ) -> tuple[bool, T | None]:
     """Handle operations on objects that require specific packages.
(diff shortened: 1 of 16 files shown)


lamindb/core/loaders.py
```diff
@@ -172,7 +175,7 @@ def load_rds(path: UPathStr) -> UPathStr:
 """Suffixes with defined artifact loaders."""
 
 
-def load_to_memory(filepath: UPathStr, **kwargs):
+def load_to_memory(filepath: UPathStr, **kwargs) -> None | UPathStr:
```
sunnyosun, line 178: Any?

sunnyosun:
Is it possible to get more lines covered?
Screenshot 2025-03-05 at 21 26 01

sunnyosun:
I'm not so sure about `ScverseDataStructures` as I'd prefer to see them listed.

Zethson:
> Is it possible to get more lines covered? Screenshot 2025-03-05 at 21 26 01

Two of those files are in coverage ignore but it apparently doesn't work. Sigh.

lamindb/core/loaders.py
```diff
@@ -156,7 +158,8 @@ def load_rds(path: UPathStr) -> UPathStr:
     ".parquet": pd.read_parquet,
     ".parquet.gz": pd.read_parquet,  # this doesn't work for externally gzipped files, REMOVE LATER
     ".fcs": load_fcs,
-    ".zarr": load_anndata_zarr,
+    ".zarr": load_zarr,
+    ".spatialdata.zarr": load_zarr,
```
sunnyosun, line 162: Why do we need `.spatialdata.zarr` here since `.zarr` uses the same reader?

sunnyosun (approved):
Good from my side! But I'd suggest either @Koncopd or @falexwolf to also take a look!

lamindb/core/loaders.py
sunnyosun: It never returns None, does it?

Koncopd:
I will take a look tmr.

lamindb/core/_compat.py
```diff
@@ -0,0 +1,60 @@
...
+
+T = TypeVar("T")
+
+
+def is_package_installed(package_name: str) -> bool:
+    spec = importlib.util.find_spec(package_name)
+    return spec is not None
+
+
+def with_package(package_name: str, operation: Callable[[Any], T]) -> T:
```
Zethson, line 12: Since at least one more person will look at this PR, I make a note here. I know that this is a bit ugly but allows us to write some safer code. Else, we also always have to write lots of statements where we check for available packages, the type of whatever we're looking for, and a PLEASE INSTALL XYZ message. It gets old quick.

Happy to revert if ugly or we can solve it differently. Else, I'll use it also in other places.

lamindb/core/loaders.py
Zethson: Hmm, using `.spatialdata.zarr` is more explicit but sure - in this case we can omit it :) Done.

falexwolf:
> Adds support for loading SpatialData Artifacts into memory

Curious: Is that a typical usage pattern? Won't most people use `.open()`? (Supporting it is good either way even if it's only going to be used for small objects.)

> Adds a new ScverseDataStructures type

I like it.

lamindb/core/types.py
```diff
@@ -7,3 +11,23 @@
     StrField,
     TransformType,
 )
+from lamindb.core._compat import is_package_installed
+
+if TYPE_CHECKING:
+    from anndata import AnnData
+    from mudata import MuData
+    from spatialdata import SpatialData
+else:
```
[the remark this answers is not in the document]
falexwolf: I also find the `else` block highly confusing. Never seen it before. Are we sure we don't just want to move it above the `if TYPE_CHECKING`?

lamindb/core/types.py
```diff
@@ -7,3 +11,23 @@
...
+    from anndata import AnnData
+
+    if is_package_installed("mudata"):
```
falexwolf, line 23: Have we always called all these at package import-time? At least at some point I made an effort that some things are only imported if they're needed.

Always checking for installed and then importing spatialdata seems like something that won't scale well if we want to do it with more packages in the future?

lamindb/models/artifact.py
```diff
@@ -364,7 +365,7 @@ def get_relative_path_to_directory(
 
 def get_artifact_kwargs_from_data(
     *,
-    data: Path | UPath | str | pd.DataFrame | AnnData | MuData,
+    data: Path | UPath | str | pd.DataFrame | ScverseDataStructures,
```
falexwolf, line 368: That's really nice!

tests/core/test_load.py
```diff
@@ -69,3 +150,50 @@ def test_no_loader(zip_file):
...
+
+def test_load_image(image_filepath):
+    artifact = ln.Artifact(image_filepath, key=str(image_filepath))
+    result = artifact.load()
+    assert Path(result).name == image_filepath.name
+
+
+def test_load_svg(svg_filepath):
+    artifact = ln.Artifact(svg_filepath, key=str(svg_filepath))
+    result = artifact.load()
```
falexwolf, line 191: Why doesn't this error? Because the loader is configured to display an image and that will only work in ipython?

tests/core/data_fixtures.py
```diff
@@ -0,0 +1,56 @@
...
+import mudata as md
+import numpy as np
+import pandas as pd
+import pytest
+import spatialdata as sd
+from scipy.sparse import csr_matrix
+
+
+@pytest.fixture(scope="module")
+def adata():
```
falexwolf, line 11: Because there is no clean-up needed for simple in-memory objects, we don't need a fixture.

I'd say we should simply use the `core.datasets()` module consistently rather than introducing all these ad hoc datasets.

The good thing then is that we can use the same datasets in our docs.

lamindb/core/storage/_tiledbsoma.py
```diff
@@ -24,10 +24,10 @@
 
 
 def _load_h5ad_zarr(objpath: UPath):
-    from lamindb.core.loaders import load_anndata_zarr, load_h5ad
+    from lamindb.core.loaders import load_h5ad, load_zarr
 
     if objpath.is_dir():
-        adata = load_anndata_zarr(objpath)
+        adata = load_zarr(objpath)
```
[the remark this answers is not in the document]
Koncopd: Also there are no checks for this upstream as it is called on an input in `save_tiledbsoma_experiment`.

tests/core/test_load.py
Zethson: Correct. Else, it just returns a Path which I think is somewhat weird or inconsistent but not the scope if this PR.

```python
def load_svg(path: UPathStr) -> None | UPathStr:
    """Display `.svg` in ipython, otherwise return path."""
    if is_run_from_ipython:
        from IPython.display import SVG, display

        display(SVG(filename=path))
        return None
    else:
        return path
```

lamindb/core/storage/_tiledbsoma.py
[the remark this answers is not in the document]
Zethson: https://github.com/laminlabs/lamindb/pull/2532/commits/27e961ca88b1334183d74ab02e6c977dfb763e45

is this a compromise? Then we retain some DRY but can be explicit if needed. We just don't always know what kind of  `.zarr` we get so I can't always use the appropriate one from the get go. I think this however should be okay.

lamindb/core/storage/_zarr.py
```diff
@@ -48,29 +52,62 @@ def identify_zarr_type(
...
     fs, storepath_str = infer_filesystem(storepath)
     if isinstance(fs, LocalFileSystem):
         # this is faster than through an fsspec mapper for local
         open_obj = storepath_str
     else:
         open_obj = create_mapper(fs, storepath_str, check=True)
-    adata = read_zarr(open_obj)
-    return adata
+
+    actual_type = identify_zarr_type(storepath)
```
Koncopd, line 84: what i don't like here is that this makes a mapper (open_obj) again in addition to open_obj above. Maybe it can just return it?

Zethson: Ahh makes sense. I added an additional method that only create the `open_obj` and I am then reusing that.

tests/core/data_fixtures.py
Zethson: Do we really want to clutter https://docs.lamin.ai/lamindb.core.datasets with even more test datasets? This will only grow...

Zethson: I'll spin this out into an issue where I think more about how we can organize this with our goals:

1. Reuse test datasets in tutorials
2. Reuse fixtures across the many test functions
3. Keep everything sane and readable.

I think this goes beyond the scope of this PR now although I did a tiny tiny step here.

lamindb/core/storage/_zarr.py
```diff
@@ -9,25 +9,44 @@
...
 if TYPE_CHECKING:
     from anndata import AnnData
+    from fsspec import FSMap
     from lamindb_setup.core.types import UPathStr
 
+    from lamindb.core.types import ScverseDataStructures
 
-def identify_zarr_type(
+
+def create_zarr_open_obj(
```
Koncopd, line 33: One last thing, i don't think this is needed as a separate function, you can move the code block inside `load_zarr`.

Zethson:
Addressed all comment except for moving some fixtures to datasets and very targeted tests for nested Exceptions. The PR got big so I'd like to get it in now. I'll follow up!

tests/core/data_fixtures.py
falexwolf: > Do we really want to clutter https://docs.lamin.ai/lamindb.core.datasets with even more test datasets? This will only grow...

Yes, I think we should organize all datasets for unit tests and core docs properly in `core.datasets`. The alternative is that they're implicitly scattered all over the place which is much worse. I'm confident we can eliminate many of the datasets that date back years and aren't used anywhere anymore.

PR #2534, opened 2025-03-06
🩹 Do not overwrite `._state_db` of models when the current instance is passed to `.using`

Koncopd:
Passing the current instance to `.using` causes problems downstream.


Zethson (approved):
Thank you

PR #2537, opened 2025-03-07
🐛 Fix passing a path manually when no path found by `nbproject`

Koncopd:
Manual passing of path to `ln.track` was broken, i.e. `ln.track(path="path/to/nb.ipynb")` raised an error if no path was found by `nbproject` automatically.


sunnyosun (approved):
Thank you, Sergei! Great that you caught it!

PR #2543, opened 2025-03-10
✨ Implement MuDataCurator

sunnyosun:
This PR implements a schema-based `MuDataCurator`.

Example:
```python

# add labels
perturbation = ln.ULabel(name="Perturbation", is_type=True).save()
ln.ULabel(name="Perturbed", type=perturbation).save()
ln.ULabel(name="NT", type=perturbation).save()
replicate = ln.ULabel(name="Replicate", is_type=True).save()
ln.ULabel(name="rep1", type=replicate).save()
ln.ULabel(name="rep2", type=replicate).save()
ln.ULabel(name="rep3", type=replicate).save()

# define obs schema
obs_schema = ln.Schema(
    name="mudata_papalexi21_subset_obs_schema",
    features=[
        ln.Feature(name="perturbation", dtype="cat[ULabel[Perturbation]]").save(),
        ln.Feature(name="replicate", dtype="cat[ULabel[Replicate]]").save(),
    ],
).save()

obs_schema_rna = ln.Schema(
    name="mudata_papalexi21_subset_rna_obs_schema",
    features=[
        ln.Feature(name="nCount_RNA", dtype=int).save(),
        ln.Feature(name="nFeature_RNA", dtype=int).save(),
        ln.Feature(name="percent.mito", dtype=float).save(),
    ],
    coerce_dtype=True,
).save()

obs_schema_hto = ln.Schema(
    name="mudata_papalexi21_subset_hto_obs_schema",
    features=[
        ln.Feature(name="nCount_HTO", dtype=int).save(),
        ln.Feature(name="nFeature_HTO", dtype=int).save(),
        ln.Feature(name="technique", dtype=bt.ExperimentalFactor).save(),
    ],
    coerce_dtype=True,
).save()

var_schema_rna = ln.Schema(
    name="mudata_papalexi21_subset_rna_var_schema",
    itype=bt.Gene.symbol,
    dtype=float,
).save()

# define composite schema
mudata_schema = ln.Schema(
    name="mudata_papalexi21_subset_mudata_schema",
    otype="MuData",
    components={
        "obs": obs_schema, 
        "rna:obs": obs_schema_rna, 
        "hto:obs": obs_schema_hto, 
        "rna:var": var_schema_rna,
    },
).save()

mdata = ln.core.datasets.mudata_papalexi21_subset()
curator = ln.curators.MuDataCurator(mdata, mudata_schema)
artifact = curator.save_artifact(key="mudata_papalexi21_subset.h5mu")
assert artifact.schema == mudata_schema
```

### DIFF
--- a/lamindb/curators/__init__.py
+++ b/lamindb/curators/__init__.py
@@ -63,7 +63,12 @@ from lamindb.models import (
     ULabel,
 )
 from lamindb.models._feature_manager import parse_staged_feature_sets_from_anndata
-from lamindb.models.artifact import add_labels, data_is_anndata, data_is_mudata, data_is_spatialdata
+from lamindb.models.artifact import (
+    add_labels,
+    data_is_anndata,
+    data_is_mudata,
+    data_is_spatialdata,
+)
 from lamindb.models.feature import parse_dtype, parse_dtype_single_cat
 from lamindb.models._from_values import _format_values
 
--- a/lamindb/curators/__init__.py
+++ b/lamindb/curators/__init__.py
@@ -7,6 +7,7 @@
 
    Curator
    DataFrameCurator
+   SlotsCurator
    AnnDataCurator
    MuDataCurator
 
@@ -423,7 +424,7 @@ class DataFrameCurator(Curator):
             self._dataset,
             description=description,
             fields=self._cat_manager.categoricals,
-            columns_field=result["field"],
+            index_field=result["field"],
             key=key,
             artifact=self._artifact,
             revises=revises,
@@ -528,7 +529,7 @@ class AnnDataCurator(SlotsCurator):
             self._dataset,
             description=description,
             fields=self.slots["obs"]._cat_manager.categoricals,
-            columns_field=(
+            index_field=(
                 parse_dtype_single_cat(self.slots["var"]._schema.itype, is_itype=True)[
                     "field"
                 ]
@@ -631,8 +632,8 @@ class MuDataCurator(SlotsCurator):
 
         # in form of {modality: var_field}
         self._var_fields: dict[str, FieldAttr] = {}
-        # in form of {modality: obs_fields}
-        self._obs_fields: dict[str, dict[str, FieldAttr]] = {}
+        # in form of {modality: categoricals}
+        self._categoricals: dict[str, dict[str, FieldAttr]] = {}
         for slot, slot_schema in schema.slots.items():
             # Assign to _slots
             if ":" in slot:
@@ -650,11 +651,11 @@ class MuDataCurator(SlotsCurator):
                 ),
                 slot_schema,
             )
-            # Assign to _var_fields and _obs_fields
+            # Assign to _var_fields and _categoricals
             if modality is not None:
                 # makes sure that all modalities are present
                 self._var_fields[modality] = None
-                self._obs_fields[modality] = {}
+                self._categoricals[modality] = {}
             if modality_slot == "var":
                 var_field = parse_dtype_single_cat(slot_schema.itype, is_itype=True)[
                     "field"
@@ -668,10 +669,10 @@ class MuDataCurator(SlotsCurator):
             else:
                 obs_fields = self._slots[slot]._cat_manager.categoricals
                 if modality is None:
-                    self._obs_fields[slot] = obs_fields
+                    self._categoricals[slot] = obs_fields
                 else:
                     # note that this is NOT nested since the nested key is always "obs"
-                    self._obs_fields[modality] = obs_fields
+                    self._categoricals[modality] = obs_fields
 
         # this is for consistency with BaseCatManager
         self._columns_field = self._var_fields
@@ -690,10 +691,10 @@ class MuDataCurator(SlotsCurator):
             self.validate()
         return save_artifact(  # type: ignore
             self._dataset,
-            description=description,
-            fields=self._obs_fields,
-            columns_field=self._var_fields,
             key=key,
+            description=description,
+            fields=self._categoricals,
+            index_field=self._var_fields,
             artifact=self._artifact,
             revises=revises,
             run=run,
@@ -827,10 +828,10 @@ class CatManager:
             settings.verbosity = "warning"
             self._artifact = save_artifact(  # type: ignore
                 self._dataset,
+                key=key,
                 description=description,
                 fields=self.categoricals,
-                columns_field=self._columns_field,
-                key=key,
+                index_field=self._columns_field,
                 artifact=self._artifact,
                 revises=revises,
                 run=run,
@@ -3490,7 +3491,7 @@ def validate_categories_in_df(
 def save_artifact(
     data: pd.DataFrame | ad.AnnData | MuData,
     fields: dict[str, FieldAttr] | dict[str, dict[str, FieldAttr]],
-    columns_field: FieldAttr | dict[str, FieldAttr] | None = None,
+    index_field: FieldAttr | dict[str, FieldAttr] | None = None,
     description: str | None = None,
     organism: str | None = None,
     key: str | None = None,
@@ -3504,7 +3505,7 @@ def save_artifact(
     Args:
         data: The DataFrame/AnnData/MuData object to save.
         fields: A dictionary mapping obs_column to registry_field.
-        columns_field: The registry field to validate variables index against.
+        index_field: The registry field to validate variables index against.
         description: A description of the artifact.
         organism: The organism name.
         type: The artifact type.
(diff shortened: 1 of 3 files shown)


lamindb/curators/__init__.py
```diff
@@ -519,6 +524,177 @@ def save_artifact(
...
+
+    @doc_args(SAVE_ARTIFACT_DOCSTRING)
+    def save_artifact(
+        self,
+        *,
+        key: str | None = None,
+        description: str | None = None,
+        revises: Artifact | None = None,
+        run: Run | None = None,
+    ):
```
Zethson, line 682: We can add a return type here, right?

lamindb/curators/__init__.py
```diff
@@ -519,6 +524,177 @@ def save_artifact(
...
+    # the example in the docstring is tested in test_curators_quickstart_example
+    """Curator for a MuData object.
+
+    See also :class:`~lamindb.Curator` and :class:`~lamindb.Schema`.
+
+    .. versionadded:: 1.3.0
+
+    Args:
+        dataset: The MuData-like object to validate & annotate.
+        schema: A `Schema` object that defines the validation constraints.
```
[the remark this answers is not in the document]
falexwolf: The Sphinx ref should be in the type annotation? 🤔

lamindb/curators/__init__.py
```diff
@@ -519,6 +524,177 @@ def save_artifact(
...
+
+    Example::
```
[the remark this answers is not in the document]
falexwolf: We need double-colon, @Zethson! Otherwise it doesn't format correctly!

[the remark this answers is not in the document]
falexwolf: I'll make a post on `#eng-devops-progress`. Maybe this wasn't communicated.

lamindb/curators/__init__.py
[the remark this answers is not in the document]
Zethson: Ah yeah that's also good enough.

lamindb/curators/__init__.py
```diff
@@ -519,6 +524,177 @@ def save_artifact(
...
+    """
+
+    def __init__(
+        self,
+        dataset: MuData | Artifact,
+        schema: Schema,
+    ) -> None:
+        super().__init__(dataset=dataset, schema=schema)
+        if not data_is_mudata(self._dataset):
+            raise InvalidArgument("dataset must be MuData-like.")
```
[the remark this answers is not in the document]
falexwolf: That's an interesting one: Can we have a discussion on #eng-devops-disc? If you're right we should eliminate `InvalidArgument` entirely, shouldn't we?

There is some pros to having it and I wasn't aware of a con until reading this.

falexwolf:
Will you follow up with refactoring `organism` in the next PR, @sunnyosun?

```
bt.settings.organism = "human" # organism check will be refactored
```

lamindb/curators/__init__.py
```diff
@@ -519,6 +524,178 @@ def save_artifact(
...
+        if schema.otype != "MuData":
+            raise InvalidArgument("Schema otype must be 'MuData'.")
+
+        # in form of {slot: DataFrameCurator}
+        self._slots: dict[str, DataFrameCurator] = {}
```
falexwolf, line 614: This here could be inherited from `AnnDataCurator`, couldn't it?

lamindb/curators/__init__.py
```diff
@@ -519,6 +524,178 @@ def save_artifact(
...
+        # in form of {modality: var_field}
+        self._var_fields: dict[str, FieldAttr] = {}
```
falexwolf, line 616: What does this mean and why do we need it?

Does it store `categoricals`? If so, can we rename these to `categoricals` for consistency?

lamindb/curators/__init__.py
```diff
@@ -519,6 +524,178 @@ def save_artifact(
...
+        # in form of {modality: obs_fields}
+        self._obs_fields: dict[str, dict[str, FieldAttr]] = {}
```
falexwolf, line 618: Could this be inherited or are we really validating modality-specific `.obs`? This might be a design flaw in `MuData` then.

falexwolf: Ideally we have only a single `.obs` for the whole `MuData`?

sunnyosun:
> Tests for `validate()` are missing right? Most of it is tested upstream but I think we should still have a few calls.
> 

No, `.validate()` is called in `.save_artifact()`, codecov also confirms that it's covered.

sunnyosun:
> I wonder though whether we need more complex test cases to also ensure that we have error behavior that is useful for users. Stuff like specifying a modality in a schema that does not exist or maybe useful logging when someone attaches a schema that only partially curates the objects because modalities were missing in the schema, ....

It errors with `mdata.__getitem__`, I'd not implement anything right now since users familiar with the mudata should be able to understand it. We can improve the UX once more people use it.

lamindb/curators/__init__.py
sunnyosun: Hmm, Maybe it's better that `self._slots: dict[str, AnnDataCurator]`?

lamindb/curators/__init__.py
sunnyosun: We are validating modality-specific obs, I agree we could also get it from `.obs`. But I'm not entirely sure if all modality-specific obs will also show up in global `.obs`. @Zethson Do you know?

Zethson: AFAIK the modality specific obs do NOT show up in the global obs. The global obs are shared among all modalities whereas the modality specific ones are not.

sunnyosun: But in the ln.core.datasets.mudata_papalexi21_subset(), they in fact do with `modality:` as prefix. I'm a bit confused, as I've also worked with other datasets that don't do that.
Screenshot 2025-03-10 at 17 06 39

Zethson: Yes, I updated my comment.

sunnyosun: And in comparison, if I access a modality specific obs, the global obs are not there:
Screenshot 2025-03-10 at 17 07 59

sunnyosun: I think having a single `.obs` for mudata is a bit problematic:
1. modality-specific obs columns might not present
2. even when they are present, the column name has `modality:` as a prefix, which we need to manually strip to match feature names. There will also be duplication issues with `rna:x` and `hto:x` both point to the same feature but present in two columns.

So overall I think it's cleaner to model the mudata schema by modalities.

Another option of the composite schema could be:
```
mudata_schema = ln.Schema(
    name="mudata_papalexi21_subset_mudata_schema",
    otype="MuData",
    components={
        "rna": rna_anndata_schema,
        "hto": hto_anndata_schema,
    },
).save()
```
But the issue is then it's difficult to pass the global `.obs` as a component.

lamindb/curators/__init__.py
sunnyosun: You mean rename `self._obs_fields` to `self._categoricals`?

lamindb/curators/__init__.py
sunnyosun: I'm still not sure, because the keys in `self.slots` should match the keys of the components in schema.

Currently, the components keys are like: `obs`, `rna:obs`, `rna:var`, `hto:obs`. Which is very different from what `slots` keys in `anndata` (which are just obs and var).

If we inherit from `AnnDataCurator` or make `self.slots: dict[str, AnnDataCurator]`, the keys are `rna`, `hto`, which don't match components.

tests/curators/test_curators_quickstart_example.py
```diff
@@ -180,3 +252,35 @@ def test_anndata_curator_no_var(small_dataset1_schema: ln.Schema):
...
+    curator = ln.curators.MuDataCurator(mdata, mudata_schema)
+    assert curator.slots.keys() == {
+        "obs",
+        "rna:obs",
+        "hto:obs",
+        "rna:var",
+    }
+    artifact = curator.save_artifact(key="mudata_papalexi21_subset.h5mu")
+    assert artifact.schema == mudata_schema
+    assert artifact.features.slots.keys() == {
```
falexwolf, line 279: Where do we test the actual content of the feature sets?

falexwolf:
Overall this is great because it uses very few lines of code and delegates everything to `DataFrameCurator`! :)

tests/curators/test_curators_quickstart_example.py
sunnyosun: Adding values uses he same code used in DataFrameCurator and AnnDataCurator

falexwolf: Ok! ☺️

PR #2546, opened 2025-03-10
✨ Add `QuerySet.artifacts_open()` and `QuerySet.artifacts_mapped()`

Koncopd:


Koncopd:
>Maybe raising an error if the QuerySet is not artifacts?

Sure.

>Why not just QuerySet.open() and QuerySet.mapped()? I think that's more in parallel with the current APIs. 

Hm, i wanted to emphasize that it works only for query sets of artifacts.

sunnyosun:
> Also a thing to discuss. Now i raise an error if the model of a query set is not `Artifact`. Maybe i should just query related artifacts instead in this case?

I think the error is good! Querying the related would be too implicit in this case.

lamindb/models/collection.py
```diff
@@ -94,6 +94,62 @@
...
+
+
+def _load_concat_artifacts(
+    artifacts: list[Artifact], join: Literal["inner", "outer"] = "outer", **kwargs
+) -> pd.DataFrame | ad.AnnData:
+    suffixes = {artifact.suffix for artifact in artifacts}
+    # Why is that? - Sergei
+    if len(suffixes) != 1:
+        raise ValueError(
+            "Can only load collections where all artifacts have the same suffix"
```
sunnyosun, line 127: The error needs to also adapt to different functions, right? Now it's not only loading collections, can also be Queryset

falexwolf (changes requested):
I think we need a class `ArtifactQuerySet` if we want to do this.

We cannot bloat the generic `QuerySet` API with methods that won't make sense for almost all query sets, can we?

lamindb/models/query_set.py
```diff
@@ -712,6 +731,100 @@ def latest_version(self) -> QuerySet:
...
+        self,
+        join: Literal["inner", "outer"] = "outer",
+        is_run_input: bool | None = None,
+        **kwargs,
+    ) -> pd.DataFrame | AnnData:
+        """Cache and load a query set of artifacts to memory.
+
+        Returns an in-memory concatenated `DataFrame` or `AnnData` object.
+
+        See also {meth}`~lamindb.models.Collecton.load`.
```
sunnyosun, line 744: update the ref link

lamindb/models/collection.py
```diff
@@ -94,6 +94,62 @@
...
+            f"the artifacts have incompatible file types: {suffixes_str}"
+            if len(suffixes) > 1
+            else f"the file type {suffixes_str} is not supported by pyarrow."
+        )
+        raise ValueError(err_msg)
+    return _open_pyarrow_dataset(paths)
+
+
+def _load_concat_artifacts(
+    artifacts: list[Artifact], join: Literal["inner", "outer"] = "outer", **kwargs
```
sunnyosun, line 121: **kwargs → **concat_kwargs?

falexwolf: Please no `**kwargs` at all!

falexwolf:
Quite generally the name choice `artifacts_load()` etc. is very confusing and I'd not guess what it does.

If we have `ArtifactQuerySet.load()` that would be more understandable.

falexwolf:
Also: we have to find a way of getting rid of monkey-patching Django's QuerySet. I can't think of a simple solution so far but it's pretty annoying that we have this in our codebase (same for QueryManager).

Koncopd:
>I think we need a class ArtifactQuerySet if we want to do this.

Yes, but it is much harder to implement as we have to account for cases where `django` `QuerySet` is returned. Now we use monkey patching for this, but it is impossible to do for something like `ArtifactQuerySet`. I will think on what to do, i have some ideas for how to get rid of monkey patching here.

falexwolf:
> Yes, but it is much harder to implement as we have to account for cases where django QuerySet is returned. Now we use monkey patching for this, but it is impossible to do for something like ArtifactQuerySet.

I know it's much harder. But we have to invest the effort now. No more new hacks. They'll all just bite us. We gotta keep focusin on removing the hacks of the early days to improve quality and performance.

> I will think on what to do, i have some ideas for how to get rid of monkey patching here.

That'd be great!

Koncopd:
To do this properly, i need something like this https://github.com/laminlabs/lamindb/pull/2637 first.

Koncopd:
Continued here https://github.com/laminlabs/lamindb/pull/2743

PR #2549, opened 2025-03-10
🚸 Restrict "column 'col' not in dataframe" error to at most 10 columns

### LINKED ISSUE #2547: Improve columns not in df msg
### Add a description

It's too long. only show first 10 or something like for others

Zethson:
Fixes #2547 

| Before | After |
|--------|-------|
| ValidationError: column 'date' not in dataframe. Columns in dataframe: ['transaction_amount_usd_cent', 'currency_name', 'transaction_id', 'transaction_date', 'transaction_time', 'merchant_id', 'merchant_name', 'merchant_category', 'payment_method', 'card_type', 'card_last_four', 'customer_id', 'customer_email', 'transaction_status', 'is_recurring', 'geography_country', 'geography_city', 'device_type', 'platform', 'is_international', 'exchange_rate', 'fees_applied'] | ValidationError: column 'date' not in dataframe. Found 22 columns including: 'transaction_amount_usd_cent', 'currency_name', 'transaction_id', 'transaction_date', 'transaction_time', 'merchant_id', 'merchant_name', 'merchant_category', 'payment_method', 'card_type'... |

This is in line with the rest of our UX where we also only show the 10 first hits. The difference becomes bigger of course the more columns we have. I observed that a user got spammed with a stupid amount of columns for their dataset which made the whole notebook unreadable.

### DIFF
--- a/lamindb/curators/__init__.py
+++ b/lamindb/curators/__init__.py
@@ -7,7 +7,9 @@
 
    Curator
    DataFrameCurator
+   SlotsCurator
    AnnDataCurator
+   MuDataCurator
 
 CatManager:
 
@@ -19,6 +21,7 @@ CatManager:
    AnnDataCatManager
    MuDataCatManager
    TiledbsomaCatManager
+   SpatialDataCatManager
    CurateLookup
 
 """
@@ -55,15 +58,18 @@ from lamindb.base.types import FieldAttr  # noqa
 from lamindb.core._settings import settings
 from lamindb.models import (
     Artifact,
-    Collection,
     Feature,
     Record,
     Run,
     Schema,
     ULabel,
 )
-from lamindb.models._feature_manager import parse_staged_feature_sets_from_anndata
-from lamindb.models.artifact import add_labels, data_is_anndata, data_is_spatialdata
+from lamindb.models.artifact import (
+    add_labels,
+    data_is_anndata,
+    data_is_mudata,
+    data_is_spatialdata,
+)
 from lamindb.models.feature import parse_dtype, parse_dtype_single_cat
 from lamindb.models._from_values import _format_values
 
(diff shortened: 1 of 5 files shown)


sunnyosun (changes requested):
This looks so complicated and so much code, is there a way you can utilize https://github.com/laminlabs/lamindb/blob/1373325203a1cb4af8eb421c3400fea26405618b/lamindb/models/_from_values.py#L301? Which was used everywhere to truncate logging list.

Zethson:
@sunnyosun thank you. I don't know why I didn't think of it but it makes it much more concise.

lamindb/curators/__init__.py
```diff
@@ -402,8 +402,18 @@ def validate(self) -> None:
                 self._cat_manager_validate()
             except pandera.errors.SchemaError as err:
                 self._is_validated = False
+                err_msg = str(err)
+
+                # Reduce Pandera validation stdout for many columns
+                if "column" in err_msg and "not in dataframe" in err_msg:
+                    missing_col = err_msg.split("column '")[1].split("'")[0]
+                    display_cols_str = _format_values(
+                        list(self._dataset.columns), n=10, quotes=True, sep="'"
```
sunnyosun, line 411: Could this be simplified to `display_cols_str = _format_values(self._dataset.columns, n=10)`?

lamindb/curators/__init__.py
```diff
@@ -402,8 +402,18 @@ def validate(self) -> None:
...
+                    )
+                    err_msg = f"column '{missing_col}' not in dataframe. {len(list(self._dataset.columns))} columns in dataframe including: {display_cols_str}"
```
sunnyosun, line 413:
```
err_msg = f"column '{missing_col}' not in dataframe. {len(list(self._dataset.columns))} columns in dataframe: {display_cols_str}"
```

sunnyosun: what happens if more than 1 columns are not in dataframe?

sunnyosun: This is OK if you need it urgently, but is a bit patchy, we should ideally have a parser for all pandera errors.

sunnyosun:
This is OK if you need it urgently, but it's a bit patchy, we should ideally have a parser for all pandera errors.

Zethson:
No this is not urgent so I might revisit this at a later point with more general code.

Issue #2429, opened 2025-02-05
Add SpatialDataCurator to docs again


Zethson:
### Add a description

The class seems to be gone and the return type. I remember @falexwolf removing parts here temporarily because of docs building issues.

falexwolf:
The docs build start failing because of how the import was handled or some dependency issues and the only way I could get lamindb v1 docs to build was by temporarily removing it

PR #2552, opened 2025-03-12
✨ Add Schema based SpatialDataCurator

Zethson:
fixes

- https://github.com/laminlabs/lamindb/issues/2519
- https://github.com/laminlabs/lamindb/issues/2429

Adds a new Schema based SpatialDataCurator. Example:

```python
import lamindb as ln
import bionty as bt

# define sample schema
sample_schema = ln.Schema(
    name="blobs_sample_level_metadata",
    features=[
        ln.Feature(name="assay", dtype=bt.ExperimentalFactor).save(),
        ln.Feature(name="disease", dtype=bt.Disease).save(),
        ln.Feature(name="development_stage", dtype=bt.DevelopmentalStage).save(),
    ],
    coerce_dtype=True
).save()

# define table obs schema
blobs_obs_schema = ln.Schema(
    name="blobs_obs_level_metadata",
    features=[
        ln.Feature(name="sample_region", dtype="str").save(),
    ],
    coerce_dtype=True
).save()

# define table var schema
blobs_var_schema = ln.Schema(
    name="blobs_var_schema",
    dtype=int
).save()

# define composite schema
spatialdata_schema = ln.Schema(
    name="blobs_spatialdata_schema",
    otype="SpatialData",
    components={
        "sample": sample_schema,
        "table:obs": blobs_obs_schema,
        "table:var": blobs_var_schema,
}).save()

# curate a SpatialData
spatialdata = ln.core.datasets.spatialdata_blobs()
curator = ln.curators.SpatialDataCurator(spatialdata, spatialdata_schema)
try:
    curator.validate()
except ln.errors.ValidationError as error:
    print(error)

# validate again (must pass now) and save artifact
artifact = curator.save_artifact(key="example_datasets/spatialdata1.zarr")
assert artifact.schema == spatialdata_schema
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,6 +60,7 @@ default_storage_unit_core
 default_storage_unit_storage
 test.ipynb
 test2.ipynb
+test3.ipynb
 run-tests
 test-django-validation/
 
--- a/.gitignore
+++ b/.gitignore
@@ -60,7 +60,6 @@ default_storage_unit_core
 default_storage_unit_storage
 test.ipynb
 test2.ipynb
-test3.ipynb
 run-tests
 test-django-validation/
 
--- a/lamindb/curators/__init__.py
+++ b/lamindb/curators/__init__.py
@@ -2090,7 +2090,8 @@ class TiledbsomaCatManager(CatManager):
             When specific :class:`~bionty.Source` instances are pinned and may lack default values (e.g., "unknown" or "na"),
             using the exclude parameter ensures they are not validated.
 
-    Example:
+    Example::
+
         import bionty as bt
         curator = ln.Curator.from_tiledbsoma(
             "./my_array_store.tiledbsoma",
--- a/lamindb/curators/__init__.py
+++ b/lamindb/curators/__init__.py
@@ -259,9 +259,9 @@ class SlotsCurator(Curator):
         self._slots: dict[str, DataFrameCurator] = {}
 
         # used in MuDataCurator and SpatialDataCurator
-        # in form of {table_key: var_field}
+        # in form of {table/modality_key: var_field}
         self._var_fields: dict[str, FieldAttr] = {}
-        # in form of {table_key: categoricals}
+        # in form of {table/modality_key: categoricals}
         self._categoricals: dict[str, dict[str, FieldAttr]] = {}
 
     @property
--- a/lamindb/curators/__init__.py
+++ b/lamindb/curators/__init__.py
@@ -283,7 +283,6 @@ class DataFrameCurator(Curator):
         schema: A `Schema` object that defines the validation constraints.
 
     Example::
-
         import lamindb as ln
         import bionty as bt
 
@@ -454,7 +453,6 @@ class AnnDataCurator(SlotsCurator):
         schema: A `Schema` object that defines the validation constraints.
 
     Example::
-
         import lamindb as ln
         import bionty as bt
 
@@ -597,7 +595,6 @@ class MuDataCurator(SlotsCurator):
         schema: A `Schema` object that defines the validation constraints.
 
     Example::
-
         import lamindb as ln
         import bionty as bt
 
@@ -898,7 +895,6 @@ class CatManager:
     A `CatManager` object makes it easy to validate, standardize & annotate datasets.
 
     Example::
-
         cat_manager = ln.CatManager(
             dataset,
             # define validation criteria as mappings
@@ -1053,7 +1049,7 @@ class DataFrameCatManager(CatManager):
     Returns:
         A curator object.
 
-    Examples::
+    Example::
         import bionty as bt
         curator = ln.Curator.from_df(
             df,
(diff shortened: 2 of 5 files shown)


.gitignore
```diff
@@ -60,6 +60,7 @@ default_storage_unit_core
 default_storage_unit_storage
 test.ipynb
 test2.ipynb
+test3.ipynb
```
sunnyosun, line 63: Are you spamming the gitignore with your test notebooks? :D

lamindb/curators/__init__.py
```diff
@@ -545,6 +553,38 @@ def save_artifact(
         )
 
 
+def _assign_var_fields_categoricals_multimodal(
+    table_key: str | None,
+    table_slot: str,
```
sunnyosun, line 558: I think you are calling them based only spatialdata names, which is a bit confusing to call them table since table is often related to dataframe.

Also since they can be split from `slot` with `:`, why do you need these two params in addition to `slot`?

lamindb/curators/__init__.py
```diff
@@ -703,22 +730,186 @@ def save_artifact(
...
+        super().__init__(dataset=dataset, schema=schema)
+        if not data_is_spatialdata(self._dataset):
+            raise InvalidArgument("dataset must be SpatialData-like.")
+        if schema.otype != "SpatialData":
+            raise InvalidArgument("Schema otype must be 'SpatialData'.")
+
+        # in form of {table_key: var_field}
+        self._var_fields: dict[str, FieldAttr] = {}
+        # in form of {table_key: categoricals}
+        self._categoricals: dict[str, dict[str, FieldAttr]] = {}
```
sunnyosun, line 821: How about move `self._var_fields` and `self._categoricals` initialization into SlotsCurator?

lamindb/curators/__init__.py
```diff
@@ -703,22 +730,186 @@ def save_artifact(
...
+    @doc_args(SLOTS_DOCSTRING)
+    def slots(self) -> dict[str, DataFrameCurator]:
+        """{}"""  # noqa: D415
+        return self._slots
+
+    @doc_args(VALIDATE_DOCSTRING)
+    def validate(self) -> None:
+        """{}"""  # noqa: D415
+        for _, curator in self._slots.items():
+            curator.validate()
```
sunnyosun, line 869: `slots` and `validate` are already in `SlotsCurator`

lamindb/curators/__init__.py
```diff
@@ -703,22 +730,186 @@ def save_artifact(
...
+            self._dataset,
+            key=key,
+            description=description,
+            fields=self._categoricals,
+            index_field=self._var_fields,
+            artifact=self._artifact,
+            revises=revises,
+            run=run,
+            schema=self._schema,
+        )
```
sunnyosun, line 894: This and also the `MudataCurator.save_artifact` can both go into `SlotsCurator` once you move `self._var_fields` and `self._categoricals` initialization as well. This whole thing can be much more compact.

lamindb/curators/__init__.py
```diff
@@ -703,22 +730,186 @@ def save_artifact(
...
+        # curate a SpatialData
+        spatialdata = ln.core.datasets.spatialdata_blobs()
+        curator = ln.curators.SpatialDataCurator(spatialdata, spatialdata_schema)
+        try:
+            curator.validate()
+        except ln.errors.ValidationError as error:
+            print(error)
+
+        # add new categorical values and standardize
+        curator.slots["sample"].cat.add_new_from(key="developmental_stage")
```
sunnyosun, line 796: We want to make sure users don't blindly call `add_new_from`, so I'd remove these calls from the examples and register new records at the beginning, similar to AnnData and MuData examples.

We have an example of `add_new_from` in `curate-df`, so users can still find it.

sunnyosun (changes requested):

lamindb/curators/__init__.py
sunnyosun: Can `sample_region` be of dtype `str` if it's not meant to have controlled vocabulary?

.gitignore
Zethson: Yes, because I have a few POCs that I don't want to delete. I'll remove this one though :)

lamindb/curators/__init__.py
Zethson: > I think you are calling them based only spatialdata names, which is a bit confusing to call them table since table is often related to dataframe.

Renamed to modality and slot_type.

> Also since they can be split from slot with :, why do you need these two params in addition to slot?

We don't need them but it keeps the code shorter in the method. We have to get the modality and slot_type before this already anyways and should not repeat that in the assign function. I played a bit with it but I think the way I did it here is better despite us having to pass a few more parameters.

lamindb/curators/__init__.py
Zethson: The `AnnDataCurator` isn't using these. They are internal variables but I'd probably still keep them only in MuData and SpatialDataCurator. What do you think?

lamindb/curators/__init__.py
Zethson: Moved now and noted that this is MDC and SDC specific.

lamindb/curators/__init__.py
Zethson: 1. Removed the calls.
2. Can't be validated because I came up with this term (by design). Using a name that exists now.
3. Yes, to make the example simpler. Done.

sunnyosun (approved):
Awesome!

PR #2553, opened 2025-03-12
✨ Allow to pass synchronization arguments in `Artifact.cache()`

Koncopd:
Allow to propagate `kwargs` from `Artifact.cache()` to `UPath.synchronize()`. 
In particular, to avoid printing progress:
```python
artifact.cache(print_progress=False)
```


sunnyosun (approved):
Thank you!

PR #2554, opened 2025-03-12
🚸 Refactor `organism` constraints during validation

sunnyosun:
Before | After
--- | ---
Passing `organism` was required in cases in which it didn't make sense. | Passing `organism` is only required in few cases.
Screenshot 2025-03-13 at 11 07 44 | Screenshot 2025-03-13 at 11 06 54

Changes:

- Passing `organism` to `CanCurate.validate()` is now only required if:
  1. registry has a REQUIRED `organism` FK (e.g. won't check nullable [wetlab.Biosample.organism](https://github.com/laminlabs/wetlab/blob/main/wetlab/models.py#L746))
  2. AND `field` used for validating and creating records is NOT unique (won't check for `ontology_id`, `ensembl_gene_id`, `uid`, ...)
- Robust mapping of ensembl prefix to organism via: [curated mapper](https://lamin.ai/laminlabs/bionty-assets/transform/iySVjI9UouaX).
- Removed all `using_key` logic in `CanCurate`, since users can call curate functions from `QuerySet`

### DIFF
--- a/lamindb/models/_from_values.py
+++ b/lamindb/models/_from_values.py
@@ -7,14 +7,13 @@ import pandas as pd
 from django.core.exceptions import FieldDoesNotExist
 from lamin_utils import colors, logger
 
-from .record import Record
-
 if TYPE_CHECKING:
     from collections.abc import Iterable
 
     from lamindb.base.types import FieldAttr, ListLike, StrField
 
     from .query_set import RecordList
+    from .record import Record
 
 
 # The base function for `from_values`
@@ -37,8 +36,9 @@ def get_or_create_records(
     organism_record = _get_organism_record(field, organism, values=iterable)
     # TODO: the create is problematic if field is not a name field
     if create:
+        create_kwargs = {}
         if organism_record:
-            create_kwargs = {"organism": organism_record}
+            create_kwargs["organism"] = organism_record
         return RecordList(
             [
                 registry(**{field.field.name: value}, **create_kwargs)
@@ -58,23 +58,7 @@ def get_or_create_records(
 
     # new records to be created based on new values
     if len(nonexist_values) > 0:
-        source_record = None
-        if _from_source:
-            if isinstance(source, Record):
-                source_record = source
-
-        if source_record:
-            from bionty.core._add_ontology import check_source_in_db
-
-            check_source_in_db(registry=registry, source=source_record)
-
-            _from_source = not source_record.in_db
-        elif hasattr(registry, "source_id"):
-            _from_source = True
-        else:
-            _from_source = False
-
-        if _from_source:
+        if _from_source and hasattr(registry, "source_id"):
             if (
                 organism_record is None
                 and len(records) > 0
@@ -85,7 +69,7 @@ def get_or_create_records(
                 iterable_idx=nonexist_values,
                 field=field,
                 organism=organism_record,
-                source=source_record,
+                source=source,
                 msg=msg,
                 mute=mute,
             )
@@ -98,15 +82,15 @@ def get_or_create_records(
             unmapped_values = nonexist_values
         # unmapped new_ids will NOT create records
         if len(unmapped_values) > 0:
+            # first log the success message
             if len(msg) > 0 and not mute:
                 logger.success(msg)
             s = "" if len(unmapped_values) == 1 else "s"
             print_values = colors.yellow(_format_values(unmapped_values))
-            name = registry.__name__
             n_nonval = colors.yellow(f"{len(unmapped_values)} non-validated")
             if not mute:
                 logger.warning(
-                    f"{colors.red('did not create')} {name} record{s} for "
+                    f"{colors.red('did not create')} {registry.__name__} record{s} for "
                     f"{n_nonval} {colors.italic(f'{field.field.name}{s}')}: {print_values}"  # type: ignore
                 )
     return RecordList(records)
@@ -181,8 +165,6 @@ def get_existing_records(
         msg = ""
 
     # get all existing records in the db
-    # if necessary, create records for the values in kwargs
-    # k:v -> k:v_record
     query = {f"{field.field.name}__in": iterable_idx.values}  # type: ignore
     if organism is not None:
         query["organism"] = organism
--- a/lamindb/models/_from_values.py
+++ b/lamindb/models/_from_values.py
@@ -385,7 +385,7 @@ def _get_organism_record(  # type: ignore
     check = not _is_simple_field_unique(field=field) or organism is not None
 
     if field_str == "ensembl_gene_id" and len(values) > 0 and organism is None:  # type: ignore
-        return _organism_from_ensembl_id(values[0])  # type: ignore
+        return _organism_from_ensembl_id(values[0], registry)  # type: ignore
 
     if _require_organism(registry) and check:
         from bionty._bionty import create_or_get_organism_record
@@ -393,12 +393,10 @@ def _get_organism_record(  # type: ignore
         organism_record = create_or_get_organism_record(
             organism=organism, registry=registry, field=field_str
         )
-        if organism_record and organism_record._state.adding:
-            organism_record.save()
-        return organism_record
+        return organism_record.save()
 
 
-def _organism_from_ensembl_id(id: str) -> Record | None:  # type: ignore
+def _organism_from_ensembl_id(id: str, registry: type[Record]) -> Record | None:  # type: ignore
     import bionty as bt
 
     from .artifact import Artifact  # has to be here to avoid circular imports
@@ -413,4 +411,4 @@ def _organism_from_ensembl_id(id: str) -> Record | None:  # type: ignore
     if prefix in ensembl_prefixes.index:
         sname = ensembl_prefixes.loc[prefix, "scientific_name"]
 
-        return bt.Organism.from_source(scientific_name=sname)
+        return bt.Organism.from_source(scientific_name=sname).save()
--- a/lamindb/models/_from_values.py
+++ b/lamindb/models/_from_values.py
@@ -31,8 +31,7 @@ def get_or_create_records(
     from .query_set import RecordList
 
     registry = field.field.model  # type: ignore
-    if isinstance(field, str):
-        field = registry._meta.get_field(field)
+    field = registry._meta.get_field(field) if isinstance(field, str) else field
     organism_record = _get_organism_record(field, organism, values=iterable)
     # TODO: the create is problematic if field is not a name field
     if create:
@@ -63,7 +62,7 @@ def get_or_create_records(
             if (
                 organism_record is None
                 and len(records) > 0
-                and _require_organism(registry)
+                and _is_organism_required(registry)
             ):
                 organism_record = records[0].organism
             records_bionty, unmapped_values = create_records_from_source(
@@ -103,6 +102,7 @@ def get_existing_records(
     organism: Record | None = None,
     mute: bool = False,
 ) -> tuple[list, pd.Index, str]:
+    """Get existing records from the database."""
     # NOTE: existing records matching is agnostic to the source
     model = field.field.model  # type: ignore
 
@@ -112,7 +112,7 @@ def get_existing_records(
         field=field,
         organism=organism,
         mute=True,
-        public_aware=False,  # standardize only based on the DB reference
+        source_aware=False,  # standardize only based on the DB reference
         return_mapper=True,
     )
     iterable_idx = iterable_idx.to_frame().rename(index=syn_mapper).index
@@ -186,6 +186,7 @@ def create_records_from_source(
     msg: str = "",
     mute: bool = False,
 ) -> tuple[list, pd.Index]:
+    """Create records from source."""
     model = field.field.model  # type: ignore
     records: list = []
     # populate additional fields from bionty
@@ -286,6 +287,7 @@ def create_records_from_source(
 
 
 def index_iterable(iterable: Iterable) -> pd.Index:
+    """Get unique values from an iterable."""
     idx = pd.Index(iterable).unique()
     # No entries are made for NAs, '', None
     # returns an ordered unique not null list
@@ -295,6 +297,7 @@ def index_iterable(iterable: Iterable) -> pd.Index:
 def _format_values(
     names: Iterable, n: int = 20, quotes: bool = True, sep: str = "'"
 ) -> str:
+    """Format values for printing."""
     if isinstance(names, dict):
         items = {
             f"{key}: {value}": None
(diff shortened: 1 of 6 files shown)


lamindb/models/_from_values.py
```diff
@@ -22,56 +22,55 @@ def get_or_create_records(
     field: StrField,
     *,
     create: bool = False,
-    from_source: bool = False,
     organism: Record | str | None = None,
     source: Record | None = None,
     mute: bool = False,
+    _from_source: bool = False,
```
falexwolf, line 28: Why is this private, now?

Could you document all arguments, please? Or provide a comment that explains that all of these parameters are present in `from_values()`?

Could we rename this function to `from_values()`?

lamindb/models/_from_values.py
```diff
@@ -22,56 +22,55 @@ def get_or_create_records(
...
-        else:
-            from_source = False
-
-        if from_source:
+        if _from_source and hasattr(registry, "source_id"):
+            # if can and needed, get organism record from the existing records
+            if (
+                organism_record is None
+                and len(records) > 0
+                and _require_organism(registry)
```
falexwolf, line 66: Can you rename this function to something that makes clear what it does? Like `check_if_organism_is_required()`?

lamindb/models/_from_values.py
```diff
@@ -84,41 +83,36 @@ def get_or_create_records(
...
             if not mute:
                 logger.warning(
-                    f"{colors.red('did not create')} {name} record{s} for "
+                    f"{colors.red('did not create')} {registry.__name__} record{s} for "
                     f"{n_nonval} {colors.italic(f'{field.field.name}{s}')}: {print_values}"  # type: ignore
                 )
     return RecordList(records)
 
 
 def get_existing_records(
```
falexwolf, line 100: Can you add a comment to explain what this function does?

lamindb/models/_from_values.py
```diff
@@ -84,41 +83,36 @@ def get_or_create_records(
...
 
-    # standardize based on the DB reference
     # log synonyms mapped terms
     syn_mapper = model.standardize(
         iterable_idx,
         field=field,
         organism=organism,
         mute=True,
-        public_aware=False,
+        public_aware=False,  # standardize only based on the DB reference
```
falexwolf, line 115: This parameter seems misnamed just like the `.public` property given we use the `source` convention everywhere else. Can you rename it?

falexwolf: Everything should be 100% consistent.

lamindb/models/_from_values.py
```diff
@@ -345,50 +339,84 @@ def _bulk_create_dicts_from_df(
...
+def _is_simple_field_unique(field: FieldAttr) -> bool:
+    """Check if the field is an id field."""
+    # id field is a unique field that's not a relation
+    field = field.field
+    if field.unique and not field.is_relation:
+        return True
+    return False
+
+
 def _get_organism_record(  # type: ignore
```
falexwolf, line 368: Why is this private?

lamindb/models/_from_values.py
```diff
@@ -345,50 +339,84 @@ def _bulk_create_dicts_from_df(
...
+    from .artifact import Artifact  # has to be here to avoid circular imports
 
-def _ensembl_prefix(id: str, field: StrField, organism: Record | None) -> str | None:
-    if field.field.name == "ensembl_gene_id" and organism is None:  # type: ignore
-        if id.startswith("ENSG"):
-            organism = "human"  # type: ignore
-        elif id.startswith("ENSMUSG"):
-            organism = "mouse"  # type: ignore
+    ensembl_prefixes = (
+        Artifact.using("laminlabs/bionty-assets")
```
falexwolf, line 410: How often is this called? Latency won't be great still but of course we should improve latency so much that it can be called in such situtations. ☺️ 

So, I think it's good that you're using it.

lamindb/models/_from_values.py
```diff
@@ -345,50 +339,84 @@ def _bulk_create_dicts_from_df(
...
+        .get(key="ensembl_prefixes.parquet")
```
falexwolf, line 411: How do we guarantee that we'll never change this name: `ensembl_prefixes.parquet`? Is versioning of this file important?

lamindb/models/artifact.py
```diff
@@ -1490,17 +1490,19 @@ def n_objects(self) -> int:
...
-        >>> artifact = ln.Artifact("s3://my-bucket/my-file.csv").save()
-        >>> artifact.path
-        S3QueryPath('s3://my-bucket/my-file.csv')
+            import lamindb as ln
 
-        File in local storage:
+            # File in cloud storage, here AWS S3:
+            artifact = ln.Artifact("s3://my-bucket/my-file.csv").save()
+            artifact.path
+            #S3QueryPath('s3://my-bucket/my-file.csv')
```
falexwolf, line 1500: When you want to indicate outputs, please use the convention from pydantic et al.

```
#> S3QueryPath('s3://my-bucket/my-file.csv') 
```

lamindb/models/artifact.py
```diff
@@ -1490,17 +1490,19 @@ def n_objects(self) -> int:
...
 
-        >>> ln.Artifact("./myfile.csv", key="myfile.csv").save()
-        >>> artifact.path
-        PosixPath('/home/runner/work/lamindb/lamindb/docs/guide/mydata/myfile.csv')
+            # File in local storage:
+            ln.Artifact("./myfile.csv", key="myfile.csv").save()
+            artifact.path
+            #PosixPath('/home/runner/work/lamindb/lamindb/docs/guide/mydata/myfile.csv')
```
falexwolf, line 1505: Output convention.

lamindb/models/artifact.py
```diff
@@ -1547,17 +1549,20 @@ def from_df(
...
-            4        0.050       0.036        0.014       0.002                 0
-            >>> artifact = ln.Artifact.from_df(df, description="Iris flower collection batch1")
-            >>> artifact.save()
+        Example::
+
+            import lamindb as ln
+
+            df = ln.core.datasets.df_iris_in_meter_batch1()
+            df.head()
+            #   sepal_length sepal_width petal_length petal_width iris_organism_code
```
falexwolf, line 1558: output convention

lamindb/models/artifact.py
```diff
@@ -1600,12 +1605,13 @@ def from_anndata(
...
-            >>> adata = ln.core.datasets.anndata_with_obs()
-            >>> artifact = ln.Artifact.from_anndata(adata, description="mini anndata with obs")
-            >>> artifact.save()
+        Example::
+
+            import lamindb as ln
+
+            adata = ln.core.datasets.anndata_with_obs()
+            artifact = ln.Artifact.from_anndata(adata, key="mini_anndata_with_obs.h5ad")
+            artifact.save()
```
falexwolf, line 1614: move `.save()` on the previous line

lamindb/models/artifact.py
```diff
@@ -1662,12 +1668,13 @@ def from_mudata(
...
-            >>> mdata = ln.core.datasets.mudata_papalexi21_subset()
-            >>> artifact = ln.Artifact.from_mudata(mdata, description="a mudata object")
-            >>> artifact.save()
+        Example::
+
+            import lamindb as ln
+
+            mdata = ln.core.datasets.mudata_papalexi21_subset()
+            artifact = ln.Artifact.from_mudata(mdata, key="mudata_papalexi21_subset.h5mu")
+            artifact.save()
```
falexwolf, line 1677: move save() on the previous line

lamindb/models/artifact.py
```diff
@@ -1754,9 +1765,12 @@ def from_tiledbsoma(
...
 
-        Examples:
-            >>> artifact = ln.Artifact.from_tiledbsoma("s3://mybucket/store.tiledbsoma", description="a tiledbsoma store")
-            >>> artifact.save()
+        Example::
+
+            import lamindb as ln
+
+            artifact = ln.Artifact.from_tiledbsoma("s3://mybucket/store.tiledbsoma", description="a tiledbsoma store")
+            artifact.save()
```
falexwolf, line 1773: move save on the previous line

lamindb/models/artifact.py
```diff
@@ -1798,10 +1812,13 @@ def from_dir(
...
-            >>> dir_path = ln.core.datasets.generate_cell_ranger_files("sample_001", ln.settings.storage)
-            >>> artifacts = ln.Artifact.from_dir(dir_path)
-            >>> ln.save(artifacts)
+        Example::
+
+            import lamindb as ln
+
+            dir_path = ln.core.datasets.generate_cell_ranger_files("sample_001", ln.settings.storage)
+            artifacts = ln.Artifact.from_dir(dir_path)
+            ln.save(artifacts)
```
falexwolf, line 1821: We should use `RecordList` here which has `.save()` so that this can go onto a single line, but I fear we're not yet doing this, do we?

lamindb/models/artifact.py
```diff
@@ -2203,12 +2222,11 @@ def cache(self, is_run_input: bool | None = None, **kwargs) -> Path:
...
-        Examples:
-
-            Sync file from cloud and return the local path of the cache:
+        Example::
 
-            >>> artifact.cache()
-            PosixPath('/home/runner/work/Caches/lamindb/lamindb-ci/lndb-storage/pbmc68k.h5ad')
+            # Sync file from cloud and return the local path of the cache
+            artifact.cache()
+            # PosixPath('/home/runner/work/Caches/lamindb/lamindb-ci/lndb-storage/pbmc68k.h5ad')
```
falexwolf, line 2229: output convention

lamindb/models/artifact.py
```diff
@@ -2245,18 +2263,19 @@ def delete(
...
+            artifact.delete() # delete a single file artifact
 
-            >>> artifact = ln.Artifact.filter(key="some.tiledbsoma". is_latest=False).first()
-            >>> artiact.delete() # delete an old version, the data will not be deleted
+            artifact = ln.Artifact.filter(key="some.tiledbsoma". is_latest=False).first()
+            artiact.delete() # delete an old version, the data will not be deleted
 
-            >>> artifact = ln.Artifact.filter(key="some.tiledbsoma". is_latest=True).one()
-            >>> artiact.delete() # delete all versions, the data will be deleted or prompted for deletion.
+            artifact = ln.Artifact.filter(key="some.tiledbsoma". is_latest=True).one()
```
falexwolf, line 2277: use `.get()` instead of `.filter(...).one()`

lamindb/models/artifact.py
```diff
@@ -2350,9 +2369,12 @@ def save(self, upload: bool | None = None, **kwargs) -> Artifact:
...
 
-        Examples:
-            >>> artifact = ln.Artifact("./myfile.csv", description="myfile")
-            >>> artifact.save()
+        Example::
+
+            import lamindb as ln
+
+            artifact = ln.Artifact("./myfile.csv", key="myfile.parquet")
+            artifact.save()
```
falexwolf, line 2377: `save()` on previous line

lamindb/models/can_curate.py
```diff
@@ -577,13 +542,15 @@ def validate(
...
-            array([ True,  True, False, False])
+        Example::
+
+            import bionty as bt
+
+            bt.Gene.from_values(["A1CF", "A1BG", "BRCA2"], field="symbol", organism="human").save()
+
+            gene_symbols = ["A1CF", "A1BG", "FANCD1", "FANCD20"]
+            bt.Gene.validate(gene_symbols, field=bt.Gene.symbol, organism="human")
+            # array([ True,  True, False, False])
```
falexwolf, line 553: output convention

lamindb/models/can_curate.py
```diff
@@ -622,45 +589,39 @@ def from_values(
...
         field_str = get_name_field(cls, field=field)
         return get_or_create_records(
             iterable=values,
             field=getattr(cls, field_str),
             create=create,
-            from_source=from_source,
             organism=organism,
             source=source,
             mute=mute,
+            _from_source=True if cls.__module__.startswith("bionty.") else False,
```
falexwolf, line 615: Are you sure about the `.` here?

lamindb/models/can_curate.py
```diff
@@ -680,15 +641,15 @@ def standardize(
             return_mapper: If `True`, returns `{input_value: standardized_name}`.
             case_sensitive: Whether the mapping is case sensitive.
             mute: Whether to mute logging.
-            public_aware: Whether to standardize from Bionty reference. Defaults to `True` for Bionty registries.
+            public_aware: Whether to standardize from public source. Defaults to `True` for BioRecord registries.
```
falexwolf, line 644: This parameter is misnamed, is should be called `source_aware`.

lamindb/models/can_curate.py
```diff
@@ -776,15 +741,16 @@ def remove_synonym(self, synonym: str | ListLike):
...
+        Example::
+
+            import bionty as bt
+
+            # save "T cell" record
+            record = bt.CellType.from_source(name="T cell").save()
+            assert record.synonyms == "T-cell|T lymphocyte|T-lymphocyte"
+
+            # remove a synonym
+            assert record.remove_synonym("T-cell") == "T lymphocyte|T-lymphocyte"
```
falexwolf, line 753: Does `remove_synonym()` return the value of the updated synonym field? 🤔

falexwolf (changes requested):
Much better than before! I have mostly small stylistic requests. ☺️

lamindb/models/_from_values.py
```diff
@@ -345,50 +339,84 @@ def _bulk_create_dicts_from_df(
...
+        # organism is not required or not a relation
+        if organism_field.null or not organism_field.is_relation:
+            return False
+        else:
+            return True
     except FieldDoesNotExist:
         return False
 
 
+def _is_simple_field_unique(field: FieldAttr) -> bool:
```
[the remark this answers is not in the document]
sunnyosun: it's checking whether a single field is unique, that's exactly what it does

lamindb/models/_from_values.py
sunnyosun: because it's a super low level function that nobody should use

lamindb/models/_from_values.py
sunnyosun: Only if the database has never seen an ensembl id prefix. This won't be called, if ensembl ids with the same prefix are already in the DB.

lamindb/models/_from_values.py
sunnyosun: probably not for now, once we version it, I will modify accordingly.

lamindb/models/artifact.py
sunnyosun: I thought of it, but think this is better, because otherwise if user run into errors, it's hard to understand where it comes from.

lamindb/models/artifact.py
sunnyosun: No, not for `from_dir`

PR #2560, opened 2025-03-14
🐛 Fix organism requirement

sunnyosun:


lamindb/models/_from_values.py
```diff
@@ -405,15 +405,13 @@ def _get_organism_record(  # type: ignore
...
 
-    from .artifact import Artifact  # has to be here to avoid circular imports
-
-    ensembl_prefixes = (
-        Artifact.using("laminlabs/bionty-assets")
-        .get(key="ensembl_prefixes.parquet", is_latest=True)
-        .load(is_run_input=False)
-        .set_index("gene_prefix")
+    localpath = s3_bionty_assets(
+        ".lamindb/0QeqXlKq9aqW8aqe0000.parquet", bt.base.settings.versionsdir
```
Koncopd, line 411: Might this filename change?

Koncopd: I am afraid this is not really safe to pull by a specific filename.

sunnyosun: If a new version is curated, it will be 0001, but there will still be the 0000 file. So this shouldn't break anything, right?

Koncopd: But shouldn't it point to the new version then?

sunnyosun: Yes, then it will be updated here. But this file is very very rarely updated.

PR #2561, opened 2025-03-15
🚸 Skip `ln.track()` when connected in read-only mode

falexwolf:


lamindb/core/_context.py
```diff
@@ -302,11 +302,11 @@ def track(
...
 
         instance_settings = ln_setup.settings.instance
+        # similar logic here: https://github.com/laminlabs/lamindb/pull/2527
+        # TODO: refactor upon new access management
         if (
-            # similar logic here: https://github.com/laminlabs/lamindb/pull/2527
-            # TODO: refactor upon new access management
             instance_settings.dialect == "postgresql"
-            and "_read" in instance_settings.db
+            and "read:" in instance_settings.db
```
Koncopd, line 309: I believe it should be "_read". Please look at the `db_user` table, there are a lot of connections with "_read.something..." pattern not "read:".

PR #2566, opened 2025-03-16
🚸 Add more constructor signatures and specific inherited types

falexwolf:
## Constructor signatures

Before | After
--- | ---
image | image

## `Artifact.get()` signature

Before | After
--- | ---
image | image

## `Registry.get()` signature

Now returns the inherited type for all registries and thereby we have auto-complete right away:

image

Claude [helped](https://claude.ai/share/338cc542-8d19-43c6-935b-df598783cc9e).

## Future work

We'll add similar typing improvements for other functionality in the future. Here is what we could add for `filter()`.

```python
from typing import TypeVar, Type, Optional, Union, Any, Generic
from django.db.models.base import ModelBase

T = TypeVar('T', bound='Record')

# Make QuerySet generic
class QuerySet(Generic[T]):
    # Your QuerySet implementation here
    pass

class Registry(ModelBase):
    def filter(
        cls: Type[T], 
        *queries: Any, 
        **expressions: Any
    ) -> QuerySet[T]:
        """Query records.

        Args:
            queries: One or multiple `Q` objects.
            expressions: Fields and values passed as Django query expressions.

        Returns:
            A :class:`~lamindb.models.QuerySet`.

        See Also:
            - Guide: :doc:`docs:registries`
            - Django documentation: `Queries <https://docs.djangoproject.com/en/stable/topics/db/queries/>`__

        Examples:
            >>> ln.ULabel(name="my label").save()
            >>> ln.ULabel.filter(name__startswith="my").df()
        """
        from .query_set import QuerySet

        return QuerySet[T](model=cls).filter(*queries, **expressions)


class Record(metaclass=Registry):
    """Base record class."""
    pass
```

### DIFF
--- a/tests/permissions/scripts/setup_access.py
+++ b/tests/permissions/scripts/setup_access.py
@@ -2,9 +2,9 @@ import lamindb as ln  # noqa
 import hubmodule.models as hm
 from laminhub_rest.core.db import DbRoleHandler
 
-full_access = ln.models.Space(name="full access", uid="00000001").save()
-select_access = ln.models.Space(name="select access", uid="00000002").save()
-no_access = ln.models.Space(name="no access", uid="00000003").save()
+full_access = ln.models.Space(name="full access", uid="00000001").save()  # type: ignore
+select_access = ln.models.Space(name="select access", uid="00000002").save()  # type: ignore
+no_access = ln.models.Space(name="no access", uid="00000003").save()  # type: ignore
 
 account = hm.Account(id=ln.setup.settings.user._uuid.hex).save()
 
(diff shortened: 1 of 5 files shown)


tests/permissions/scripts/setup_access.py
```diff
@@ -2,9 +2,9 @@
 import hubmodule.models as hm
 from laminhub_rest.core.db import DbRoleHandler
 
-full_access = ln.models.Space(name="full access", uid="00000001").save()
-select_access = ln.models.Space(name="select access", uid="00000002").save()
-no_access = ln.models.Space(name="no access", uid="00000003").save()
+full_access = ln.models.Space(name="full access", uid="00000001").save()  # type: ignore
```
Koncopd, line 5: What is the problem here? Why type error?

falexwolf: Because the user-facing constructor doesn't accept `uid` as an arg.

PR #2569, opened 2025-03-17
🚸 Add `uint` as a valid `dtype`, in `AnnDataCurator`, make `'obs'` schema optional and allow `'uns'` schema

falexwolf:
`uint` one is now a valid `dtype`.

The `AnnDataCurator` now understands validation with composite schemas that contain a slot `uns`.

```python
var_schema = ln.Schema(
    name="scRNA_seq_var_schema",
    dtype="num",
).save()
uns_schema = ln.Schema(
    name="flexible_uns_schema",
    itype=ln.Feature,
).save()
anndata_schema = ln.Schema(
    name="small_dataset1_anndata_schema",
    otype="AnnData",
    components={"var": var_schema, "uns": uns_schema},
).save()
```

Providing an "obs" schema is no longer required.

### DIFF
--- a/docs/faq/key.ipynb
+++ b/docs/faq/key.ipynb
@@ -137,9 +137,7 @@
     "\n",
     "1. {class}`~lamindb.Storage`: Manages the default storage root that can be either local or in the cloud. For more details we refer to {doc}`docs:faq/storage`.\n",
     "2. {class}`~lamindb.Artifact`: Manages datasets with an optional `key` that acts as a relative path within the current default storage root (see {class}`~lamindb.Storage`). An example is a single h5 artifact.\n",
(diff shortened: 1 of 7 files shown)


falexwolf:
@sunnyosun, I want to do a big refactor next that 

(1) clarifies under which conditions things are linked in the annotation step
(2) clarifies whether the data in a feature needs to comply in order to be annotated
(2) speeds up things by not double-validating across the `validate()` and the `annotate()` step

falexwolf:
Ok, there is still a logical flaw here, @sophiamaedler; the content of `.uns` is currently not parsed and used for linking.

We should probably add the parsing; give me a little time to think about it.

falexwolf:
I'm not adding the parsing of labels for "uns" for now because it'd be inconsistent with how things behave for `.obs`.

So you have to continue to do `artifact.features.add_values(adata.uns)` if you want to link the feature values.

I want to enable automated value parsing for categoricals in obs, uns and any other slot in the next PR that will also do some refactoring so that the whole logic runs much faster.

PR #2574, opened 2025-03-18
♻️ Eliminate monkey patching of `django.db.models.QuerySet` and `django.db.models.Manager`

Koncopd:


Koncopd:
Ok, this fails because `lamindb` overloads `QuerySet.get`.

PR #2575, opened 2025-03-18
✨ Add `mute` argument to `Artifact.cache()` and `Artifact.load()` and and clarify kwargs

Koncopd:


lamindb/models/artifact.py
```diff
@@ -2188,7 +2191,9 @@ def load(self, is_run_input: bool | None = None, **kwargs) -> Any:
             filepath, cache_key = filepath_cache_key_from_artifact(
                 self, using_key=settings._using_key
             )
-            cache_path = _synchronize_cleanup_on_error(filepath, cache_key=cache_key)
+            cache_path = _synchronize_cleanup_on_error(
+                filepath, cache_key=cache_key, print_progress=not mute
```
falexwolf, line 2195: 👍

lamindb/models/artifact.py
```diff
@@ -2146,12 +2146,15 @@ def finalize():
...
-    def load(self, is_run_input: bool | None = None, **kwargs) -> Any:
+    def load(
+        self, mute: bool = False, is_run_input: bool | None = None, **kwargs
+    ) -> Any:
         """Cache and load into memory.
 
         See all :mod:`~lamindb.core.loaders`.
 
         Args:
+            mute: Whether to print the progress of caching.
```
falexwolf, line 2157: Also this should be consistent across the API: "Mute logging." Ideally with one docstring variable that's reused across the codebase and then just injected into the docstring via `@doc_args()`. This will also ensure that we consistently use this pattern when it comes up next time.

lamindb/models/artifact.py
```diff
@@ -2225,7 +2232,9 @@ def cache(self, is_run_input: bool | None = None, **kwargs) -> Path:
 
         Args:
             is_run_input: Whether to track this artifact as run input.
-            **kwargs: Keyword arguments for synchronization.
+            mute: Whether to print the progress of caching.
+            **kwargs: Keyword arguments for internal caching functions.
```
falexwolf, line 2236: Please shorten the explanation and move it into a shared docstring variable. You can just say "Internal arguments for debugging." or something like this? That's what you explained why it's needed, right?

lamindb/models/artifact.py
```diff
@@ -2146,12 +2146,15 @@ def finalize():
         _track_run_input(self, is_run_input)
         return access
 
-    def load(self, is_run_input: bool | None = None, **kwargs) -> Any:
+    def load(
+        self, mute: bool = False, is_run_input: bool | None = None, **kwargs
```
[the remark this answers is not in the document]
Koncopd: done

lamindb/models/artifact.py
Koncopd: I would still want to provide more explanations here, like what we are muting or silencing. I have changed to "Silence logging of caching progress".

lamindb/models/artifact.py
Koncopd: done

PR #2583, opened 2025-03-20
📝 Improve contribution guidelines

alexras:
I've added a couple commands and some additional information to `CONTRIBUTING` based on my experience setting up the repo this morning. I'm not sure if all of these are strictly necessary, but figured I'd document them just in case.


falexwolf:
Thanks for this, Alex!

CONTRIBUTING.md
```diff
@@ -15,10 +15,15 @@ For installation from GitHub, call:
 ```bash
 git clone --recursive https://github.com/laminlabs/lamindb
 pip install laminci
+uv venv
```
Zethson, line 18: I would prefer to keep the instructions more agnostic and not require/suggest `uv`. I am of course a fan and user but if we stick to pip and `python -m venv`, I would be happier. Non `uv` workflows still have significant user bases. `uv` users know when to add the `uv` prefix :)

CONTRIBUTING.md
```diff
@@ -15,10 +15,15 @@ For installation from GitHub, call:
...
+source .venv/bin/activate
 nox -s install
+uv pip install -r pyproject.toml "lamindb[dev]"
```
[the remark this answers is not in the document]
Zethson: Well, `nox -s install` is not complete. It does not have all submodules and not the additional dependency groups such as `[dev]`. I think we could also take this as an opportunity to add `doc` and `test` subgroups but that's not that important.

TDLR: I would prefer it if we improved the nox install command to do what we originally planned it to do. The suggested solution here isn't wrong but our workflows are centralized around `nox` and as long as that's the case, I would vote for consistency here.

Zethson (changes requested):
Great, thanks! A few minor comments

falexwolf:
> TDLR: I would prefer it if we improved the nox install command to do what we originally planned it to do. The suggested solution here isn't wrong but our workflows are centralized around nox and as long as that's the case, I would vote for consistency here.

Agreed.

CONTRIBUTING.md
falexwolf: Agreed :)

falexwolf (approved):
Thanks, looks good now!

Zethson (approved):
https://github.com/laminlabs/lamindb/blob/main/noxfile.py#L47

I think we should have `wetlab` here and maybe others but I'll have a look at this myself.

Thank you!

Issue #2584, opened 2025-03-21
Possible to save malformed key for Artifact


paupaiz:
### Add a description

<img width="1365" alt="Image" src="https://github.com/user-attachments/assets/24f325d9-c937-4f1b-8c23-fce350aff4c0" />

sunnyosun:
key should be `schema/raw_adata.h5ad`

falexwolf:
The issue is `key` update validation. Discussing on Slack. Pau is not needed to clarify.

falexwolf:
Closing this in favor of this [Slack thread](https://laminlabs.slack.com/archives/C08CM9GQ622/p1742552519054509). Updates to key should be validated.

falexwolf:
I was wrong in my initial reading of the tests: the validation that @sunnyosun added accounts for all standard cases in which `.key` is changed. I made this more explicit here:

- https://github.com/laminlabs/lamindb/pull/2589

I'll now try a combination strategy of updating `.suffix` and `.key` together. Maybe this is what you did, @paupaiz ☺

falexwolf:
Indeed, there is one combination of parameters in which I can reproduce the corrupted scenario. The following passes because the validation looks only at `.suffix` and the previous value of `.key`:

```python
assert artifact.suffix == ".parquet"
assert artifact.key is None
artifact.key = "my-test-dataset"
artifact.suffix = ""
artifact.save()
```

If `artifact.key` is not `None`, then also this combination strategy is caught.

It'd be great if you could briefly confirm whether it's possible that you made these calls, @paupaiz!

I'll fix this in any case of course but would look more for other potential loopholes in case this seems not plausible.

falexwolf:
Ok, discovered another way. We're not monitoring updates to `key` _right after_ object construction and then it can hit if one doesn't init `key` right away but through this pattern

```python
artifact = ln.Artifact.from_df(df, description="Test")
assert artifact.key is None
artifact.key = "my-test-dataset"
artifact.save()
```

It's very likely that this is what happened. I'll address this.

falexwolf:
Now it should be impossible to arrive at an inconsistent state for `.suffix` / `.key`:

- https://github.com/laminlabs/lamindb/pull/2589

PR #2589, opened 2025-03-21
🐛 More extensive validation for updates to `artifact.key` and `artifact.suffix`

falexwolf:
This PR fixes different loopholes for setting inconsistent values for `artifact.key` and `artifact.suffix` through more validation.

It also removes the outdated `key` FAQ from the docs.

A future PR will refactor the structure of the test so that validation appears in its own section.

## Loophole 1

```python
artifact = ln.Artifact.get(...)  # or any other way of querying it from the database
assert artifact.suffix == ".parquet"
assert artifact.key is None
artifact.key = "my-test-dataset"
artifact.suffix = ""
artifact.save()
```

## Loopehole 2

```python
artifact = ln.Artifact.from_df(df, description="Test")
assert artifact.key is None
artifact.key = "my-test-dataset"
artifact.save()
```

## Loophole 3

```python
artifact = ln.Artifact.from_df(df, description="Test")
assert artifact.key is None
assert artifact.suffix == ".parquet"
artifact.suffix = "whatever"
artifact.save()
```

## Materials

This PR addresses this issue:

- https://github.com/laminlabs/lamindb/issues/2584

This is a follow-up to the PR that introduced validation of `.key` changes:

- https://github.com/laminlabs/lamindb/pull/2326


falexwolf:
@Zethson, I went through the "key" FAQ and found it to be so outdated that I can't think of keeping meaningful content from it.

I think some of the folder generation examples are cool but should likely appear in a use case somehow but I also don't know how right now.

I would delete it here and restore it when time comes in an overhauled form.

Let me know if I'm missing something in the argument!

tests/core/test_artifact.py
```diff
@@ -280,12 +276,42 @@ def test_create_from_dataframe(df):
     assert artifact.n_observations == 2
     assert hasattr(artifact, "_local_filepath")
     artifact.save()
-    # can do backed now, tested in test_storage.py
```
falexwolf, line 283: @Koncopd, I'm not sure if you or @sunnyosun added it -- the downstream streaming functionality shouldn't appear in `test_artifact.py` and given the comment says it's already tested under "storage" it seems save to remove it here. Agreed?

Zethson:
Please delete!

tests/core/test_artifact.py
Koncopd: Why shouldn't it appear? It is artifact functionality. I would prefer to keep it here instead of deleting.

Koncopd: But yes, it looks like this is tested also in `test_storage.py`.

Koncopd: I mean i am fine if you delete this, but i also don't get why not keep it here.

falexwolf:
> Please delete!

Thanks for confirming!

tests/core/test_artifact.py
falexwolf: > But yes, it looks like this is tested also in test_storage.py.

Great!

> I mean i am fine if you delete this

Then I'll delete it and we can restore a version of this after thinking through whether it's actually needed given test_storage.py takes care of testing it

tests/storage/test_storage.py
```diff
@@ -262,12 +262,13 @@ def test_write_read_tiledbsoma(storage):
     artifact_soma = save_tiledbsoma_experiment(
         [test_file],
         description="test tiledbsoma",
-        key="scrna/my-big-dataset",  # can also be None, but that's trivial
+        key="scrna/my-big-dataset.tiledbsoma",  # can also be None, but that's trivial
```
falexwolf, line 265: @Koncopd, I'm very surprised that this inconsistent suffix in the key was possible before. Let me know if you think this should have been the case.

It clashes with all other conventions which is why it was now raised as an error.

Koncopd: I think you are correct, this was a mistake actually.

PR #2592, opened 2025-03-22
🚸 Avoid requiring `coerce_dtype` for `"int"` and `"float"` in case an integer or float `pd.Series.dtype` only deviates by numerical precision/range

falexwolf:
To avoid requiring `coerce_dtype` for `"int"` and `"float"` in case an integer or float `pd.Series.dtype` only deviates by numerical precision/range, we now interpret `"int"` / `"float"` / `"num"` dtypes as unions of their underlying numerical `dtype` families as documented below.

Before | After
--- | ---
image | image

According to the sources and AI I consulted, it is known behavior that pandera doesn't have a built-in way to express an `AnyInt` dtype. pandera interprets `int` as `np.int64`, _not_ as `np.integer`. `np.integer` would provide the correct behavior but doesn't have a counterpart in a validation library or another context where it'd be serialized.

Specifically, I've found this [related comment](https://github.com/unionai-oss/pandera/issues/726#issuecomment-1511391725) on the pandera repo, and had [this](https://claude.ai/share/a37cf745-5256-463f-995e-cd96aa06b7bf) and [this](https://claude.ai/share/3990d0b2-9e72-4d96-9feb-5f437ae4d656) chats with Claude. Citing from the comment:

> The default generic Int type shold allow any of int8, int16, int32 or int64. It is just Int. There are other IntN classes for the specific bid widths. But if I have a Schema with just column: int I'm not too interested in pandera checking for bid width. Perhaps, the rules in the Int class should be relaxed?

This comment did not find a response in the past two years.

The implementation here follows the suggestion in the second Claude chat. I thought about not using pandera for this at all, but then one would need to come up with custom "delayed Check" logic in our codebase.

### DIFF
--- a/docs/curate.ipynb
+++ b/docs/curate.ipynb
@@ -97,7 +97,7 @@
     "        ln.Feature(name=\"assay_oid\", dtype=bt.ExperimentalFactor.ontology_id).save(),\n",
     "        ln.Feature(name=\"donor\", dtype=str, nullable=True).save(),\n",
     "        ln.Feature(name=\"concentration\", dtype=str).save(),\n",
-    "        ln.Feature(name=\"treatment_time_h\", dtype=float, coerce_dtype=True).save(),\n",
+    "        ln.Feature(name=\"treatment_time_h\", dtype=\"num\", coerce_dtype=True).save(),\n",
     "    ],\n",
     ").save()\n",
     "# display the associated features as a dataframe\n",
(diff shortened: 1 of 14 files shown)


falexwolf:
Note to Alex: This could also work via `np.integer`.

falexwolf:
Docs breakage is due to previous PR by @sunnyosun ☺️ -- not this one

PR #2608, opened 2025-03-29
♻️ Order `IsVersioned.versions` query set

falexwolf:
All of the following will now return an ordered query set representing the version history of the record.

```python
artifact.versions
transform.versions
collection.versions
```


Koncopd (approved):
looks good to me

PR #2625, opened 2025-04-02
🚸 Add space management to `ln.track()`

falexwolf:
Assigning spaces to records was possible before, but to conveniently add all entities written during a run to a desired space, you can now pass `space` to `ln.track()`.

```python
ln.track(space="My space")
```

This parallels what we do for labeling entities with projects with `ln.track(project="My project")`.

Note that `bionty` entities are -- for now -- excluded from this because in most use cases they should be saved in the`"All"` space of an instance and be accessible to all instance collaborators.

### DIFF
--- a/lamindb/models/record.py
+++ b/lamindb/models/record.py
@@ -62,7 +62,6 @@ from lamindb_setup.core._hub_core import connect_instance_hub
 from lamindb_setup.core._settings_store import instance_settings_file
 from lamindb_setup.core.upath import extract_suffix_from_path
 
-from lamindb import context as run_context
 from lamindb.base import deprecated
 from lamindb.base.fields import (
     CharField,
@@ -671,6 +670,8 @@ class BasicRecord(models.Model, metaclass=Registry):
         abstract = True
 
     def __init__(self, *args, **kwargs):
+        from lamindb import context as run_context
+
         skip_validation = kwargs.pop("_skip_validation", False)
         if not args:
             if run_context.space is not None:
--- a/lamindb/models/record.py
+++ b/lamindb/models/record.py
@@ -670,12 +670,13 @@ class BasicRecord(models.Model, metaclass=Registry):
         abstract = True
 
     def __init__(self, *args, **kwargs):
-        from lamindb import context as run_context
-
         skip_validation = kwargs.pop("_skip_validation", False)
         if not args:
-            if run_context.space is not None:
-                kwargs["space"] = run_context.space
+            if self.__class__ is Record and not self.__class__.__name__ == "Storage":
+                from lamindb import context as run_context
+
+                if run_context.space is not None:
+                    kwargs["space"] = run_context.space
             if skip_validation:
                 super().__init__(**kwargs)
             else:
@@ -686,9 +687,6 @@ class BasicRecord(models.Model, metaclass=Registry):
                 from .transform import Transform
 
                 validate_fields(self, kwargs)
-
-                if run_context.space is not None:
-                    kwargs["space"] = run_context.space
 
                 # do not search for names if an id is passed; this is important
                 # e.g. when synching ids from the notebook store to lamindb
--- a/lamindb/models/record.py
+++ b/lamindb/models/record.py
@@ -841,16 +841,20 @@ class BasicRecord(models.Model, metaclass=Registry):
                 if k != "run":
                     logger.important(f"{k} records: {', '.join(v)}")
 
-        if self.__class__.__name__ in {
-            "Artifact",
-            "Transform",
-            "Run",
-            "ULabel",
-            "Feature",
-            "Schema",
-            "Collection",
-            "Reference",
-        }:
+        if (
+            self.__class__.__name__
+            in {
+                "Artifact",
+                "Transform",
+                "Run",
+                "ULabel",
+                "Feature",
+                "Schema",
+                "Collection",
+                "Reference",
+            }
+            and self._branch_code >= 1
+        ):
             import lamindb as ln
 
             if ln.context.project is not None:
--- a/lamindb/models/record.py
+++ b/lamindb/models/record.py
@@ -675,6 +675,8 @@ class BasicRecord(models.Model, metaclass=Registry):
             if (
                 issubclass(self.__class__, Record)
                 and not self.__class__.__name__ == "Storage"
+                # do not save bionty entities in restricted spaces by default
+                and self.__class__.__module__ != "bionty.models"
             ):
                 from lamindb import context as run_context
 
--- a/lamindb/models/record.py
+++ b/lamindb/models/record.py
@@ -672,7 +672,10 @@ class BasicRecord(models.Model, metaclass=Registry):
     def __init__(self, *args, **kwargs):
         skip_validation = kwargs.pop("_skip_validation", False)
         if not args:
-            if self.__class__ is Record and not self.__class__.__name__ == "Storage":
+            if (
+                issubclass(self.__class__, Record)
+                and not self.__class__.__name__ == "Storage"
+            ):
                 from lamindb import context as run_context
 
                 if run_context.space is not None:
--- a/lamindb/models/record.py
+++ b/lamindb/models/record.py
@@ -62,6 +62,7 @@ from lamindb_setup.core._hub_core import connect_instance_hub
 from lamindb_setup.core._settings_store import instance_settings_file
 from lamindb_setup.core.upath import extract_suffix_from_path
 
+from lamindb import context as run_context
 from lamindb.base import deprecated
 from lamindb.base.fields import (
     CharField,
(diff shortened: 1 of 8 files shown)


falexwolf:
I'm now getting an access permission error even though `testuser1` is a space collaborator. Did something change in the past hours?

```
tests/permissions/test_permissions.py F                                                                                                                                                                                                       [100%]

===================================================================================================================== FAILURES ======================================================================================================================
__________________________________________________________________________________________________________________ test_lamin_dev ___________________________________________________________________________________________________________________

    def test_lamin_dev():
        script_path = Path(__file__).parent.resolve() / "scripts/example_script.py"
    
        result = subprocess.run(  # noqa: S602
            f"python {script_path}",
            shell=True,
            capture_output=True,
        )
        print(result.stdout.decode())
        print(result.stderr.decode())
>       assert result.returncode == 0
E       assert 1 == 0
E        +  where 1 = CompletedProcess(args='python /Users/falexwolf/repos/laminhub/rest-hub/sub/lamindb/tests/permissions/scripts/example_s...^^^^^\ndjango.db.utils.ProgrammingError: new row violates row-level security policy for table "lamindb_transform"\n\n').returncode

tests/permissions/test_permissions.py:97: AssertionError
--------------------------------------------------------------------------------------------------------------- Captured stdout call ----------------------------------------------------------------------------------------------------------------

  File "/Users/falexwolf/miniconda3/envs/py312/lib/python3.12/site-packages/django/db/backends/utils.py", line 105, in _execute
    return self.cursor.execute(sql, params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
psycopg2.errors.InsufficientPrivilege: new row violates row-level security policy for table "lamindb_transform"


The above exception was the direct cause of the following exception:

  File "/Users/falexwolf/repos/laminhub/rest-hub/sub/lamindb/tests/permissions/scripts/example_script.py", line 11, in <module>
    ln.track(space=space_name)
  File "/Users/falexwolf/repos/laminhub/rest-hub/sub/lamindb/lamindb/core/_context.py", line 358, in track
    self._create_or_load_transform(
  File "/Users/falexwolf/repos/laminhub/rest-hub/sub/lamindb/lamindb/core/_context.py", line 628, in _create_or_load_transform
    ).save()
      ^^^^^^
  File "/Users/falexwolf/repos/laminhub/rest-hub/sub/lamindb/lamindb/models/record.py", line 809, in save
    super().save(*args, **kwargs)
  File "/Users/falexwolf/miniconda3/envs/py312/lib/python3.12/site-packages/django/db/models/base.py", line 892, in save
    self.save_base(
  File "/Users/falexwolf/miniconda3/envs/py312/lib/python3.12/site-packages/django/db/models/base.py", line 998, in save_base
    updated = self._save_table(
              ^^^^^^^^^^^^^^^^^
  File "/Users/falexwolf/miniconda3/envs/py312/lib/python3.12/site-packages/django/db/models/base.py", line 1161, in _save_table
    results = self._do_insert(
              ^^^^^^^^^^^^^^^^
  File "/Users/falexwolf/miniconda3/envs/py312/lib/python3.12/site-packages/django/db/models/base.py", line 1202, in _do_insert
    return manager._insert(
           ^^^^^^^^^^^^^^^^
  File "/Users/falexwolf/miniconda3/envs/py312/lib/python3.12/site-packages/django/db/models/manager.py", line 87, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/falexwolf/miniconda3/envs/py312/lib/python3.12/site-packages/django/db/models/query.py", line 1847, in _insert
    return query.get_compiler(using=using).execute_sql(returning_fields)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/falexwolf/miniconda3/envs/py312/lib/python3.12/site-packages/django/db/models/sql/compiler.py", line 1836, in execute_sql
    cursor.execute(sql, params)
  File "/Users/falexwolf/miniconda3/envs/py312/lib/python3.12/site-packages/django/db/backends/utils.py", line 79, in execute
    return self._execute_with_wrappers(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/falexwolf/miniconda3/envs/py312/lib/python3.12/site-packages/django/db/backends/utils.py", line 92, in _execute_with_wrappers
    return executor(sql, params, many, context)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/falexwolf/miniconda3/envs/py312/lib/python3.12/site-packages/django/db/backends/utils.py", line 100, in _execute
    with self.db.wrap_database_errors:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/falexwolf/miniconda3/envs/py312/lib/python3.12/site-packages/django/db/utils.py", line 91, in __exit__
    raise dj_exc_value.with_traceback(traceback) from exc_value
  File "/Users/falexwolf/miniconda3/envs/py312/lib/python3.12/site-packages/django/db/backends/utils.py", line 105, in _execute
    return self.cursor.execute(sql, params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
django.db.utils.ProgrammingError: new row violates row-level security policy for table "lamindb_transform"
```

Koncopd:
Maybe @fredericenard reinstalled RLS? But i don't see this test passing before, are you sure it is correct actually?

Koncopd:
`testuser1` is actually read collaborator of the space.

Koncopd:
I would also prefer not to have such test in permissions to avoid collisions. We should add more things like that if we can avoid it in my opinion.

falexwolf:
> testuser1 is actually read collaborator of the space, not write. If you want to save things in the space, you need to make testuser1 a write collaborator.

Thank you!

> I would also prefer not to have such a test in permissions to avoid collisions. We should not add more things like that if we can avoid it in my opinion.

This test won't lead to a collision because all operations can be called concurrently by any number of runners without producing an error.

If you have a better idea for creating an end-to-end integration test, let's do something better! But not having any integration test at all seems to risky, doesn't it? This stuff is not allowed to break.

falexwolf:
@Koncopd, can you re-review, please?

tests/permissions/test_permissions.py
```diff
@@ -114,3 +116,16 @@ def test_write_role():
...
+def test_lamin_dev():
+    script_path = Path(__file__).parent.resolve() / "scripts/example_script.py"
+    result = subprocess.run(  # noqa: S602
+        f"python {script_path}",
+        shell=True,
+        capture_output=True,
+    )
+    print(result.stdout.decode())
+    print(result.stderr.decode())
+    assert result.returncode == 0
```
Koncopd, line 131:
```suggestion
    result = subprocess.run(  # noqa: S602
        f"python {script_path}",
        shell=True,
        check=True,
    )
```

Koncopd:
Doing

lamindb/models/record.py
```diff
@@ -671,75 +671,86 @@ class Meta:
...
-                has_consciously_provided_uid = kwargs.pop(
-                    "_has_consciously_provided_uid"
-                )
+        if not args:
             if (
-                isinstance(self, (CanCurate, Collection, Transform))
-                and settings.creation.search_names
-                and not has_consciously_provided_uid
+                issubclass(self.__class__, Record)
+                and not self.__class__.__name__ == "Storage"
```
Koncopd, line 677: Hm, then why does `Storage` have `space` at all?

lamindb/models/record.py
```diff
@@ -671,75 +671,86 @@ class Meta:
...
+                # do not save bionty entities in restricted spaces by default
+                and self.__class__.__module__ != "bionty.models"
```
Koncopd, line 679: This looks pretty bad to me actually. Is it even documented? An even if it is, stacking such exceptions is not a good thing.

Koncopd:
Ok, done.

>This test won't lead to a collision because all operations can be called concurrently by any number of runners without producing an error.

Ok, then.

lamindb/models/record.py
falexwolf: What do you suggest?

>  Is it even documented?

No, we should debate whether we want to do it. But my gut feeling is that this is the right default; we only want to write into a dedicate space in bionty if the user is very explicit about it. Otherwise I expect that many things in bionty end up in a restricted space by accident.

tests/permissions/test_permissions.py
falexwolf: Without these print statements an error is impossible to debug. So, I would much prefer to leave them in the code. Because this test depends on the state of lamin-dev it's not so unlikely that it breaks and then it will save us time if we know why it broke.

Koncopd: `check=True` reports the traceback if an error happens.

lamindb/models/record.py
Koncopd: Yes, also ping me if you do the migration, because then the `Storage` table has to get its own RLS that is pretty special.

falexwolf: This won't happen in the next 6-9 months. It'd happen only in LaminDB 2.0

tests/permissions/test_permissions.py
falexwolf: And the stdout?

falexwolf: But sure that's way more elegant, we wouldn't even need to define the result even though I'm often checking for certain logging messages (not here though)

lamindb/models/record.py
Koncopd: I guess this is fine if we make a note somewhere that `bionty` models ignore this settings. By explicit you mean manually assigning space to an object, right?

tests/permissions/test_permissions.py
Koncopd: >If check is true, and the process exits with a non-zero exit code, a [CalledProcessError](https://docs.python.org/3/library/subprocess.html#subprocess.CalledProcessError) exception will be raised. Attributes of that exception hold the arguments, the exit code, and stdout and stderr if they were captured.

lamindb/models/record.py
Koncopd: Ok, but we just need to keep this mind.

lamindb/models/record.py
falexwolf: Yes. I think users should manually set a space on a bionty object rather than doing this through `ln.track()` for "all objects created in a session".

lamindb/models/record.py
falexwolf: Yes!

tests/permissions/test_permissions.py
falexwolf: Got it!

falexwolf:
What do you still want for approval, @Koncopd? Just the `check=True` thing to re-run CI?

Koncopd:
`check` would be good, yes. But not absolutely necessary.

lamindb/models/record.py
Koncopd: Maybe a remark somewhere about this behavior?

falexwolf: Ok, am adding this to the `ln.track()` documentation and guide.

Koncopd: Thx.

PR #2627, opened 2025-04-02
🐛 Fix SpatialData and MuData check

Zethson:
We were previously loading both mudata and spatialdata if they were installed and users were trying to create ANY Artifact. With this PR we check for types using class names instead of instance checks so that we do no longer need these imports.

To test:

```python
import lamindb as ln
import sys
from types import ModuleType

def get_loaded_packages() -> list[str]:
    modules = []
    for name, module in sys.modules.items():
        if isinstance(module, ModuleType) and not name.startswith('_'):
            if '.' not in name:
                modules.append(name)
    return sorted(modules)

ln.Artifact("LICENSE").save()

loaded_packages = get_loaded_packages()

assert not "mudata" in loaded_packages
assert not "spatialdata" in loaded_packages
```

### DIFF
--- a/lamindb/curators/__init__.py
+++ b/lamindb/curators/__init__.py
@@ -42,6 +42,8 @@ from lamindb_setup.core import deprecated
 from lamindb_setup.core._docs import doc_args
 from lamindb_setup.core.upath import UPath
 
+from lamindb.core._compat import is_package_installed
+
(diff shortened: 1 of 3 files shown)


falexwolf (approved):
Looks great! Thanks!

Issue #2604, opened 2025-03-28
🍱 Migrations for v1.7


falexwolf:
## Changes for v1.7

Mark transfer transforms as `__lamindb_transfer__/` instead of `transfers/` so that we can easily hide them on the UI.

## Changes for v1.6

Squash migrations.

Changes via https://github.com/laminlabs/lamindb/pull/2794

* enable feature name duplication  ✅ 
* hash can be duplicated for different keys for artifact (conditional uniqueness) ✅ 
* hash can be duplicated for transform for different keys
* fix how instances are referred to for `type` ✅ 
* fix default space uid to be "A" ✅ 
* `_expect_many` should default to `None` on Feature so that the auto-display of single values as opposed to a set makes sense, and a user can enforce one (single) or the other (many) in the future  ✅ 
* populate hash for all feature values also for the simple ones so that we have an easy way to universally identify a unique feature value ✅ 

Changes in other PRs.

* `_branch_code` to be integer PK of `Branch` registry ✅ 
*  update `WriteLog` to have sensible FKs into branch, user, run, and space ✅ 
* remove contenttypes ✅ 
* `WriteLog` ✅
* introduce `Record` registry ✅
* integrate `Feature` into `Param` registry ✅

Koncopd:
Please warn me when you do it, i will need to change some it in RLS also.

PR #2635, opened 2025-04-03
✨ Define optional features on a per-schema level & improve schema hash calculation

sunnyosun:
This PR adds `Feature.with_config(optional: bool)` to define schema-specific optional features.

It also updates the hash calculation for the schema hash to account for the schema configuration in all cases. Previously, if features were defined, the hash calculation disregarded the schema configuration.

image

Interacting with optionals:

- Get optional features via `schema.optionals.get()`
- Set optional features via `schema.optionals.set()`
- Add an additional optional feature via `schema.optionals.add()`

Internal [Notion page](https://www.notion.so/laminlabs/Overhaul-the-Schema-design-1ca2aeaa55e1803cb3fed41f955fcc7c).

### DIFF
--- a/lamindb/models/feature.py
+++ b/lamindb/models/feature.py
@@ -454,7 +454,7 @@ class Feature(Record, CanCurate, TracksRun, TracksUpdates):
         self.default_value = default_value
         self.nullable = nullable
         self.coerce_dtype = coerce_dtype
-        self.required: bool = False
+        self.optional: bool | None = None
         dtype_str = kwargs.pop("dtype", None)
         if cat_filters:
             assert "|" not in dtype_str  # noqa: S101
@@ -500,9 +500,9 @@ class Feature(Record, CanCurate, TracksRun, TracksUpdates):
         super().save(*args, **kwargs)
         return self
 
-    def with_config(self, required: bool = False) -> Feature:
+    def with_config(self, optional: bool | None = None) -> Feature:
         """Set required flag."""
-        self.required = required
+        self.optional = optional
         return self
 
     @property
--- a/lamindb/models/feature.py
+++ b/lamindb/models/feature.py
@@ -501,7 +501,9 @@ class Feature(Record, CanCurate, TracksRun, TracksUpdates):
 
     def with_config(self, optional: bool | None = None) -> tuple[Feature, dict]:
         """Pass addtional configurations to the schema."""
-        return self, {"optional": optional}
+        if optional is not None:
+            return self, {"optional": optional}
+        return self, {}
 
     @property
     def default_value(self) -> Any:
--- a/lamindb/models/feature.py
+++ b/lamindb/models/feature.py
@@ -449,11 +449,13 @@ class Feature(Record, CanCurate, TracksRun, TracksUpdates):
         nullable = kwargs.pop("nullable", True)  # default value of nullable
         cat_filters = kwargs.pop("cat_filters", None)
         coerce_dtype = kwargs.pop("coerce_dtype", False)
+        required = kwargs.pop("required", False)
         kwargs = process_init_feature_param(args, kwargs)
         super().__init__(*args, **kwargs)
         self.default_value = default_value
         self.nullable = nullable
         self.coerce_dtype = coerce_dtype
+        self.required = required
         dtype_str = kwargs.pop("dtype", None)
         if cat_filters:
             assert "|" not in dtype_str  # noqa: S101
@@ -500,25 +502,6 @@ class Feature(Record, CanCurate, TracksRun, TracksUpdates):
         return self
 
     @property
-    def coerce_dtype(self) -> bool:
-        """Whether dtypes should be coerced during validation.
-
-        For example, a `objects`-dtyped pandas column can be coerced to `categorical` and would pass validation if this is true.
-        """
-        if self._aux is not None and "af" in self._aux and "2" in self._aux["af"]:  # type: ignore
-            return self._aux["af"]["2"]  # type: ignore
-        else:
-            return False
-
-    @coerce_dtype.setter
-    def coerce_dtype(self, value: bool) -> None:
-        if self._aux is None:  # type: ignore
-            self._aux = {}  # type: ignore
-        if "af" not in self._aux:
-            self._aux["af"] = {}
-        self._aux["af"]["2"] = value
-
-    @property
     def default_value(self) -> Any:
         """A default value that overwrites missing values (default `None`).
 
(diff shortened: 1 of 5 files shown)


falexwolf (changes requested):
Great!

Slowly working through this.

First UX.

I think that `feature.with_config(is_optional=True)` would be even clearer and easier to read. We have the `is_...` convention for many booleans and I'm almost always happy about adding `is_...` rather than not adding it. It will also pre-empt confusion with type annotations. WDYT?

I think that more intuitive and flexible than
```python
schema.optional = [feature1, feature2, feature3]
```
would be
```python
schema.update_is_optional({feature1: True, feature2: False})
```

The mere attribute accessor is not so intuitive because changing the optionality state for some features is a quite sophisticated operation; much more than just changing a single attribute.

sunnyosun:
> Great!
> 
> Slowly working through this.
> 
> First UX.
> 
> I think that `feature.with_config(is_optional=True)` would be even clearer and easier to read. We have the `is_...` convention for many booleans and I'm almost always happy about adding `is_...` rather than not adding it. It will also pre-empt confusion with type annotations. WDYT?
> 
> I think that more intuitive and flexible than
> 
> ```python
> schema.optional = [feature1, feature2, feature3]
> ```
> 
> would be
> 
> ```python
> schema.update_is_optional({feature1: True, feature2: False})
> ```
> 
> The mere attribute accessor is not so intuitive because changing the optionality state for some features is a quite sophisticated operation; much more than just changing a single attribute.

I think I still prefer `optional=True` because that's much shorter and pretty.

lamindb/models/feature.py
```diff
@@ -499,24 +500,10 @@ def save(self, *args, **kwargs) -> Feature:
...
-    @coerce_dtype.setter
-    def coerce_dtype(self, value: bool) -> None:
-        if self._aux is None:  # type: ignore
-            self._aux = {}  # type: ignore
-        if "af" not in self._aux:
-            self._aux["af"] = {}
-        self._aux["af"]["2"] = value
+    def with_config(self, optional: bool | None = None) -> Feature:
+        """Set required flag."""
+        self.optional = optional
```
falexwolf, line 505: I'd rename `self.optional` to `self._is_optional`. I don't think this should be a public attribute because it'll confuse users when they notice that this is not persisted with the feature. It's a one time thing.

I even think it's somewhat problematic to have the attribute at all because then somebody might naively create two schemas:

```python
schema1 = Schema(features=[feature1.with_config(is_optional=True)]
schema2 = Schema(features=[feature1]
```

In `schema2`, the feature1 would still be optional because the state is shared in memory even if it's not persisted to the database.

Hence, `schema2` doesn't behave the user expects it to behave from reading the code.

I suggest to do the following instead:
```python
def with_config(self, optional: bool | None = None) -> tuple[Feature, dict]:
   ...
```

This way the method is just a controlled way to combine the feature with some additional configuration without actually changing the feature object. WDYT?

lamindb/models/feature.py
```diff
@@ -532,12 +519,9 @@ def default_value(self) -> Any:
...
     @default_value.setter
-    def default_value(self, value: bool) -> None:
-        if self._aux is None:  # type: ignore
-            self._aux = {}  # type: ignore
-        if "af" not in self._aux:
-            self._aux["af"] = {}
-        self._aux["af"]["0"] = value
+    def default_value(self, value: str | None) -> None:
+        self._aux = self._aux or {}
+        self._aux.setdefault("af", {})["0"] = value
```
falexwolf, line 524: That is much prettier! 😄

sunnyosun:
Can `schema.update_is_optional` just be `optional`? I really don't like the `is_` thing, we've had it for ULabel types and they just look so visually cluttering and unprofessional. We already decided to not use it for ULabels, why introduce here? I don't think `optional` is less clear than `is_optional`.

lamindb/models/schema.py
```diff
@@ -213,7 +213,9 @@ class Meta(Record.Meta, TracksRun.Meta, TracksUpdates.Meta):
     For a composite schema, the hash of hashes.
     """
     minimal_set: bool = BooleanField(default=True, db_index=True, editable=False)
-    """Whether the schema contains a minimal set of linked features (default `True`).
+    """Deprecated. Use `optional` instead.
```
falexwolf, line 216: Do we still want to deprecate it? Isn't it useful to be able to set `minimal_set = False` in some cases and thereby make all features optional? 🤔

sunnyosun: OK!

lamindb/models/schema.py
```diff
@@ -579,11 +588,27 @@ def coerce_dtype(self) -> bool:
...
+            return (
+                self.members.filter(uid__in=self._aux["af"]["1"])
+                .order_by("links_schema__id")
+                .all()
+            )
+        else:
+            return Feature.objects.none()  # empty QuerySet
+
+    @optional.setter
+    def optional(self, value: list[Record]) -> None:
```
falexwolf, line 607: I didn't find the test for this.

Like with all other schema properties, we should list out all datasets that would violate the change.

Say, if somebody makes a feature required that was previously optional, certain datasets that were previously validated would no longer validate.

This is all super complicated and I wonder whether we can take a pragmatic approach by printing a warning instead of trying to query all datasets and testing whether features were linked or not.

lamindb/models/schema.py
```diff
@@ -426,13 +471,16 @@ def from_values(  # type: ignore
         Raises:
             ValidationError: If some values are not valid.
 
-        Examples:
+        Example::
```
falexwolf, line 474: Just a note: I recently realized that only `Examples:` is recognized as a subheading. It looks much prettier than `Examples::`.

But I don't know how to achieve the code formatting then. 🤔

lamindb/models/schema.py
```diff
@@ -213,18 +294,14 @@ class Meta(Record.Meta, TracksRun.Meta, TracksUpdates.Meta):
     For a composite schema, the hash of hashes.
     """
     minimal_set: bool = BooleanField(default=True, db_index=True, editable=False)
```
falexwolf, line 296: Please make a comment here that this is deprecated.

And add to the list of future migrations:

- https://github.com/laminlabs/lamindb/issues/2604

lamindb/models/schema.py
```diff
@@ -318,28 +394,33 @@ def __init__(
         type: Feature | None = kwargs.pop("type", None)
         is_type: bool = kwargs.pop("is_type", False)
         otype: str | None = kwargs.pop("otype", None)
-        minimal_set: bool = kwargs.pop("minimal_set", True)
```
falexwolf, line 321: Please don't hard remove right away. Instead, if a user supplies this, print a warning that the setting will be ignored and that the user should instead use `Feature().with_config(optional=True)` for __all__ features.

falexwolf (approved):
Good!

falexwolf:
> Deprecate Schema.minimal_set. Ignores minimal_set arg passed to Schema.

@sunnyosun, can you please comprehensively explain which edge cases appear if we continue to support `minimal_set` in addition to the more fine-grained optional features?

PR #2637, opened 2025-04-03
♻️ Eliminate monkey patching of `django.db.models.QuerySet` and `django.db.models.Manager`

Koncopd:
This PR removes monkey patching of `django.db.models.QuerySet` and `django.db.models.Manager`.

Our custom `QueryManager` is now used as a base `Manager` of `BasicRecord`, thus enforcing that all `lamindb` models use it by default (through `objects`).

Instead of monkey patching of `django.db.models.QuerySet`, this PR introduces a subclass `BasicQuerySet`, that is returned by `QueryManager`.

`Record.filter()` returns a subclass of `BasicQuerySet` called `QuerySet` that differs from `BasicQuerySet` only in `.filter `and `.get` re-implementation. The split between `BasicQuerySet` and `QuerySet` is needed because other problems arise if `.filter `and `.get` are redefined in the base query set class returned by the base `Manager`.


lamindb/models/record.py
```diff
@@ -666,8 +666,11 @@ class BasicRecord(models.Model, metaclass=Registry):
     It's mainly used for LinkORMs and similar.
     """
 
+    objects = QueryManager()
+
     class Meta:
         abstract = True
+        base_manager_name = "objects"
```
Koncopd, line 673: Looks like this triggers a migration.

Koncopd:
Ok, this works.

falexwolf (approved):
This looks great, but please add _extensive_ and understandable prose to the PR at the top because it's otherwise not easy to understand these changes.

image

Koncopd:
Sure, sure, i wasn't sure just if the migration is fine or we don't want it to the point of discarding this.

falexwolf:
Can you create the migration file, please? I'm 99% sure it's a non-breaking migration but it'd be great to take a look.

It's a good moment for this because I'll make several non-breaking migrations in the upcoming days and deploy them soon.

Koncopd:
@falexwolf I have added, but it looks pretty strange, hard to understand, why it is needed at all.

falexwolf:
Can you give examples for these problems or try to describe them?

image

falexwolf:
Ok, I think I got it from looking at the code. All good! Please merge!

Koncopd:
>Can you give examples for these problems or try to describe them?

Internal `django` functions start to use our redefined `.get` and fail in unexpected ways. I didn't find any way to circumvent this.

PR #2640, opened 2025-04-04
✨ Add a stub for a history table

falexwolf:
Roughly based on design here for now:

- https://github.com/laminlabs/laminhub/issues/1787


lamindb/models/history.py
```diff
@@ -0,0 +1,31 @@
+from datetime import datetime
+from uuid import UUID
+
+# from .record import User --> only needed for ForeignKey
+# from django.contrib.contenttypes.models import ContentType --> only needed for ForeignKey
```
falexwolf, line 5: I think it might be less hassle to not actually work with foreign key constraints in this context; also less space -- the constraints and indexes are the most costly

lamindb/models/history.py
```diff
@@ -0,0 +1,31 @@
...
+        primary_key=True
+    )  # snowflake ID would be ideal because then we'd have some ordering; otherwise timsstamped uuid also ok
+    table_id: int = SmallIntegerField()  # ForeignKey into ContentType but unprotected so that tables can be deleted; so not using ForeignKey
+    operation: int = (
+        SmallIntegerField()
+    )  # code INSERT, DELETE, UPDATE and other things we might want
+    record_id: int = BigIntegerField()
+    values: dict = JSONField()
+    created_at: datetime = DateTimeField(
+        db_default=functions.Now(),
```
falexwolf, line 29: This operates on the database level.

PR #2641, opened 2025-04-04
🐛 Set db token only on the transaction level

Koncopd:
Paired with:

- https://github.com/laminlabs/lamindb-setup/pull/1015


falexwolf (approved):
Looks great!

PR #2642, opened 2025-04-05
✨ Add a `writelog` table, implement trigger installation

alexras:
This diff introduces a set of tables for tracking edits to LaminDB's database, and a mechanism for installing Postgres database triggers that write to those tables.

Supersedes:

- https://github.com/laminlabs/lamindb/pull/2640


lamindb/models/history.py
```diff
@@ -0,0 +1,53 @@
...
+    """Stores the edit history for all LaminDB tables."""
+
+    seqno = models.AutoField(primary_key=True)
+    migration_history_id = models.BigIntegerField()
+    id = models.UUIDField()
+    table_name = models.CharField(max_length=100)
+    record_id = models.BigIntegerField()
+    record_data = models.JSONField(null=True)
+    event_type = models.PositiveSmallIntegerField()
+    created_at = models.DateTimeField(auto_now_add=True)
```
falexwolf, line 14: The `auto_now_add` is likely not going to work because it operates on the Python level.

See here: https://github.com/laminlabs/lamindb/pull/2640/files#r2038202637

lamindb/models/history.py
```diff
@@ -0,0 +1,53 @@
+from django.db import models
+
+
+class History(models.Model):
+    """Stores the edit history for all LaminDB tables."""
+
+    seqno = models.AutoField(primary_key=True)
```
falexwolf, line 7: I still don't understand how an auto-incrementing integer helps in presence of two copies of the database that are separately evolving. Let's discuss this on a call tomorrow morning! ☺️

lamindb/models/history.py
```diff
@@ -0,0 +1,53 @@
...
+    migration_history_id = models.BigIntegerField()
+    id = models.UUIDField()
+    table_name = models.CharField(max_length=100)
```
falexwolf, line 10: If we need to track chars here let's be save and not limit at `max_length=100` but at least allow `255`. Because django concatenates app name and table name and table names can indeed get long that's an unnecessary risk.

Given we have an enumeration of tables I still think that having an id rather than a char name would be nice.

lamindb/models/history.py
```diff
@@ -0,0 +1,53 @@
+from django.db import models
+
+
+class History(models.Model):
```
falexwolf, line 4: I have a `created_by` field in the history table here:

- https://github.com/laminlabs/lamindb/pull/2640

Doesn't have to be part of this version. Just FYI.

lamindb/models/history.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class HistoryLock(models.Model):
+    """A lock table for history triggers.
+
+    History triggers for a given table record all changes to
+    that table. We need to make sure that when we're replaying
+    history from another node, we don't record the replayed
+    changes as additional history. We'll accomplish that in a
+    database-agnostic way using this lock table.
```
falexwolf, line 28: Why do you want to solve it with this additional table rather than "just excluding the history table from the trigger definition"? 🤔

tests/core/history/test_postgres_trigger_installer.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def create_spy(obj, method_name):
+    original_method = getattr(obj, method_name)
+    mock = Mock(wraps=original_method)
+    setattr(obj, method_name, mock)
+    return mock
+
+
+class FakeCursor:
```
falexwolf, line 16: Cool!

lamindb/migrations/0092_historylock.py
```diff
@@ -0,0 +1,30 @@
...
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("lamindb", "0091_history"),
+    ]
+
+    operations = [
```
falexwolf, line 11: Can you just move this into the same migration file, but I can do this when refactoring the PR. ☺️

lamindb/models/history.py
```diff
@@ -0,0 +1,53 @@
+from django.db import models
+
+
+class History(models.Model):
+    """Stores the edit history for all LaminDB tables."""
+
+    seqno = models.AutoField(primary_key=True)
+    migration_history_id = models.BigIntegerField()
```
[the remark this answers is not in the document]
alexras: By "indexing into", do you mean that it should refer to the largest `id` present in `django_migrations` at the time that the history record is created?

I suppose that if we assume that both sides of a history synchronization have the same modules installed, then that would be a reasonable proxy for the state of the migrations (and it'd certainly be easier to track).

lamindb/core/history/_trigger_installer.py
```diff
@@ -0,0 +1,229 @@
...
+        rows = cursor.fetchall()
+
+        for row in rows:
+            table_name = row[0]
+            tables.add(table_name)
+
+        return tables
+
+
+class SQLiteHistoryRecordingTriggerInstaller(HistoryRecordingTriggerInstaller):
```
falexwolf, line 203: Just to be sure: we'll be able to add an sqlite version of the triggers in the future, right? I would ask AI to translate it from Postgres dialect to SQLite.

I'm also thinking to potentially create a test group only for the history tracking because it's entirely orthogonal to everything else.

It's the first feature that will truly differ across SQLite and Postgres and it's the first feature that operates on the database level. Everything else in lamindb is just python logic on top of Django and fsspec.

lamindb/models/history.py
alexras: I'm not sure the trigger can extract `created_by`, especially since both SQLAlchemy-driven and Django-driven operations can modify the tables' state. Let me know if there's an easy way to do that that I'm missing.

lamindb/core/history/_trigger_installer.py
```diff
@@ -0,0 +1,229 @@
...
+    def _get_db_tables(self) -> set[str]:
+        """Get the schema and table names for all tables in the database for which we might want to track history."""
+        all_models = apps.get_models()
+
+        tables = set()
+
+        for model in all_models:
+            # Record the model's underlying table, as well as
+            # the underlying table for all of its many-to-many
+            # relationships
```
falexwolf, line 62: I think it helps here that Django doesn't define ManyToMany symmetrically the way SQLAlchemy does. Hence just iterating over `_meta.many_to_many` seems indeed a simple and reliable solution. ☺️

lamindb/models/history.py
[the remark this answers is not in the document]
falexwolf: > By "indexing into", do you mean that it should refer to the largest id present in django_migrations at the time that the history record is created?

Yes.

> I suppose that if we assume that both sides of a history synchronization have the same modules installed, then that would be a reasonable proxy for the state of the migrations (and it'd certainly be easier to track).

Oh, you're right. The order of installation of Django apps between two copies of the database might differ and hence the content of the django migration table in these two copies. 🤔 -- This would give rise to a conflict of the two copies of the `django_migration` table across these two database copies. Because there is no `uid` for the record this conflict cannot be easily resolved. 🤔 

This makes me realize that the `record_id` is likely not a good thing to track across two copies of the database that are independently evolving. The history table should track the record `uid`, shouldn't it? The integer `record_id` will immediately start to diverge for any table that has independent inserts in each database copy. Only the record uid is valid across both instances.

The same records will be present in both database instances with different integer ids. This is the same situation that we have in the transfer flow: https://docs.lamin.ai/transfer -- when we map one record including all its relations from the source database into the database we recurse through foreign keys and then insert:

https://github.com/laminlabs/lamindb/blob/1eb9e30e9dc20ef2f26305e9dd02cfa863c10121/lamindb/models/record.py#L812-L814

The linked function also checks whether the record `uid` is already present in the target database and if it is doesn't transfer it from the source database.

But for the `django_migration` table, because django doesn't ship a uid, the integer id is indeed problematic. Hmm. Didn't think of this. 🤔

lamindb/models/history.py
falexwolf: We can likely get that from connection/transaction that's live that defines that variable. But let's not worry about it right now. Will be easy to add down the road.

lamindb/models/history.py
```diff
@@ -0,0 +1,53 @@
...
+    id = models.UUIDField()
+    table_name = models.CharField(max_length=100)
+    record_id = models.BigIntegerField()
```
falexwolf, line 11: I think we need to use the record uid here because the integer record ids will be meaningless across two copies of the database. See https://github.com/laminlabs/lamindb/pull/2642#discussion_r2038234500

lamindb/models/history.py
alexras: Now that I think about this a little more, I think it depends on how you want to handle history synchronization.

When you're talking about synchronizing history between a pair of nodes (I'm using "nodes" here as a shorthand to refer to LaminDB instances), either the two nodes share a single common history, or each node has its own separate history.

Suppose you have two nodes, $N_1$ and $N_2$. 

If the nodes share a common history (which is what I've been assuming, although perhaps I haven't been super explicit about it), then let's say $N_1$ has history records $h_1 ... h_i$ and $N_2$ has history records $h_1 ... h_j$. If $i < j$, $N_2$ has updates that $N_1$ doesn't. If $i > j$, $N_1$ has updates that $N_2$ doesn't. If $N_1$ wants to pull from $N_2$, it will pull all history records $h_{i + 1} ... h_j$ in turn and apply those changes to its own tables, then add $h_{i + 1} .. h_j$ to its history. 

While $N_1$ is updating its tables, we don't want to record history, because we're not creating _new_ history, we're just replaying history that $N_1$ doesn't have yet. That's why we need a way to temporarily turn off the triggers. SymmetricDS does this as well. 

If each node has its own history, let's say $N_1$ has history records $h_{N_1,1} ... h_{N_1,i}$ and $N_2$ has history records $h_{N_2,1} ... h_{N_2,j}$. If $N_1$ wants to pull changes from $N_2$, it will pull each history record from $N_2$ in turn and apply it to its own history, creating new history records $h_{N_1,i + 1} ... h_{N_1,i + j + 1}$. In this case, you _are_ creating new history on $N_1$, so you want its history triggers to be turned on while changes from $N_2$ are being applied.

Separate histories can get tricky if history changes introduce things like write-delete conflicts (where $N_2$ has deleted a record that $N_1$ has subsequently updated) and write-write conflicts (where both $N_1$ and $N_2$ have performed different writes to the same record). You also need to give history records a unique ID so you can skip records that you've already applied in the past, and you need to propagate that unique ID down to the history table somehow. 

All these reasons make me lean toward the model where all nodes share a single history. Does all this make sense?

lamindb/core/history/_trigger_installer.py
alexras: > Just to be sure: we'll be able to add an sqlite version of the triggers in the future, right?

Yup! The SQLite triggers won't look exactly the same, but they will look _very_ similar.

> I'm also thinking to potentially create a test group only for the history tracking because it's entirely orthogonal to everything else.

Sounds good! I'll move this out of core into its own top-level `history` module.

lamindb/core/history/_trigger_installer.py
alexras: Yeah, that was a pleasant surprise.

lamindb/models/history.py
alexras: Unfortunately `contenttypes`' enumeration of tables doesn't include many-to-many tables. I can create a separate table that tracks all table names to make sure that tables have unique IDs that are shared across instances if you think that'd be helpful.

lamindb/models/history.py
alexras: Good point. We're making the trigger add this anyway, so we don't need that annotation in Django-land.

lamindb/models/history.py
alexras: Really what we want is some way of ordering history records that's not dependent on, say, the resolution of the record's `created_at` timestamp (since multiple records could be created in the same milliseconds). Making that sequence number auto-increment is an easy way to solve this if you've got separate histories on each node since sequence numbers only need to be local to a given node in that case. If we're sharing history between nodes, we may want to make this non-auto-incrementing and set it to `MAX(seqno) + 1` every time we add a record.

falexwolf: Ok!

lamindb/models/history.py
falexwolf: > I can create a separate table that tracks all table names to make sure that tables have unique IDs
I think the table of tables would be helpful, yes. 🤔

lamindb/models/history.py
falexwolf: All this makes sense. And I know see that during sync, you don't want to use triggers to "implicitly backfill missing history records", but you want to "explicitly populate the history table with missing records as part of the sync process". This totally makes sense because only through that you'll have the correct time stamps and hence a single history table across all nodes.

lamindb/models/history.py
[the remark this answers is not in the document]
falexwolf: > That'll make the trigger logic somewhat more complicated, but I think if we introspect on a given table's foreign key relationships, we can figure out how to store UIDs everywhere we would store IDs in the history table.

Yes!

falexwolf:
All of this looks great, @alexras! Will do a detailed review once you're getting the PR out of draft mode. Likely I won't find much to improve.

tests/history/conftest.py
```diff
@@ -0,0 +1,43 @@
...
+
+def pytest_sessionstart():
+    t_execute_start = perf_counter()
+
+    ln_setup._TESTING = True
+    pgurl = setup_local_test_postgres()
+    ln.setup.init(
+        storage="./default_storage_unit_core",
+        modules="bionty",
+        name="lamindb-unit-tests-core",
```
Koncopd, line 24: This should have different `name` and `storage` to avoid clashes, these are already used by unit-core.

falexwolf: This should be an entirely local test suite; there is no need to to hit cloud resources as it's self-contained.

lamindb/history/_trigger_installer.py
```diff
@@ -0,0 +1,564 @@
...
+from ._types import KeyConstraint, UIDColumns
+
+
+class HistoryEventTypes(enum.Enum):
+    INSERT = 0
+    UPDATE = 1
+    DELETE = 2
+
+
+class DjangoMigration(models.Model):
```
falexwolf, line 27: Note for @falexwolf, potentially de-duplicate with the below, but for now better to keep `history` self-contained.
 https://github.com/laminlabs/lamindb/blob/6537063258d341d11606a73cf964584f2c1092d4/lamindb/models/record.py#L1654

lamindb/history/_trigger_installer.py
```diff
@@ -0,0 +1,564 @@
...
+                event_type := {HistoryEventTypes.INSERT.value};
+            ELSIF (TG_OP = 'UPDATE') THEN
+                event_type := {HistoryEventTypes.UPDATE.value};
+            END IF;
+        END IF;
+
+        INSERT INTO lamindb_history
+            (id, history_migration_state_id, table_id, record_uid, record_data, event_type, created_at)
+        VALUES
+            (gen_random_uuid(), latest_migration_id, {table_id_var}, record_uid, record_data, event_type, now());
```
falexwolf, line 504: Can we replace this with an 18-char base62 random string instead of a UUID? -- Also see the comment in the schema definition.

```sql
CREATE OR REPLACE FUNCTION base62(n_char integer) RETURNS text AS $$
DECLARE
    alphabet text := '0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ';
    result text := '';
    i integer;
BEGIN
    FOR i IN 1..n_char LOOP
        result := result || substr(alphabet, 1 + floor(random() * 62)::integer, 1);
    END LOOP;
    RETURN result;
END;
$$ LANGUAGE plpgsql;
```

lamindb/models/history.py
falexwolf: Let's double-check: the sequences in node 1 & 2 increment independently from one another. If we backfill the history table of 1 with new records present in 2, the history table in node 1 will start to deviate from the one in node 2. Isn't this a problem?

--

If it's no problem, we should use a `BigAutoField` because the history table will be the largest table and might exceed 2B rows at some point.

lamindb/models/history.py
```diff
@@ -0,0 +1,89 @@
...
+
+class History(models.Model):
+    """Stores the edit history for all LaminDB tables."""
+
+    seqno = models.AutoField(primary_key=True)
+    history_migration_state = models.ForeignKey(
+        HistoryMigrationState, on_delete=models.PROTECT
+    )
+    table = models.ForeignKey(HistoryTableState, on_delete=models.PROTECT)
+    id = models.UUIDField()
```
falexwolf, line 45: Can we call this `uid` and make it a `base62` string? I added code for this up there: https://github.com/laminlabs/lamindb/pull/2642/files#r2081960745

The user facing links to things are always based on a `uid` similar to how stripe or modal or other platforms do it.

This really is mostly a design/aesthetic decision (sharing an id like `3ixi4FetqaJk` and links like this https://lamin.ai/vitessce/examples/transform/3ixi4FetqaJk is just a bit more user friendly)

Sometimes it's also nice that one can encode information in some of the chars more easily than in the UUID.

And 16 byte vs. 18 byte don't make a difference.

lamindb/models/history.py
```diff
@@ -0,0 +1,89 @@
...
+    # Many-to-many tables don't have row UIDs, so this needs to be nullable.
+    record_uid = models.JSONField(null=True)
+    record_data = models.JSONField(null=True)
+    event_type = models.PositiveSmallIntegerField()
+    created_at = models.DateTimeField()
```
falexwolf, line 50: We should add a `space_uid` that's based on the `record.space.uid` of the record whose history is logged. This makes a great chunk of access management for the history table easier (RLS is easier). It should be trivial because you have logic for this already above.

We should add a `created_by_uid` string field that's automatically populated based on the database connection that codes the user identity. Lamin Team can handle this.

tests/history/conftest.py
```diff
@@ -0,0 +1,43 @@
...
+        db=pgurl,
+    )
+    ln.setup.register()  # temporarily
```
falexwolf, line 27: Not needed, is it? It makes everything slow and complicated.

falexwolf (commented):
Just discussed: to disambiguate from other type of histories of the lamindb instance, will rename `history` to `writelog`.

lamindb/models/history.py
falexwolf: After discussion with @alexras I now understand that this is no issue because we're not replaying this field.

lamindb/models/history.py
falexwolf: Please add a `branch_code = models.SmallIntegerField()`.

falexwolf: Please add a `run_uid = CharField(max_length=20)`.

falexwolf (approved):

PR #2644, opened 2025-04-07
♻️ Organize categorical curation code with `CatColumn`

sunnyosun:
- Add `CatColumn` to handle `validate`, `add_new` for each categorical
- Deprecate `MuDataCatManager` and `SpatialDataCatManager` with refactors
- Deprecated passing `all` to `standardize` and `add_new_from` in favor of using `.non_validated.keys()`
- Remove deprecated method `add_new_from_columns`

@falexwolf What's still not great is the transformation of var matrix because it messes up object synchronization, I think we should introduce curating index instead.

### DIFF
--- a/lamindb/curators/__init__.py
+++ b/lamindb/curators/__init__.py
@@ -275,7 +275,8 @@ class SlotsCurator(Curator):
     @doc_args(VALIDATE_DOCSTRING)
     def validate(self) -> None:
         """{}"""  # noqa: D415
-        for _, curator in self._slots.items():
+        for slot, curator in self._slots.items():
+            logger.info(f"validating slot {slot} ...")
             curator.validate()
 
     @doc_args(SAVE_ARTIFACT_DOCSTRING)
@@ -561,7 +562,7 @@ class AnnDataCurator(SlotsCurator):
                 ln.Feature(name="perturbation", dtype="cat[ULabel[Perturbation]]").save(),
                 ln.Feature(name="sample_note", dtype=str).save(),
                 ln.Feature(name="cell_type_by_expert", dtype=bt.CellType).save(),
-                ln.Feature(name="cell_type_by_model", dtype=bt.CellType").save(),
+                ln.Feature(name="cell_type_by_model", dtype=bt.CellType).save(),
             ],
         ).save()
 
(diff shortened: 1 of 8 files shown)


lamindb/curators/__init__.py
```diff
@@ -609,6 +610,12 @@ def __init__(
             for slot, slot_schema in schema.slots.items()
             if slot in {"obs", "var", "uns"}
         }
+        # TODO: better way to handle this!
```
falexwolf, line 613: I don't get this, but just talked in person.

lamindb/curators/__init__.py
```diff
@@ -911,6 +918,296 @@ def __init__(
         self._columns_field = self._var_fields
 
 
+class CatColumn:
+    """Categorical column for `DataFrame`.
```
falexwolf, line 922: Why not call it `CatColumManager` or something like this?

lamindb/curators/__init__.py
```diff
@@ -911,6 +918,296 @@ def __init__(
...
+        key: The name of the column to validate. Only used for logging.
+        values_setter: A callable that sets the values.
+        organism: The organism to validate against.
+        source: The source to validate against.
+    """
+
+    def __init__(
+        self,
+        values_getter: Callable | Iterable[str],
+        field: FieldAttr,
```
falexwolf, line 936: Why doesn't this take a `dtype` so that one can express things like sub-types or species or ontology versions?

tests/curators/test_dataframe_curators_accounting_example.py
```diff
@@ -128,8 +128,8 @@ def test_invalid_label(transactions_schema):
     schema = ln.Schema.get(name="transaction_dataframe")
     curator = ln.curators.DataFrameCurator(invalid_df, schema)
 
-    with pytest.raises(ln.errors.ValidationError) as err:
+    with pytest.raises(ln.errors.ValidationError):
         curator.validate()
-    exconly = """lamindb.errors.ValidationError: 1 term is not validated: 'GBP'
-    → fix typos, remove non-existent values, or save terms via .add_new_from("currency_name")"""
-    assert err.exconly() == exconly
+    # exconly = """lamindb.errors.ValidationError: 1 term is not validated: 'GBP'
```
falexwolf, line 133: Please get this back!

falexwolf (approved):
Very good step forward! 😄

tests/curators/test_dataframe_curators_accounting_example.py
sunnyosun: Next PR

lamindb/curators/__init__.py
sunnyosun: New features are not yet implemented, this PR is only refactoring existing code.

lamindb/curators/__init__.py
sunnyosun: Not sure, because `ColumnCatManager` sounds like it should inherit from `CatManager`.

PR #2645, opened 2025-04-07
⬇️ Fix `numcodecs==0.16.0` incompatibility with `zarr v2`

Koncopd:


Koncopd:
`permissions` test is fixed in another PR.

falexwolf (approved):

PR #2648, opened 2025-04-09
🐛 Fix re-validation of `var_index`

sunnyosun:
On lamindata running:
```python

df = pd.DataFrame(
    {"CD8A": [1, 2, 3], "CD4": [3, 4, 5], "CD14": [5, 6, 7], "perturbation": ["DMSO", "IFNG", "DMSO"], "cell_type_by_expert": ["B cell", "T cell", "T cell"], "cell_type_by_model": ["B cell", "T cell", "T cell"]},
    index=["sample1", "sample2", "sample3"],
)
adata1 = ad.AnnData(
    df[["CD8A", "CD4", "CD14"]], obs=df[["perturbation", "cell_type_by_expert", "cell_type_by_model"]]
)
curator = ln.Curator.from_anndata(adata1, var_index=bt.Gene.symbol, categoricals={"perturbation": ln.ULabel.name, "cell_type_by_expert": bt.CellType.name,  "cell_type_by_model": bt.CellType.name}, organism="human")
```

Before | After
--- | ---
Screenshot 2025-04-09 at 15 29 10 | Screenshot 2025-04-09 at 16 05 31

### DIFF
--- a/lamindb/curators/__init__.py
+++ b/lamindb/curators/__init__.py
@@ -948,7 +948,6 @@ class CatColumn:
         self._validated: None | list[str] = None
         self._non_validated: None | list[str] = None
         self._synonyms: None | dict[str, str] = None
-        self._last_values_hash = self._compute_hash(self.values)
 
     @property
     def values(self):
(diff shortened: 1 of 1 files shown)


falexwolf:
I'm a bit skeptical. 🤔 

I don't think that other validation library do this for the good reason that the user expects `.validate()` to _actually_ validate. Hashing is a somewhat complicated topic and you're introducing new ad-hoc hashing functions that we need to maintain and that might have errors in them.

I see the value when users curate "column by column" though. So, this might indeed justify the additional complexity.

falexwolf:
In the best case validation should be as fast as hashing. I don't think we're there but in theory validation could be comparably fast I guess.

PR #2668, opened 2025-04-14
🚸 Enable to validate & annotate features *without* passing them to a schema

falexwolf:
## Summary

This PR introduces a simple way to validate & annotate datasets without having to spell out all involved features. A schema with `flexible == True` will validate all known features matching the `itype` and automatically annotate categoricals. Before this PR, in such cases, only the feature identifiers themselves were validated and no annotation with labels took place.

```python
schema = ln.Schema(itype=ln.Feature).save()  # <-- if you're just defining the itype, this sets flexible = True
curator = ln.curators.DataFrameCurator(df, schema)
curator.save_artifact(key="my_dataset.parquet")
```

You can also validate & annotate features in addition to those that you're explicitly defining:
```python
schema = ln.Schema(
    features=[ln.Feature(name="my_required_feature", dtype=int).save()],
    flexible=True,
)
```

If you don't pass `flexible=True`, the curator will disregard features that aren't part of the schema
```python
schema = ln.Schema(
    features=[ln.Feature(name="my_required_feature", dtype=int).save()],
)
```

## Context

Avoiding to spell out features is important because it frees the user from updating schemas with new features. While the previous PR enabled the later it still means considerable work for the user that's only justified for a highly-curated dataset; not for a dataset that a user wants to "quickly validate & annotate" in daily operations.

- https://github.com/laminlabs/lamindb/pull/2635

## Outlook

In the future, for users who work with well-organized data that should typically pass validation, this could make it very easy to indicate ingestion of a validated & annotated artifact via:

```python
ln.Artifact.from_df(..., schema=schema).save()
```

## Docs changes

Before | After
--- | ---
image | image
/ | image

## Materials

Internal [Notion page](https://www.notion.so/laminlabs/Overhaul-the-Schema-design-1ca2aeaa55e1803cb3fed41f955fcc7c).

### DIFF
--- a/lamindb/curators/__init__.py
+++ b/lamindb/curators/__init__.py
@@ -395,6 +395,7 @@ class DataFrameCurator(Curator):
         categoricals = []
         features = []
         if schema.mode == "validate-all-annotate-cat":
+            print(self._dataset.columns)
             features += Feature.filter(name__in=self._dataset.columns).list()
         if schema.n > 0:
             features += schema.features.all().list()
@@ -447,8 +448,6 @@ class DataFrameCurator(Curator):
                 strict=schema.maximal_set,
                 ordered=schema.ordered_set,
             )
-        else:
-            assert schema.itype is not None  # noqa: S101
         self._cat_manager = DataFrameCatManager(
             self._dataset,
             columns=parse_cat_dtype(schema.itype, is_itype=True)["field"],
--- a/lamindb/curators/__init__.py
+++ b/lamindb/curators/__init__.py
@@ -395,7 +395,7 @@ class DataFrameCurator(Curator):
         categoricals = []
         features = []
         feature_ids: set[int] = set()
-        if schema.mode == "validate-all-annotate-cat":
+        if schema.mode == "all-itype":
             features += Feature.filter(name__in=self._dataset.keys()).list()
             feature_ids = {feature.id for feature in features}
         if schema.n > 0:
--- a/lamindb/curators/__init__.py
+++ b/lamindb/curators/__init__.py
@@ -395,7 +395,7 @@ class DataFrameCurator(Curator):
         categoricals = []
         features = []
         feature_ids: set[int] = set()
-        if schema.mode == "all-itype":
+        if schema.itype is not None:
             features += Feature.filter(name__in=self._dataset.keys()).list()
             feature_ids = {feature.id for feature in features}
         if schema.n > 0:
--- a/lamindb/curators/__init__.py
+++ b/lamindb/curators/__init__.py
@@ -396,7 +396,7 @@ class DataFrameCurator(Curator):
         features = []
         feature_ids: set[int] = set()
         if schema.mode == "validate-all-annotate-cat":
-            features += Feature.filter(name__in=self._dataset.columns).list()
+            features += Feature.filter(name__in=self._dataset.keys()).list()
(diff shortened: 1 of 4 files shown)


lamindb/curators/__init__.py
```diff
@@ -393,12 +393,18 @@ def __init__(
     ) -> None:
         super().__init__(dataset=dataset, schema=schema)
         categoricals = []
+        features = []
+        if schema.mode == "validate-all-annotate-cat":
+            print(self._dataset.columns)
```
sunnyosun, line 398: remove

lamindb/curators/__init__.py
```diff
@@ -393,12 +393,18 @@ def __init__(
...
+            features += Feature.filter(name__in=self._dataset.columns).list()
```
sunnyosun, line 399: this list is not unique.

lamindb/curators/__init__.py
```diff
@@ -393,12 +393,18 @@ def __init__(
...
         if schema.n > 0:
+            features += schema.features.all().list()
```
sunnyosun, line 401: This list is not unique

lamindb/curators/__init__.py
```diff
@@ -393,12 +393,18 @@ def __init__(
     ) -> None:
         super().__init__(dataset=dataset, schema=schema)
         categoricals = []
+        features = []
+        if schema.mode == "validate-all-annotate-cat":
```
sunnyosun, line 397: "validate-all-annotate-cat" is way too long and i won't know what it does.

lamindb/models/schema.py
```diff
@@ -166,13 +168,15 @@ class Schema(Record, CanCurate, TracksRun):
             their corresponding :class:`~lamindb.Schema` objects for composite schemas.
         name: `str | None = None` A name.
         description: `str | None = None` A description.
-        dtype: `str | None = None` The simple type. Defaults to
-            `None` for sets of :class:`~lamindb.Feature` records.
-            Otherwise defaults to `"num"` (e.g., for sets of :class:`~bionty.Gene`).
-        itype: `str | None = None` The feature identifier type (e.g. :class:`~lamindb.Feature`, :class:`~bionty.Gene`, ...).
+        mode: `Literal["validate-passed", "validate-all-annotate-cat"] | None = None`
```
sunnyosun, line 171: unclear after reading this.

lamindb/models/schema.py
```diff
@@ -248,7 +247,8 @@ class Meta(Record.Meta, TracksRun.Meta, TracksUpdates.Meta):
     _name_field: str = "name"
     _aux_fields: dict[str, tuple[str, type]] = {
         "0": ("coerce_dtype", bool),
-        "1": ("_index_feature_uid", str),
+        "1": ("optionals", str),
```
sunnyosun, line 250: optionals is a list

sunnyosun (changes requested):

sunnyosun:
I think neither of these two options tell me what it does ["validate-passed", "validate-all-annotate-cat"], even after reading the doc string.

sunnyosun (approved):
Great!

falexwolf:
Ok, @sunnyosun & I realized we need another argument. This argument is now called `flexible` in lose analogy to pandera's `strict`.

PR #2669, opened 2025-04-14
✨ Schema from/to pydantic

sunnyosun:


falexwolf:
Maybe a thing we need at the same time is a `DictCurator`? Given `pydantic` defines a dict schema?

Issue #2663, opened 2025-04-11
`ln.track()` without parameters does not detect notebook renames


Zethson:
### Add a description

To reproduce:

1. have a notebook `test2.ipynb` with:

```python
import lamindb as ln
ln.track()
ln.finish()
```

2. Execute it and notice that the Transform looks fine. It has source code etc. Great.
3. Rename the notebook to `test.ipynb` and execute it again.
4. Look at Transforms:

```python
ln.Transform.df()
```

<details>
<summary>DataFrame Details (click to expand)</summary>

<div>

<table border="1" class="dataframe">
  <thead>
    <tr style="text-align: right;">
      <th></th>
      <th>uid</th>
      <th>key</th>
      <th>description</th>
      <th>type</th>
      <th>source_code</th>
      <th>hash</th>
      <th>reference</th>
      <th>reference_type</th>
      <th>space_id</th>
      <th>_template_id</th>
      <th>version</th>
      <th>is_latest</th>
      <th>created_at</th>
      <th>created_by_id</th>
      <th>_aux</th>
      <th>_branch_code</th>
    </tr>
    <tr>
      <th>id</th>
      <th></th>
      <th></th>
      <th></th>
      <th></th>
      <th></th>
      <th></th>
      <th></th>
      <th></th>
      <th></th>
      <th></th>
      <th></th>
      <th></th>
      <th></th>
      <th></th>
      <th></th>
      <th></th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <th>2</th>
      <td>9XyhSy9UZz3X0000</td>
      <td>test.ipynb</td>
      <td>test.ipynb</td>
      <td>notebook</td>
      <td>None</td>
      <td>None</td>
      <td>None</td>
      <td>None</td>
      <td>1</td>
      <td>None</td>
      <td>None</td>
      <td>True</td>
      <td>2025-04-11 13:07:24.894000+00:00</td>
      <td>1</td>
      <td>None</td>
      <td>1</td>
    </tr>
    <tr>
      <th>1</th>
      <td>OvzSPYnIJWKH0000</td>
      <td>test2.ipynb</td>
      <td>None</td>
      <td>notebook</td>
      <td># %%\n# !lamin init --storage run-tests --modu...</td>
      <td>OhXdXtk8nl4f0KAqYXcx_A</td>
      <td>None</td>
      <td>None</td>
      <td>1</td>
      <td>None</td>
      <td>None</td>
      <td>True</td>
[...]

PR #2670, opened 2025-04-14
🐛 Fix rename notebooks

Zethson:
Fixes https://github.com/laminlabs/lamindb/issues/2663

- Now calculating the hash for notebooks based on the py script source code
- Moves the necessary code from `finish()` to `_context`
- Has some performance penalty - let's see how much we can optimize. It should only be a fall back


falexwolf:
Closing in favor of:

- https://github.com/laminlabs/lamindb/pull/2718

Specifically, this commit solves the issue that this PR tried to solve: https://github.com/laminlabs/lamindb/pull/2718/commits/af965c6c460e8dbdda4bc1cc23342e84fc6aeec4

PR #2676, opened 2025-04-15
✨ Support curating lists as values in `DataFrameCurator`

sunnyosun:
This PR supports curation of lists as values in `DataFrameCurator`.

Note: a mix of str and list is not allowed in parquet. So we don't support `"sample_id": [["sample1", "sample2"], "sample2", "sample3"]`.

This is not supported for `AnnDataCurator` as saving lists to `.h5ad` gives `TypeError: Can't implicitly convert non-string objects to strings`. 

```python
df = pd.DataFrame(
    {
        "sample_id": [["sample1", "sample2"], ["sample2"], ["sample3"]],
        "dose": [[1.2, 2.3], [1.2], [2.3]],
        "cell_type": [["B cell", "T cell"], ["B cell"], ["T cell"]],
        "tissue": [["blood", "pulmo"], ["blood"], ["lung"]],
    }
)

schema = ln.Schema(
    name="lists schema",
    features=[
        ln.Feature(name="sample_id", dtype=list[str]).save(),
        ln.Feature(name="dose", dtype=list[float]).save(),
        ln.Feature(name="cell_type", dtype=list[str]).save(),
        ln.Feature(name="tissue", dtype=list[bt.Tissue]).save(),
    ],
).save()
```

Screenshot 2025-04-16 at 13 04 36
Screenshot 2025-04-16 at 13 05 03


tests/curators/test_curators_multivalue.py
```diff
@@ -0,0 +1,48 @@
...
+        }
+    )
+
+
+@pytest.fixture(scope="module")
+def lists_schema():
+    schema = ln.Schema(
+        name="lists schema cat",
+        features=[
+            ln.Feature(name="sample_id", dtype=str).save(),
```
falexwolf, line 25: This should be typed as `list[str]` not as `str`.

tests/curators/test_curators_multivalue.py
```diff
@@ -0,0 +1,48 @@
...
+            ln.Feature(name="dose", dtype=float).save(),
+            ln.Feature(name="cell_type", dtype=str).save(),
+            ln.Feature(name="tissue", dtype=bt.Tissue).save(),
```
falexwolf, line 28: This should be typed as `list[bt.Tissue]`.

tests/curators/test_curators_multivalue.py
```diff
@@ -0,0 +1,48 @@
...
+    )
+
+
+@pytest.fixture(scope="module")
+def lists_schema():
+    schema = ln.Schema(
+        name="lists schema cat",
+        features=[
+            ln.Feature(name="sample_id", dtype=str).save(),
+            ln.Feature(name="dose", dtype=float).save(),
```
falexwolf, line 26: This should be typed as `list[float]`.

tests/curators/test_curators_multivalue.py
```diff
@@ -0,0 +1,48 @@
...
+    schema.delete()
+    ln.Feature.filter().delete()
+    bt.Tissue.filter().delete()
+
+
+def test_curator_df_multivalue(df, lists_schema):
+    curator = ln.curators.DataFrameCurator(df, lists_schema)
+    with pytest.raises(ValidationError):
+        curator.validate()
+    assert curator._cat_manager._cat_columns["tissue"]._validated == ["blood", "lung"]
```
falexwolf, line 43: This should use a public API for a test.

falexwolf (changes requested):
Sorry that it took me so long to look at this. 😔 

Apart from the suggestions in the comments, all screenshots should be replaced with screenshots of the docs so that we ensure that this is properly documented.

tests/curators/test_curators_multivalue.py
sunnyosun: No, it will be flexible.

tests/curators/test_curators_multivalue.py
sunnyosun: No, it will be flexible.

tests/curators/test_curators_multivalue.py
sunnyosun: No, it's always typed with the individual value.

tests/curators/test_curators_multivalue.py
falexwolf: We need to distinguish `float` from `list[float]` when we define the feature. These things couldn't be more different and all other systems also distinguish it.

tests/curators/test_curators_multivalue.py
falexwolf: See comment below. It should be typed as `list[bt.Tissue]`.

falexwolf:
> all screenshots should be replaced with screenshots of the docs so that we ensure that this is properly documented.

Please document this properly! 🙏

lamindb/curators/core.py
```diff
@@ -358,24 +358,56 @@ def save_artifact(
         )
 
 
+def _is_type_or_list_of_type(value, expected_type):
```
falexwolf, line 361: Why prefix this with an underscore but not `check_dtype`? I'd say this shouldn't be prefixed for consistency.

lamindb/curators/core.py
```diff
@@ -464,7 +502,11 @@ def __init__(
                     pandera_columns[feature.name] = pandera.Column(
                         dtype=None,
                         checks=pandera.Check(
-                            check_dtype(feature.dtype),
+                            check_dtype(
+                                "str"
```
falexwolf, line 506: Why is this "str" if dtype startswith "cat"?

falexwolf: It makes a big difference and we have to be consistent about validating categoricals vs strings of course.

lamindb/curators/core.py
```diff
@@ -452,7 +484,13 @@ def __init__(
                     required = feature.uid not in optional_feature_uids
                 else:
                     required = False
-                if feature.dtype in {"int", "float", "num"}:
+                # series.dtype is "object" if the column has lists types, e.g. [["a", "b"], ["a"], ["b"]]
+                are_lists = (
+                    False
+                    if feature.name not in self._dataset.keys()
+                    else isinstance(self._dataset[feature.name][0], (list, np.ndarray))
```
falexwolf, line 491: This isn't a robust way to check that the entire series has elements that are iterable. Can't we do anything better? 🤔

tests/curators/test_curators_multivalue.py
sunnyosun: No, we can't make two features both named "tissue"; one of dtype `cat[bt.Tissue]` and the other is `list[cat[bt.Tissue]]`. This will drive users crazy if they can't use they have to make a different `tissue` feature.

falexwolf: Yes, we cannot. But we have to be able to error in case the user wishes to avoid lists in a feature.

tests/curators/test_curators_multivalue.py
```diff
@@ -0,0 +1,52 @@
...
+    schema.delete()
+    ln.Feature.filter().delete()
+    bt.Tissue.filter().delete()
+
+
+def test_curator_df_multivalue(df, lists_schema):
+    curator = ln.curators.DataFrameCurator(df, lists_schema)
+    with pytest.raises(ValidationError):
+        curator.validate()
+    assert curator._cat_manager._cat_vectors.keys() == {"columns", "tissue"}
```
falexwolf, line 43: Prettier to use the public API: `curator.cat`

lamindb/__init__.py
```diff
@@ -112,7 +112,7 @@
 
 # ruff: noqa: I001
 # denote a release candidate for 0.1.0 with 0.1rc1, 0.1a1, 0.1b1, etc.
-__version__ = "1.5.2"
+__version__ = "1.5.3"
```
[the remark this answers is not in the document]
sunnyosun: No, this is correct, the PR was made when the version was 1.5.2

PR #2683, opened 2025-04-18
🐛 Do not delete version family if user wants to retain store by passing `storage=False` to `artifact.delete()`

falexwolf:
In case `artifact._overwrite_versions` was `True`, before this PR, the following call deleted all records of the version family:

```python
artifact.delete(storage=False)
```

Now it only deletes the selected record.

### DIFF
--- a/lamindb/models/artifact.py
+++ b/lamindb/models/artifact.py
@@ -2456,8 +2456,10 @@ class Artifact(Record, IsVersioned, TracksRun, TracksUpdates):
                 and self.is_latest
                 and (storage is None or storage)
             ):
-                # includes self
-                for version in self.versions.all():
+                logger.important(
+                    "deleting the entire version family because all versions of this artifact share the same store"
+                )
+                for version in self.versions.all():  # includes self
                     _delete_skip_storage(version)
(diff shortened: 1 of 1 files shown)


Koncopd:
But it is confusing, no? Because older versions of the artifact point to the latest version of the stored data.

falexwolf:
Hmmm, you're right. 🤔

How did I miss this?

falexwolf:
I'll make another PR that leaves it at the warning.

You're right, what I did here violates integrity. I'll make another PR and leave it at a warning or maybe another confirmation.

Koncopd:
More than that, I am pretty sure it was you who asked me to implement the deletion of the whole version family in this case.

falexwolf:
In fact the storage is False way of calling this is problematic for folders unless if passed for everything other than the latest record.

falexwolf:
I wanted to have the full version family deleted if storage is None or True.

But storage is False is another case.

I wanted to just keep the latest version in storage but not the latest record in some contrived setting, and thought this was something that should be allowed. But no. That shouldn't be allowed.

PR #2688, opened 2025-04-18
✨ Enable curating the index of a dataframe

falexwolf:
You can now pass `index` to the `Schema` constructor to validate an index in a `DataFrame` in complete analogy to validating columns.

image

In particular, if you use a categorical `dtype`, the dataset will be annotated by the labels found in the index. If you use `int` or `str`, it will merely be validated.

Note that the "index feature" will appear among `schema.features` despite the constructor passing it to the new argument `index`. This is for UX and consistency with other dataframe frameworks like polars or even parquet files themselves (who don't have an index) and the `pandera` constructor (who lists the `index` argument).

## Next PR

As seen in the below example, the `index` feature isn't yet marked as such in `describe()` and isn't yet part of the feature set that belongs to the `DataFrame`. Rather than patching the current `_add_set...` family of annotations to make it appear as part of the feature set, I rather want to refactor the feature set annotation to be based on `CatColumns` in another PR.

image

## Docs

Before | After
--- | ---
image | image

### DIFF
--- a/lamindb/curators/__init__.py
+++ b/lamindb/curators/__init__.py
@@ -399,7 +399,14 @@ class DataFrameCurator(Curator):
             features += Feature.filter(name__in=self._dataset.keys()).list()
             feature_ids = {feature.id for feature in features}
         if schema.n > 0:
-            schema_features = schema.features.all().list()
+            if schema._index_feature_uid is not None:
+                schema_features = [
+                    feature
+                    for feature in schema.features.all().list()
+                    if feature.uid != schema._index_feature_uid
+                ]
+            else:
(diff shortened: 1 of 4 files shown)


sunnyosun (changes requested):
What happens if one pass index= to a composite schema?

And could we now avoid doing the var transpose with this?

falexwolf:
> What happens if one pass index= to a composite schema?

This is silently ignored just as it's silently ignored when you pass features. We have to add more validation, agreed.

> And could we now avoid doing the var transpose with this?

Next PR!

PR #2694, opened 2025-04-22
♻️ Allow passing `key` to `save_vitessce_config()`

namsaraeva:
Previously, the function only accepted `description` parameter, which resulted in multiple Artifacts not being grouped by key.

### DIFF
--- a/lamindb/integrations/_vitessce.py
+++ b/lamindb/integrations/_vitessce.py
@@ -19,7 +19,9 @@ if TYPE_CHECKING:
 # "unit test": https://github.com/laminlabs/lamindb/blob/main/docs/storage/vitessce.ipynb
 # integration test & context: https://github.com/laminlabs/lamin-spatial/blob/main/docs/vitessce.ipynb
 def save_vitessce_config(
-    vitessce_config: VitessceConfig, description: str | None = None
+    vitessce_config: VitessceConfig,
+    key: str | None = None,
+    description: str | None = None,
 ) -> Artifact:
     """Validates and saves a `VitessceConfig` object.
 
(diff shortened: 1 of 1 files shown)


sunnyosun:
@falexwolf Was there a reason that `save_vitessce_config` only accepts `description` before?

sunnyosun:
@Koncopd do you know why the tests are failing?

Koncopd:
Checking

Koncopd:
This is on my side, fixing.

Koncopd:
Fixed here https://github.com/laminlabs/lamindb/pull/2697, pls approve

falexwolf:
> @falexwolf Was there a reason that save_vitessce_config only accepts description before?

This is just because it dates back more than a year when `description` still was the default way of doing things.

sunnyosun:
So is it OK if we suggest users to use key instead?

falexwolf:
Yes!

sunnyosun (approved):

falexwolf:
I now realize that this introduced some confusion because `description` is passed to `key` in `Collection` in the same function.

Because I also saw failing tests in lamin-spatial I believed something is wrong and I started fixes:

- https://github.com/laminlabs/lamindb/pull/2997

But now I realize that all fixes are bad and we should probably keep passing `description` to `key` of Collection. Still this is not so nice, and I'd say users should prefer to pass `description` rather than `key`.

I guess this was introduced to version vitessce artifacts, which makes sense. 🤔

falexwolf:
Resolved by fixing docs and adding comments: https://github.com/laminlabs/lamindb/commit/22720335f6d6b166207d718101109f30a093127c

PR #2699, opened 2025-04-24
🥅 Do not raise exceptions on problems with `copy_or_move_to_cache` within `Artifact.save`

Koncopd:
I have a ton of problems with moving cached files when running `lamin-cli` tests locally on Windows. 

It looks like something holds cached run logs open and this causes problems with moving them within `copy_or_move_to_cache` on Windows.

As `copy_or_move_to_cache` happens after successful saving, there is no need to fail at the end of the process, copying the cache is not absolutely necessary, it will work without it anyways. This prevents failing of `Artifact.save`  if errors inside `copy_or_move_to_cache` happen.

Also runs `lamin-cli` tests for https://github.com/laminlabs/lamin-cli/pull/129, as running locally on Windows is problematic.

### DIFF
--- a/lamindb/models/save.py
+++ b/lamindb/models/save.py
@@ -157,7 +157,10 @@ def check_and_attempt_upload(
             return exception
         # copies (if on-disk) or moves the temporary file (if in-memory) to the cache
         if os.getenv("LAMINDB_MULTI_INSTANCE") is None:
-            copy_or_move_to_cache(artifact, storage_path, cache_path)
+            try:
+                copy_or_move_to_cache(artifact, storage_path, cache_path)
+            except Exception as e:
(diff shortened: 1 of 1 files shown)


lamindb/models/save.py
```diff
@@ -157,7 +157,10 @@ def check_and_attempt_upload(
             return exception
         # copies (if on-disk) or moves the temporary file (if in-memory) to the cache
         if os.getenv("LAMINDB_MULTI_INSTANCE") is None:
-            copy_or_move_to_cache(artifact, storage_path, cache_path)
+            try:
```
falexwolf, line 160: Please add a comment here because it looks problematic to add a blank try & except.

Otherwise fine for me.

Koncopd: i will add.

PR #2702, opened 2025-04-24
♻️ Only toggle is_latest for notebooks once they have source code

falexwolf:


Koncopd:
Shouldn’t this be done rather in the UI? And looks like it was done here https://github.com/laminlabs/laminhub/pull/2409#event-17394930279

It might be confusing to not have `is_latest=True` on transforms that are actually latest, even though they are drafts. Isn’t the only problem with `is_latest=True` on drafts that they are not shown correctly in UI? If so, it should be fixed in the UI, no?

falexwolf:
Yes, @Koncopd, I meanwhile agree.

We went back and forth in the PR in laminhub on where to solve this. And for a moment, I thought we should do it in lamindb. But I changed my opinion.

https://github.com/laminlabs/laminhub/pull/2295#issuecomment-2828149197

Issue #2335, opened 2025-01-09
zarr 3 released today is not compatible with lamindb


Koncopd:
### Add a description

subj

Koncopd:
Added the upper bound here for now https://github.com/laminlabs/lamindb/pull/2334

Koncopd:
Uhhh, the new API is very different, it doesn't support a general mutable mapping and thus requires a ton of work to update.

Koncopd:
https://zarr.readthedocs.io/en/latest/user-guide/v3_migration.html#getting-ready-for-3-0

Koncopd:
Another problem is that now installing `zarr v3` crushes `lamindb` import.
![image](https://github.com/user-attachments/assets/9a5f9882-0668-4901-a690-44ae985ea18b)
unfortunately can't do much here because this comes from `anndata`.

Koncopd:
Reported here https://github.com/scverse/anndata/issues/1817

PR #2711, opened 2025-04-28
⬆️ Add compatibility with zarr v3

Koncopd:
Closes https://github.com/laminlabs/lamindb/issues/2335

This updates the internals of our code to support `anndata>=0.12.0rc1` with `zarr>=3.0.0`. 

`lamindb` still has `anndata<=1.14.0` with `zarr<3.0.0` in the requirements, but nothing breaks if newer `anndata` or `zarr` are installed.

Also refactors some zarr-related code.


Koncopd:
Ok, from the side of `lamindb` the compatibility can be added without much problems.

PR #2713, opened 2025-04-28
🚸 Do not annotate with more than `n_max_records = 1000`

falexwolf:
We want to offer a good out-of-the-box annotation experience.

Over the past years, we've annotated artifacts with the full feature sets they measure. For instance, for an `AnnData` object measuring 60k variables, we would parse, validate, and annotate an inferred schema with all 60k variables.

While this process works it turns out that no access pattern that we've observed so far needs this annotation. By contrast, datasets of this dimensionality are almost always joined in a way that anticipates non-matching dimensionality across datasets. Independent of whether joins are virtual or materialized, feature identifiers are compared based on the storage level; not the database level.

This opens up the possibility to save considerably in database resources by skipping annotation, while not sacrificing the important validation of these feature identifiers. Hence, this PR introduces a new default behavior in which annotations with features or labels above `n_max_records` are skipped. The annotating schema still displays the number of features and their type. Validation is unchanged. But the 60k gene records in the above example won't be linked unless the user sets `ln.settings.annotation.n_max_records = 60_000`. The default for `n_max_record` is `1_000`.

This number is to be understood per slot for feature annotations of schemas. For label annotations of artifacts similar reasoning holds and the number is to be understood per feature.

The decisive change in the updated tests is this:

```python
if n_max_records == 2:
    assert not artifact.features.slots[slot].members.exists()
else:
    assert artifact.features.slots[slot].members.df()[
        "ensembl_gene_id"
    ].tolist() == [
        "ENSG00000153563",
        "ENSG00000010610",
        "ENSG00000170458",
    ]
```

Full test:

```python
def test_anndata_curator_varT_curation():
    ln.Schema.filter(itype="bionty.Gene.ensembl_gene_id").delete()
    varT_schema = ln.Schema(
    ).save()
    slot = "var.T"
    components = {slot: varT_schema}
    anndata_schema = ln.Schema(
        otype="AnnData",
        components=components,
    ).save()
    for with_gene_typo in [True, False]:
        adata = datasets.small_dataset1(otype="AnnData", with_gene_typo=with_gene_typo)
        if with_gene_typo:
            with pytest.raises(ValidationError) as error:
                artifact = ln.Artifact.from_anndata(
                    adata, key="example_datasets/dataset1.h5ad", schema=anndata_schema
                ).save()
            assert error.exconly() == (
                f"lamindb.errors.ValidationError: 1 term not validated in feature 'columns' in slot '{slot}': 'GeneTypo'\n"
                f"    → fix typos, remove non-existent values, or save terms via: curator.slots['{slot}'].cat.add_new_from('columns')"
            )
        else:
            for n_max_records in [2, 4]:
                ln.settings.annotation.n_max_records = n_max_records
                artifact = ln.Artifact.from_anndata(
                    adata, key="example_datasets/dataset1.h5ad", schema=anndata_schema
                ).save()
                assert artifact.features.slots[slot].n == 3  # 3 genes get linked
                assert (
                    artifact.features.slots[slot].itype == "bionty.Gene.ensembl_gene_id"
                )
                if n_max_records == 2:
                    assert not artifact.features.slots[slot].members.exists()
                else:
                    assert artifact.features.slots[slot].members.df()[
                        "ensembl_gene_id"
                    ].tolist() == [
                        "ENSG00000153563",
                        "ENSG00000010610",
                        "ENSG00000170458",
                    ]
            anndata_schema.delete()
            varT_schema.delete()
```


sunnyosun:
Isn't this a breaking change then? Because user would expect everything to be linked as previously?

Also this will change the query results right? For instance `ln.Artifact.filter(feature_sets__genes=gene)` will not return all artifacts measured that gene anymore.

falexwolf:
> Isn't this a breaking change then? Because user would expect everything to be linked as previously?

Yes.

> Also this will change the query results right? For instance ln.Artifact.filter(feature_sets__genes=gene) will not return all artifacts measured that gene anymore.

Correct.

PR #2715, opened 2025-04-29
🥅 Catch and clarify RLS error on `Record.save()`

Koncopd:
Closes https://github.com/laminlabs/pfizer-lamin-usage/issues/201

RLS error looks like this now:
image

### DIFF
--- a/lamindb/models/record.py
+++ b/lamindb/models/record.py
@@ -870,7 +870,7 @@ class BasicRecord(models.Model, metaclass=Registry):
                     raise NoWriteAccess(
                         f"You’re not allowed to write to the space '{self.space.name}'.\n"
                         "Please contact an administrator of the space if you need write access."
-                    ) from e
+                    ) from None
                 else:
                     raise
             # call the below in case a user makes more updates to the record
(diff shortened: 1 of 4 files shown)


lamindb/models/record.py
```diff
@@ -843,20 +844,33 @@ def save(self, *args, **kwargs) -> Record:
...
+                        or "duplicate key value violates unique constraint" in error_msg
+                    )
+                ):
                     pre_existing_record = self.__class__.get(hash=self.hash)
                     logger.warning(
                         f"returning {self.__class__.__name__.lower()} with same hash: {pre_existing_record}"
                     )
                     init_self_from_db(self, pre_existing_record)
+                elif (
+                    isinstance(e, ProgrammingError)
```
falexwolf, line 866: Why not using the "PermissionError" instead of "ProgramminError"?

falexwolf (approved):

lamindb/models/record.py
falexwolf: InsufficientPrivilege error I mean ☺️

Koncopd: Because `django` throws `ProgrammingError` actually, `InsufficientPrivilege` is not the final one, it is an intermediate from `psycopg2`.

Issue #2725, opened 2025-05-01
Integrate polars into .open()


Koncopd:
### Add a description

Subj

PR #2730, opened 2025-05-02
✨ Allow to use `polars` in `Artifact.open()` and `Collection.open()`

Koncopd:
Closes https://github.com/laminlabs/lamindb/issues/2725

This PR adds `polars` support to `Artifact.open()` and `Collection.open()`.

Examples:
```python
...
# an artifact with .parquet file, for example
with artifact.open(df_engine="polars") as lazy_df: # lazy_df is polars.LazyFrame
    df = lazy_df.collect().to_pandas() # here just loads into memory, but can be anything
# a collection with .parquet files, for example
with collection.open(df_engine="polars") as lazy_df: # lazy_df is polars.LazyFrame
    df = lazy_df.collect().to_pandas()
```

### DIFF
--- a/lamindb/models/artifact.py
+++ b/lamindb/models/artifact.py
@@ -48,6 +48,10 @@ from ..core.storage import (
     write_to_disk,
 )
 from ..core.storage._anndata_accessor import _anndata_n_observations
+from ..core.storage._backed_access import (
+    _track_writes_factory,
+    backed_access,
+)
 from ..core.storage._pyarrow_dataset import PYARROW_SUFFIXES
 from ..core.storage._tiledbsoma import _soma_n_observations
 from ..core.storage.paths import (
@@ -2212,10 +2216,6 @@ class Artifact(Record, IsVersioned, TracksRun, TracksUpdates):
             )
 
         from lamindb import settings
-        from lamindb.core.storage._backed_access import (
-            _track_writes_factory,
-            backed_access,
-        )
 
         using_key = settings._using_key
         filepath, cache_key = filepath_cache_key_from_artifact(
--- a/lamindb/models/artifact.py
+++ b/lamindb/models/artifact.py
@@ -2162,7 +2162,7 @@ class Artifact(Record, IsVersioned, TracksRun, TracksUpdates):
         | PyArrowDataset
         | Iterator[PolarsLazyFrame]
     ):
-        """Return a cloud-backed data object.
+        """Open a dataset for streaming.
 
         Works for `AnnData` (`.h5ad` and `.zarr`), generic `hdf5` and `zarr`,
         `tiledbsoma` objects (`.tiledbsoma`), `pyarrow` or `polars` compatible formats.
@@ -2171,13 +2171,15 @@ class Artifact(Record, IsVersioned, TracksRun, TracksUpdates):
             mode: can only be `"w"` (write mode) for `tiledbsoma` stores,
                 otherwise should be always `"r"` (read-only mode).
             df_engine: Which module to use for lazy loading of a dataframe
-                from `pyarrow` or `polars` compatible formats (`.parquet` etc).
+                from `pyarrow` or `polars` compatible formats (`.parquet`, `.csv`, `.ipc` etc).
+                This has no effect if the artifact is not a dataframe, i.e.
+                if it is an `AnnData,` `hdf5`, `zarr` or `tiledbsoma` object.
             is_run_input: Whether to track this artifact as run input.
             **kwargs: Keyword arguments for the accessor, i.e. `h5py` or `zarr` connection,
-                `pyarrow.dataset.dataset`.
+                `pyarrow.dataset.dataset`, `polars.scan_*` function.
 
         Notes:
-            For more info, see tutorial: :doc:`/arrays`.
+            For more info, see guide: :doc:`/arrays`.
 
         Example::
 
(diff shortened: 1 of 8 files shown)


falexwolf:
Can we merge this now? I'd run the final integration tests and then make the release. ☺️

Koncopd:
No, not yet.

Koncopd:
Ok, this is ready.

lamindb/models/artifact.py
```diff
@@ -2142,23 +2148,30 @@ def replace(
...
-        AnnDataAccessor,
-        BackedAccessor,
-        SOMACollection,
-        SOMAExperiment,
-        SOMAMeasurement,
-        PyArrowDataset,
-    ]:
+        self,
+        mode: str = "r",
+        df_engine: Literal["pyarrow", "polars"] = "pyarrow",
```
falexwolf, line 2153: Can we call this `engine` or `backend` or another simpler word?

Even if we have "non-df engines" in the future (whatever they will be), I don't think we want to confront the user with a list of prefixed `_engine` arguments.

Collections will in all cases in which `open()` makes sense be homogeneous and we'll never have a need to mix `df_engine` with other parameters.

lamindb/models/artifact.py
```diff
@@ -2142,23 +2148,30 @@ def replace(
...
         """Return a cloud-backed data object.
 
         Works for `AnnData` (`.h5ad` and `.zarr`), generic `hdf5` and `zarr`,
-        `tiledbsoma` objects (`.tiledbsoma`), `pyarrow` compatible formats.
+        `tiledbsoma` objects (`.tiledbsoma`), `pyarrow` or `polars` compatible formats.
 
         Args:
             mode: can only be `"w"` (write mode) for `tiledbsoma` stores,
                 otherwise should be always `"r"` (read-only mode).
+            df_engine: Which module to use for lazy loading of a dataframe
```
falexwolf, line 2173: Also rename the parameter here, please.

lamindb/models/collection.py
```diff
@@ -342,10 +345,21 @@ def append(self, artifact: Artifact, run: Run | None = None) -> Collection:
...
+    def open(
+        self,
+        df_engine: Literal["pyarrow", "polars"] = "pyarrow",
+        is_run_input: bool | None = None,
+        **kwargs,
+    ) -> PyArrowDataset | Iterator[PolarsLazyFrame]:
         """Return a cloud-backed pyarrow Dataset.
 
-        Works for `pyarrow` compatible formats.
+        Works for `pyarrow` and `polars` compatible formats.
```
falexwolf, line 356: Can you spell this out? Does this mean "a collection of parquet files"?

lamindb/models/artifact.py
falexwolf: I'd make this `engine: Literal["pyarrow", "polars"] | None = None` and then select `pyarrow` by default if it's not passed and it's a `DataFrame`.

It'd also be really nice to add a 2nd example, so that people see this doesn't just work for `AnnData`.

image

falexwolf: Also should be called "see guide":
image

falexwolf:
Why does this return an iterable even for a single artifact? 🤔 

image

lamindb/models/collection.py
Koncopd: I would prefer to keep it like this because there are a lot of formats, and they are a bit different between `pyarrow` and `polars`. It is not supported, an error is thrown describing the required suffixes.

lamindb/models/artifact.py
Koncopd: Edited, added an example.
I actually hate such defaults with `None`, you have to read the description etc to understand what is actually the default value, i believe it is much more obvious with just setting it directly.

Koncopd:
>Why does this return an iterable even for a single artifact? 🤔

This (`Iterator[LazyFrame]`) seems to be the proper way to annotate a context manager returning `LazyFrame`.

lamindb/models/artifact.py
falexwolf: But you can't if it's called df_engine, because it might then default to something else for another data format.

lamindb/models/collection.py
falexwolf: I'm just asking to give some examples for what "compatible formats" are, e.g. mention a collection of parquet files because this is what people know.

lamindb/models/artifact.py
Koncopd: Renamed.

lamindb/models/collection.py
Koncopd: done

lamindb/models/artifact.py
falexwolf: We might be introducing a `DataFrameCollection` and a `ZarrCollection` or things like this in the future.

And then the engine for dataframes could be pyarrow or polars and for zarr it could be kerchunk.

Of course I can't predict specifics etc. but `df_engine` would to me appear like a "patch" where "we get something to work for dataframes" rather than sticking with the original vision to create a cohesive abstraction for a bunch of formats.

falexwolf: And we might eliminate the `df` abbreviation from the API at some point in favor of `to_pandas`, `pandas`, `polars`, `parquet`, `pyarrow`, etc. -- sometimes one might call this a "table" (e.g. in pyarrow there are no "dataframes" I think).

falexwolf (approved):

Issue #2735, opened 2025-05-02
Schema docs unclear


Zethson:
![Image](https://github.com/user-attachments/assets/a1077538-f067-4636-8019-96ee04ddc7e1)

I think that `Schema` is one of our most difficult features as it brings a lot of things together. I was looking for some information but struggled reading the API docs and finding what I was looking for. A few random issues that I noticed that I want to tackle.

1. Let's group all `x_type` below all each other.
2. `type – Schema | None = None A type.` -> what does this even mean?
3. `is_type – bool = False Distinguish types from instances of the type.` -> an example does wonders here
4. `index – Feature | None = None A [Feature](https://ec440622.lamindb.pages.dev/lamindb.feature#lamindb.Feature) record to validate an index of a DataFrame.` -> This reads as if this feature cannot be used for curators of other types.
5. `slots – dict[str, Schema] | None = None A dictionary mapping slot names to [Schema](https://ec440622.lamindb.pages.dev/lamindb.schema#lamindb.Schema) objects.` -> The type is already in the typehint. DRY
6. `dtype – str | None = None The simple type. Defaults to None for sets of [Feature](https://ec440622.lamindb.pages.dev/lamindb.feature#lamindb.Feature) records. Otherwise defaults to "num" (e.g., for sets of [Gene](https://ec440622.lamindb.pages.dev/bionty.gene#bionty.Gene)).` -> What is a simple type? Let's be more precise.

PR #2742, opened 2025-05-05
🐛 Fix curating DataFrame index & 🚸 Automatically pass validation of ensembl gene IDs  & ✨ add curator __repr__

Zethson:
1. Fixes the curation of DataFrame indices and ensures correct behavior with a new test.

```python
ensembl_ids = [
    "ENSG00000139618", # BRCA2
    "ENSG00009000000",
]

df = pd.DataFrame(index=pd.Index(ensembl_ids, name="ensembl"))
df

schema = ln.Schema(
    index=ln.Feature(name="ensembl", dtype=bt.Gene.ensembl_gene_id).save()
).save()

curator = ln.curators.DataFrameCurator(df, schema)
```

led to

```
UnboundLocalError                         Traceback (most recent call last)
Cell In[19], line 1
----> 1 curator = ln.curators.DataFrameCurator(df, schema)
File ~/code/lamindb/lamindb/curators/core.py:430, in DataFrameCurator.init(self, dataset, schema, slot)
    425             categoricals.append(feature)
    426 if schema._index_feature_uid is not None:
    427     # in almost no case, an index should have a pandas.CategoricalDtype in a DataFrame
    428     # so, we're typing it as str here
    429     index = pandera.Index(
--> 430         schema.index.dtype if not feature.dtype.startswith("cat") else str
    431     )
    432 else:
    433     index = None
UnboundLocalError: cannot access local variable 'feature' where it is not associated with a value
```

2. Currently, we consider additional columns in `obs` DataFrames that are not a part of the `minimal_set` as just additional columns that we do not validate and simply ignore. Analogously to this behavior, we now consider additional features in `var.T` such as (for now) ensembl gene IDs as fine and do not raise a ValidationError. This is an easy improvement for cases where additional gene IDs are found because of different ensembl versions.

| Before | After |
|--------|-------|
| image | image |

With `maximal_set=True`, additional ensembl IDs are not allowed and it will raise an error accordingly.

image

3. Introduces new `__repr__` for `CatCurate` and more importantly, `Curator`

| Before | After |
|--------|-------|
| image | image |

4. Improves the Schema docstrings

Fixes https://github.com/laminlabs/lamindb/issues/2735


lamindb/core/datasets/mini_immuno.py
```diff
@@ -64,7 +64,9 @@ def get_dataset1(
     with_cell_type_synonym: bool = False,
     with_cell_type_typo: bool = False,
     with_gene_typo: bool = False,
+    with_outdated_gene: bool = False,
```
falexwolf, line 67: Thank you! Great!

lamindb/core/datasets/mini_immuno.py
```diff
@@ -64,7 +64,9 @@ def get_dataset1(
...
     with_wrong_subtype: bool = False,
+    with_index_type_mismatch: bool = False,
```
falexwolf, line 69: Thank you!

lamindb/core/datasets/mini_immuno.py
```diff
@@ -107,7 +113,12 @@ def get_dataset1(
         "study_note": "We had a great time performing this study and the results look compelling.",
     }
     # the dataset as DataFrame
-    dataset_df = pd.DataFrame(dataset_dict, index=["sample1", "sample2", "sample3"])
+    dataset_df = pd.DataFrame(
+        dataset_dict,
+        index=["sample1", "sample2", 0]  # type: ignore
```
falexwolf, line 118: Great catch!

lamindb/curators/core.py
```diff
@@ -205,6 +205,69 @@ def save_artifact(
         # constructor signature
         pass  # pragma: no cover
 
+    def __repr__(self) -> str:
+        from lamin_utils import colors
```
falexwolf, line 209: The whole thing isn't called anywhere. So, need a test.

lamindb/curators/core.py
```diff
@@ -205,6 +205,69 @@ def save_artifact(
...
+
+        if self._schema is not None:
+            # Schema might have different attributes
+            if hasattr(self._schema, "name") and self._schema.name:
```
falexwolf, line 213: Better is the below, I believe, I just saw this in code from @Koncopd -- @sunnyosun & I have always done this the wrong way.

```python
if getattr(self._schema, "name", False):
```

lamindb/curators/core.py
```diff
@@ -427,7 +490,9 @@ def __init__(
                 # in almost no case, an index should have a pandas.CategoricalDtype in a DataFrame
                 # so, we're typing it as `str` here
                 index = pandera.Index(
-                    schema.index.dtype if not feature.dtype.startswith("cat") else str
```
falexwolf, line 430: Oh, this was the bug, I guess. And there was no test for an index with a categorical dtype. Got it.

lamindb/curators/core.py
```diff
@@ -803,6 +870,7 @@ def __init__(
         feature: Feature | None = None,
         cat_manager: DataFrameCatManager | None = None,
         subtype_str: str = "",
+        maximal_set: bool = False,
```
falexwolf, line 873: A comment would be good here because it's subtly different from what it means on `Schema`. But because the arg is so generically named it's still a good name also here.

lamindb/curators/core.py
```diff
@@ -843,7 +912,18 @@ def values(self, new_values):
     @property
     def is_validated(self) -> bool:
         """Return whether the column is validated."""
```
falexwolf, line 914: There are all these occurances of the word `column` which should be replaced with `vector` because we're using this for any vector, not just columns of DataFrame (e.g. `.index` and `.columns` and presumably soon other things).

lamindb/curators/core.py
```diff
@@ -843,7 +912,18 @@ def values(self, new_values):
...
-        return len(self._non_validated) == 0
+        # ensembl gene IDs pass even if they were not validated
+        # this is a simple solution to the ensembl gene version problem
+        if self._field.field.attname == "ensembl_gene_id":
+            # if none of the ensembl gene ids were validated, we are probably not looking at ensembl gene IDs
+            if len(self.values) == len(self._non_validated):
```
falexwolf, line 919: I think we should rather make this the general behavior than only having it for ensembl_gene_id.

If there is not a single valid category, likely the dtype is wrong. We could introduce a setting `n_minimal_valid_values=1` or something like this on `CatVector` with _rough_ analogy to `minimal_set` in `Schema` (where we require defined features to be present, not just any valid values).

For now, we don't need to introduce this setting explicitly, but we could simply assume it. I think that if you validate a categorical and there is not a single valid thing in there, you probably have the wrong dtype.

Of course there will be edge cases with this but doing something only for Ensembl Gene IDs will also produce edge cases.

lamindb/models/schema.py
```diff
@@ -245,36 +245,31 @@ def add(self, features: Feature | list[Feature]) -> None:
...
 
 class Schema(Record, CanCurate, TracksRun):
-    """Schemas of datasets such as the set of columns of a `DataFrame`.
+    """A schema is a feature set, such as the set of columns of a `DataFrame`.
 
-    A simple schema is a feature set such as the set of columns of a `DataFrame`.
-
-    A composite schema has multiple slots, e.g., for an `AnnData`, one schema for slot `obs` and another one for `var`.
-
-    A schema can also merely define abstract constraints or instructions for dataset validation & annotation.
```
falexwolf, line 254: Why remove this last one (I get why you remove the first sentence).

lamindb/models/schema.py
```diff
@@ -372,13 +367,6 @@ class Meta(Record.Meta, TracksRun.Meta, TracksUpdates.Meta):
     """A description."""
     n: int = IntegerField()
     """Number of features in the schema."""
-    itype: str | None = CharField(
-        max_length=120, db_index=True, null=True, editable=False
-    )
-    """A registry that stores feature identifiers used in this schema, e.g., `'Feature'` or `'bionty.Gene'`.
```
falexwolf, line 378: Can we call out "feature identifier type"? So that people understand why it's called `itype`?

lamindb/models/schema.py
Zethson: I was thinking about this several times. I ended up following my initial impression which was: This is a difficult to understand sentence and doesn't have enough practical value.

If you think that the message is important, I am happy to try to phrase it more simply so that it's more easily understood.

lamindb/models/schema.py
```diff
@@ -245,36 +245,31 @@ def add(self, features: Feature | list[Feature]) -> None:
 
 
 class Schema(Record, CanCurate, TracksRun):
-    """Schemas of datasets such as the set of columns of a `DataFrame`.
+    """A schema is a feature set, such as the set of columns of a `DataFrame`.
```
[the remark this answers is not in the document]
Zethson: Sorry, I omitted something that I should have kept. Rephrased!

tests/curators/test_curators_examples.py
```diff
@@ -214,9 +215,21 @@ def test_dataframe_curator(small_dataset1_schema: ln.Schema, ccaplog):
...
+
     artifact.delete(permanent=True)
 
 
+def test_dataframe_curator_index():
+    """Test validating a DataFrame index."""
+    df = datasets.small_dataset1(otype="DataFrame", with_index_type_mismatch=True)
+    schema = ln.Schema(index=ln.Feature(name="test", dtype="str").save()).save()
+    curator = ln.curators.DataFrameCurator(df, schema)
+    try:
```
falexwolf, line 227: Sorry, was dragged into another meeting before.

We need `pytest.raises()` here so that we're sure the error is actually raised.

tests/curators/test_curators_examples.py
```diff
@@ -214,9 +215,21 @@ def test_dataframe_curator(small_dataset1_schema: ln.Schema, ccaplog):
...
     curator.standardize()
     curator.validate()
+
     artifact.delete(permanent=True)
 
 
+def test_dataframe_curator_index():
+    """Test validating a DataFrame index."""
+    df = datasets.small_dataset1(otype="DataFrame", with_index_type_mismatch=True)
+    schema = ln.Schema(index=ln.Feature(name="test", dtype="str").save()).save()
```
falexwolf, line 225: Please delete the schema and the feature at the end of this test so that it doesn't interact with other tests.

tests/curators/test_curators_examples.py
```diff
@@ -270,6 +283,88 @@ def test_dataframe_curator_validate_all_annotate_cat2(small_dataset1_schema):
...
+                "ENSG00000139618",  # BRCA2
+                "ENSG00000141510",  # TP53
+                "ENSG00999000001",  # Invalid ID
+                "ENSG00999000002",  # Invalid ID
+            ],
+            name="ensembl",
+        )
+    )
+    schema = ln.Schema(
+        index=ln.Feature(name="ensembl", dtype=bt.Gene.ensembl_gene_id).save()
```
falexwolf, line 299: Please also delete the feature here at the end of the test.

tests/curators/test_curators_general.py
```diff
@@ -235,3 +239,40 @@ def test_schema_ordered_set(df):
...
+def test_schema_no_minimal_set_var():
+    """If minimal_set is False, invalid ensembl gene IDs are allowed."""
+    adata = ln.core.datasets.mini_immuno.get_dataset1(otype="AnnData")
+    adata.var_names = [adata.var_names[0], adata.var_names[1], "NOT_VALID_ENSEMBL"]
+
+    var_schema = ln.Schema(
+        itype=bt.Gene.ensembl_gene_id,
+        minimal_set=False,
+    ).save()
+    schema = ln.Schema(otype="AnnData", slots={"var": var_schema}).save()
```
falexwolf, line 253: This should be `var.T`, shouldn't it? I think this is the default behavior we typically want, don't we? So that people don't have to deal with messing with different Ensembl versions.

Can you add `minimal_set=True` to `ln.schemas.anndata....`?

tests/curators/test_curators_general.py
```diff
@@ -235,3 +239,40 @@ def test_schema_ordered_set(df):
     # clean up
     ln.Schema.filter().delete()
     ln.Feature.filter().delete()
+
+
+def test_schema_no_minimal_set_var():
+    """If minimal_set is False, invalid ensembl gene IDs are allowed."""
```
falexwolf, line 245: Wait, now I'm confused. Why does this test refer to to minimal set and not to maximal set?

falexwolf (approved):

PR #2751, opened 2025-05-08
🐛 Fix SpatialData in-memory repr

Zethson:
https://laminlabs.slack.com/archives/C04FPE8V01W/p1746691319386589

SpatialData objects zarr stores are moved when saved. However, SpatialData's `__repr__` method attempts to access information from the old path. Therefore, we need to update the in-memory path to the now moved Artifact storage path.

### DIFF
--- a/tests/core/test_load.py
+++ b/tests/core/test_load.py
@@ -135,6 +135,17 @@ def test_load_spatialdata(get_small_sdata):
     assert artifact._local_filepath.resolve() == artifact.cache() == artifact.path
 
 
+def load_blobs__repr__():
+    example_blobs_sdata = ln.core.datasets.spatialdata_blobs()
+    blobs_af = ln.Artifact.from_spatialdata(
+        example_blobs_sdata, key="example_blobs.zarr"
+    ).save()
+    example_blobs_sdata = ln.Artifact.get(key="example_blobs.zarr")
+    example_blobs_sdata = blobs_af.load()
+    # Must exist and not throw errors
+    assert example_blobs_sdata.__repr__
+
+
 def test_load_html(html_filepath):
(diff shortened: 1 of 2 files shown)


Koncopd (changes requested):

Koncopd (approved):
Does it work now?

tests/core/test_load.py
```diff
@@ -135,6 +135,17 @@ def test_load_spatialdata(get_small_sdata):
     assert artifact._local_filepath.resolve() == artifact.cache() == artifact.path
 
 
+def load_blobs__repr__():
+    example_blobs_sdata = ln.core.datasets.spatialdata_blobs()
+    blobs_af = ln.Artifact.from_spatialdata(
+        example_blobs_sdata, key="example_blobs.zarr"
+    ).save()
+    example_blobs_sdata = ln.Artifact.get(key="example_blobs.zarr")
```
Koncopd, line 143: But this doesn't have `_memory_rep` anymore. So this doesn't actually test the change i think.

tests/core/test_load.py
```diff
@@ -135,6 +135,17 @@ def test_load_spatialdata(get_small_sdata):
     assert artifact._local_filepath.resolve() == artifact.cache() == artifact.path
 
 
+def load_blobs__repr__():
+    example_blobs_sdata = ln.core.datasets.spatialdata_blobs()
+    blobs_af = ln.Artifact.from_spatialdata(
```
Koncopd, line 140: This has `_memory_rep`.

Zethson:
Yes, works now. I fixed the test. When it runs through, I'll merge it. I can't think of unwanted side effects but this is delicate business.

Koncopd:
You need to upgrade the `lamin-cli` sub for the test to pass.

PR #2755, opened 2025-05-09
🚸 Improve `DoesNotExist` error message upon `DBRecord.get()`

Zethson:
Fixes https://github.com/laminlabs/pfizer-lamin-usage/issues/19

| Before | After |
| ------ | ----- |
| Before error message | After error message |


falexwolf (approved):

lamindb/models/query_set.py
```diff
@@ -689,9 +691,9 @@ def first(self) -> Record | None:
             return None
         return self[0]
 
-    def one(self) -> Record:
+    def one(self, does_not_exist_msg: str | None = None) -> Record:
```
falexwolf, line 694: Rather than exposing this message as an argument to the user I suggest you replace the call of `.one()` up there with a call to the non-user facing `one_helper()`.

`.one()` is the user-facing API that we should keep free of internal logic! ☺️

lamindb/models/query_set.py
```diff
@@ -180,13 +180,15 @@ def get(
         return super(QuerySet, qs).get(id=idlike)  # type: ignore
     elif isinstance(idlike, str):
         qs = qs.filter(uid__startswith=idlike)
+        DOESNOTEXIST_MSG = f"no record found with uid '{idlike}'. Did you mean {registry.__name__}.get(name='{idlike}')?"
```
falexwolf, line 183: We and others always capitalize error messages (as opposed to logging messages). So it should start with "No record..." ☺️

falexwolf: The problem with the hint part is that it'll be confusing for transform, artifact and collection, which all don't have `name` but `key` instead; `Gene` has `symbol` instead.

You can check for whether the registry `hasattr(registry, "_name_field")` and then use this to write a hint.

I would also say "Did you maybe forget to add a keyword like ..." because of course the alternative is that the idlike argument is simply invalid because it has a typo or was deleted or the user is connected to the wrong instance (which happens a bunch).

So, we should make the message a little bit weaker so that users don't get the impression that adding a keyword is the only way out.

falexwolf (changes requested):

Zethson:
Thank you for the useful review comments @falexwolf ! They should now be addressed.

The error message reads slightly weird now but I didn't want to make it longer. I think it's okay.

falexwolf (changes requested):
Thanks for addressing most comments! ☺️ 🙏 

And sorry for being nitpicky.

Given the whole PR is about perfecting the UX and we don't ever want to touch this message again, I'll say that the grammar/logic in the new message is incorrect.

image

You could say "a keyword as in", but you can't say "like", because the expression that you print is _not_ a keyword. The `keyword` (or `field`) here is `name`.

You ignored one other comment I had: Probably the most frequent query is `ln.Artifact.get()` and it's not OK if users get a non-sensical message in case of an error. It has to show `ln.Artifact.get(key="...")` in that case, not `ln.Artifact.get(name="...")`. Please use the `_name_field` and then you'll have it behave correctly for all of Artifact, Transform, Collection, `bt.Gene` and other entities that don't have a `name` field. It's one line of code more.

falexwolf:
> You ignored one other comment I had: Probably the most frequent query is ln.Artifact.get() and it's not OK if users get a non-sensical message in case of an error. It has to show ln.Artifact.get(key="...") in that case, not ln.Artifact.get(name="..."). Please use the _name_field and then you'll have it behave correctly for all of Artifact, Transform, Collection, bt.Gene and other entities that don't have a name field. It's one line of code more.

Sorry! I now see that you in fact accounted for it. I was blind. 🙈

falexwolf (approved):

Issue #2739, opened 2025-05-03
Better error when trying to annotate with unsaved records


falexwolf:
https://laminlabs.slack.com/archives/C07RD6EB936/p1746238098519169

falexwolf:
@Zethson, I'm assigning you here because you've already _killed_ a few Django errors of death and you kinda postulated as your mission, which I think is great! FYI @sunnyosun

Issue #2372, opened 2025-01-23
Improved error message when attempting to label an unsaved Record such as an`Artifact`


Zethson:
### Add a description

A user ran into:

```
Traceback (most recent call last):
 File "/fs/gpfs41/lv03/fileset01/pool/pool-mann-maedler-shared/niklas_workspace/lamin_xenium/xenium_to_sdata.py", line 41, in <module>
  art.ulabels.add(technology_xenium)
 File "/fs/home/schmacke/miniforge3/envs/sdata/lib/python3.10/site-packages/django/db/models/fields/related_descriptors.py", line 656, in __get__
  return self.related_manager_cls(instance)
 File "/fs/home/schmacke/miniforge3/envs/sdata/lib/python3.10/site-packages/django/db/models/fields/related_descriptors.py", line 1076, in __init__
  raise ValueError(
ValueError: "Artifact(uid='6sYRmHuqlMUvllcS0000', is_latest=True, description='Ovarian Cancer Xenium Dataset', suffix='.sdata', kind='dataset', size=28462729153, hash='XKMmMiOv9dhZD_v8Q1yp1A', n_files=3403, space_id=1, storage_id=1, run_id=13, created_by_id=5, created_at=<django.db.models.expressions.DatabaseDefault object at 0x14f6481820b0>)" needs to have a value for field "id" before this many-to-many relationship can be used.
```

and they did not understand this error at all. It's cryptic. We need a better error message that suggest that the Record that is being labeled (an `Artifact` here but could be anything) needs to be saved first.

Zethson:
```python
from django.db.models.fields.related_descriptors import ManyToManyDescriptor
from django.db.models.query import QuerySet

class EnhancedM2MDescriptor(ManyToManyDescriptor):
    def __get__(self, instance, cls=None):
        if instance is not None and not instance.pk:
            raise ValueError(
                f"Cannot access {self.field.name} on unsaved {instance.__class__.__name__}. "
                f"Save {instance.__class__.__name__} first."
            )
        
        manager = super().__get__(instance, cls)
        
        original_bulk_create = manager.bulk_create
        def enhanced_bulk_create(objs, **kwargs):
            if isinstance(objs, (list, QuerySet)):
                unsaved = [obj for obj in objs if not obj.pk]
                if unsaved:
                    raise ValueError(
                        f"Cannot bulk create relationship - these objects need to be saved first: "
[...]

PR #2756, opened 2025-05-09
🚸 Improve setting relationships of unsaved records UX

Zethson:
Fixes https://github.com/laminlabs/lamindb/issues/2739 & https://github.com/laminlabs/lamindb/issues/2372

- We are overwriting how the django `ManyToManyDescriptor.__get__` behaves which now checks whether the instance whose relationships are supposed to be modified has been saved before. If not, we raise an error.
- The implementation requires such a low level hack because the RelationshipManager is dynamically created. Therefore, simply overwriting `add` & `set` does not do the trick. I tried out several other options that could potentially have less side effects but to no avail.

| Before | After |
| ------ | ----- |
| Before error message | After error message |


lamindb/models/_django.py
```diff
@@ -18,6 +18,30 @@
...
+    will raise ValueError, suggesting explicit .save() of the record to be modified
+    before relationship creation.
+    """
+    from django.db.models.fields.related_descriptors import ManyToManyDescriptor
+
+    original_add = ManyToManyDescriptor.__get__
+
+    def patched_get(self, instance, cls=None):
+        if instance is not None and instance.pk is None:
+            raise ValueError(
```
sunnyosun, line 37: I think it's a bit confusing to assume users are assigning the relationships when they are just accessing them.

Also we need to spell out many-to-many relationships like the original message, because FK relationships don't raise this error:
Screenshot 2025-05-12 at 10 26 30

tests/core/test_models.py
```diff
@@ -87,3 +92,17 @@ def test_registry__repr__artifact():
...
     assert actual_repr.strip() == expected_repr.strip()
+
+
+def test_unsaved_relationship_modification_attempts():
+    af = ln.Artifact.from_df(
+        pd.DataFrame({"col1": [1, 2, 3], "col2": [4, 5, 6]}), description="testme"
+    )
+
+    with pytest.raises(ValueError) as excinfo:
+        af.genes.add(bt.Gene(symbol="MALAT1").save())
```
sunnyosun, line 103: I'd suggest using ULabel here for simplicity.

tests/core/test_models.py
```diff
@@ -87,3 +92,17 @@ def test_registry__repr__artifact():
...
+
+    assert (
+        str(excinfo.value)
+        == "please save the Artifact before adding relationships using '.save()'."
+    )
```
sunnyosun, line 108: Need to delete the gene record here.

sunnyosun:
It would be great to also include the following case, when the related record isn't saved:

```
af = ln.Artifact.from_df(
        pd.DataFrame({"col1": [1, 2, 3], "col2": [4, 5, 6]}), description="testme"
    ).save()
ulabel = ln.ULabel(name="test")

af.ulabels.add(ulabel)
```

sunnyosun:
@falexwolf Do you see any issues with overwriting `ManyToManyDescriptor.__get__`?

lamindb/models/_django.py
```diff
@@ -18,6 +18,30 @@
...
+    Before this patch: Cryptic errors are raised when relationships of an unsaved
+    record are attempted to be modified.
+
+    After this patch: Attempts to access M2M relationships on unsaved objects
+    will raise ValueError, suggesting explicit .save() of the record to be modified
+    before relationship creation.
+    """
+    from django.db.models.fields.related_descriptors import ManyToManyDescriptor
+
+    original_add = ManyToManyDescriptor.__get__
```
falexwolf, line 33: This should be called `original_get`.

falexwolf:
This looks indeed a bit scary just because it touches `.get()` but on the other hand the patch is surgical and seems safe.

So, I think this is a good solution. Great job!

Koncopd:
Are you sure it has to be done through patching and not through overloading something in `QueryManager`?

lamindb/models/_django.py
Koncopd: I agree, the problem is that it is triggered on any access.

Koncopd:
Maybe a different less confusing error? Like "you are trying to access relationship of an unsaved object"?

Zethson:
@Koncopd no I don't think that's possible. At least I don't seem to be able to figure out a more narrowly scoped solution that works and this is also what Claude keeps telling me.

The relevant Mangers are created dynamically using a closure. You can't import them. Django's validation happens during manager initialization, before any mutation methods are called.

The new error message is now:
> You are trying to access the many-to-many relationships of an unsaved Artifact object. Please save it first using '.save()'.

I hope that addresses both of our requests, Sunny & Sergei.

Zethson:
While the performance penalty should be minimal since it's just checking a simple attribute, I can benchmark this if you'd like. 

This implementation is the best I could devise. I also explored alternatives using Signals, but those approaches would likely be even less elegant. I'm open to suggestions for more narrowly scoped and cleaner solutions.

falexwolf:
I don't think it affects performance because it's a mere `if condition` and I don't think `.get()` is called in bulk by Django itself. If it's called many times it's the user who's doing that and then the main latency will be the network, not this if condition, so I think it's fine!

lamindb/models/__init__.py
```diff
@@ -90,3 +91,7 @@
 ArtifactParamValue = ArtifactFeatureValue  # backward compat
 RunParamValue = RunFeatureValue  # backward compat
 Param = Feature  # backward compat
+
+from lamindb.models._django import patch_many_to_many_descriptor
+
+patch_many_to_many_descriptor()
```
falexwolf, line 97: Last nit from me: Why do you re-export this into the `.models` namespace? Why not just call it within `_django` where you define it?

Alternatively, because we're subclassing `QueryManager` in `query_manager.py`, that would also a good place. `.add()` is an attribute of `QueryManager`.

falexwolf:
I like it ☺️

lamindb/models/__init__.py
Zethson: This was a deliberate choice as I thought that developers don't check `_django` but will see this patch in `models`. However, the function should not be in the `models` namespace. Therefore, I'll move it to `query_manager.py`. Thanks!

Zethson:
Could I please get either:
1. An approval.
2. A suggestion for an alternative implementation that I couldn't think of? I'm happy to explore this further but this is still the only working and reasonable solution that I found.

falexwolf (approved):

sunnyosun (approved):

PR #2760, opened 2025-05-11
♻️ Rename `Record` to `DBRecord`

falexwolf:
Towards a flexible record class.

This PR comes down to a bulk name replace across packages because we need to free up the `Record` name for the to-be-defined flexible `Record` class.

- `Record` to `DBRecord`
- `BasicRecord` to `BaseDBRecord`
- `LinkORM` to `IsLink`
- `RecordInfo` to `DBRecordInfo`
- `models.record` to `models.dbrecord`

All of these changes are done in a backward compatible way.

This PR comes with these associated PRs:

- https://github.com/laminlabs/lndocs/pull/123
- https://github.com/laminlabs/lamindb-setup/pull/1030
- https://github.com/laminlabs/bionty/pull/251
- https://github.com/laminlabs/clinicore/pull/20
- https://github.com/laminlabs/wetlab/pull/113
- https://github.com/schultzelab/lnschema-covim/pull/10
- https://github.com/schultzelab/lnschema-swarm/pull/176
- https://github.com/laminlabs/sentseqdb/pull/42

Internal [Notion page](https://www.notion.so/laminlabs/Modeling-samples-and-their-metadata-modeling-flexible-tables-1d82aeaa55e180ec9136dd23163900c4).


Koncopd:
Shouldn't `BasicRecord` be renamed to `BasicDBRecord` then?

lamindb/core/writelog/_trigger_installer.py
```diff
@@ -632,9 +632,9 @@ def install_triggers(self, table: str, cursor: CursorWrapper):
 
 def create_writelog_recording_trigger_installer(
     connection: BaseDatabaseWrapper,
-) -> WriteLogRecordingTriggerInstaller:
+) -> WriteLogDBRecordingTriggerInstaller:
```
Koncopd, line 635: Looks like an incidental rename to me.

falexwolf: Thanks for spotting! I wasn't yet through with reviewing but this is a great catch!

falexwolf:
> Shouldn't BasicRecord be renamed to BasicDBRecord then?

Yes, I thought about this but I don't like `BasicDBRecord` as a name. I thought about `BaseRecord` instead but for the sake of not changing too much at the same time I concluded that it could also be fine to keep `BasicRecord`.

If renaming I would probably go for `BaseDBRecord` 🤔 -- that would convey that it's a base class for `DBRecord` (which in turn will be a base class for the new central flexible `Record` object).

PR #2765, opened 2025-05-13
🚸  Do not validate twice within `Artifact.from_X(...)` when passing schema

falexwolf:


falexwolf:
Do not update this branch to the latest state of main, because I want to make a patch release off of it.

PR #2769, opened 2025-05-14
✨ Add schema-based `TiledbsomaExperimentCurator`

### LINKED ISSUE #2741: Refactor `SomaExperimentCurator` to be schema based
### Add a description

...

### LINKED ISSUE #2777: Pandera FutureWarnings
### Add a description

```
/opt/hostedtoolcache/Python/3.13.3/x64/lib/python3.13/site-packages/pandera/_pandas_deprecated.py:146: FutureWarning: Importing pandas-specific classes and functions from the
top-level pandera module will be **removed in a future version of pandera**.
If you're using pandera to validate pandas objects, we highly recommend updating
your import:

```
# old import
import pandera as pa

# new import
import pandera.pandas as pa
```

If you're using pandera to validate objects from other compatible libraries
like pyspark or polars, see the supported libraries section of the documentation
for more information on how to import pandera:

https://pandera.readthedocs.io/en/stable/supported_libraries.html

To disable this warning, set the environment variable:

```
export DISABLE_PANDERA_IMPORT_WARNING=True
```

  warnings.warn(_future_warning, FutureWarning)
```

We should resolve this as they clutter a lot. Observed here https://867e98ba.lamindb.pages.dev/curate#spatialdata

Zethson:
Fixes #2741 
Fixes #2777 

- Fixes some tracking related test warnings
- Fixes very verbose Pandera import warnings
- Fixes duplicated API docs
- Fixes inconsistencies with script naming. Now all scripts are using underscores. We could have also agreed on dashes but 75%+ were already using underscores
- Refactors the `is_x` functions of the ScverseDataStructures into one `is_scversedatastructure` that can be parametrized with the type to check. This harmonizes the code and makes the behavior more predictable. I understand that this can harm readability and it might not be much better that one now has to pass the name of the expected data structure but we have this behavior across other places of the code and it might ultimately help us.
- Adds lot of tests for checking paths and types of the ScverseDataStructures
- Adds more fixtures and cleans up tiledbsoma related code
- Adds support for `save_artifact` for tiledbsoma for new style Curators
- Adds a new `SomaExperimentCurator` with associated tests and guide


lamindb/models/artifact.py
```diff
@@ -523,45 +523,53 @@ def log_storage_hint(
...
+            if type_suffix in data_path.suffixes:
                 return True
+
             # check only for local, expensive for cloud
             if fsspec.utils.get_protocol(data_path.as_posix()) == "file":
-                return identify_zarr_type(data_path) == "anndata"
+                return (
+                    identify_zarr_type(
+                        data_path if class_name == "AnnData" else data,
+                        check=True if class_name == "AnnData" else False,
```
Zethson: I think a False for all 3 is also okay but I kept the behavior as it was for now.

lamindb/models/artifact.py
```diff
@@ -523,45 +523,59 @@ def log_storage_hint(
...
+        )
+    elif hasattr(data, "__class__") and data.__class__.__name__ == expected_ds:
         return True
+
+    data_type = expected_ds.lower()
     if isinstance(data, (str, Path, UPath)):
         data_path = UPath(data)
-        if ".h5ad" in data_path.suffixes:  # ".h5ad.gz" is a valid suffix
+
+        if file_suffix and f"{file_suffix}" in data_path.suffixes:
```
Koncopd, line 550: Why not `file_suffix in data_path.suffixes`?

lamindb/models/artifact.py
```diff
@@ -1920,12 +1934,13 @@ def from_tiledbsoma(
 
             artifact = ln.Artifact.from_tiledbsoma("s3://mybucket/store.tiledbsoma", description="a tiledbsoma store").save()
         """
-        if UPath(path).suffix != ".tiledbsoma":
+        if not data_is_soma_experiment(exp):
             raise ValueError(
-                "A tiledbsoma store should have .tiledbsoma suffix to be registered."
+                "data has to be a SOMA Experiment object or a path to SOMA Experiment store."
             )
+        exp = exp.uri.removeprefix("file://") if not isinstance(exp, UPathStr) else exp
```
Koncopd, line 1941: What is this? Why is it needed?

lamindb/curators/core.py
```diff
@@ -851,6 +865,92 @@ def __init__(
         self._columns_field = self._var_fields
 
 
+class SomaExperimentCurator(SlotsCurator):
```
Koncopd, line 868: This doesn't have `standardize`?

lamindb/curators/core.py
```diff
@@ -851,6 +865,92 @@ def __init__(
...
+                        )
+                        else schema_dataset
+                    ),
+                    slot_schema,
+                )
+            else:
+                # global Experiment obs slot
+                _ms, modality_slot = None, slot
+                schema_dataset = (
+                    self._dataset.obs.read()
```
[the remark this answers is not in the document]
Koncopd: The point of such curator is surely to avoid loading even `obs` in full.

lamindb/curators/core.py
```diff
@@ -851,6 +865,92 @@ def __init__(
...
+            raise InvalidArgument("dataset must be SOMAExperiment-like.")
+        if schema.otype != "tiledbsoma":
+            raise InvalidArgument("Schema otype must be 'tiledbsoma'.")
+
+        for slot, slot_schema in schema.slots.items():
+            if slot.startswith("ms:"):
+                ms, modality_slot = slot.split(":")
+                schema_dataset = (
+                    self._dataset.ms[modality_slot.removesuffix(".T")]
+                    .var.read()
```
Koncopd, line 901: Same, reading the whole thing, would avoid if possible.

lamindb/models/artifact.py
Zethson: To get the local path, we need to remove this prefix which is always present in `experiment.uri`

lamindb/models/artifact.py
Zethson: Whopsie, not sure what happened here. Will fix.

lamindb/curators/core.py
Zethson: This should come from `SlotsCurator`.

lamindb/curators/core.py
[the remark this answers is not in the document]
Zethson: I understand but then it'll require quite a bit more custom code. Usually `obs` and `var` don't get that big and it's not the biggest issue to load that into memory, is it?

But before I act, I'm waiting for Alex opinion.

lamindb/curators/core.py
Koncopd: Yes, but tiledbsoma requires certain special logic for that.

lamindb/curators/core.py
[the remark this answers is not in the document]
falexwolf: (1) Tell me if I'm wrong: The way people use SOMA typically is by streaming obs rather than fully downloading it. This is also what we show in our `cellxgene` guide for Census.

(2) Given Sergei already spent long hours to build the previous SOMA curator so that it works with the appropriate paradigm, I'm leaning towards not introducing regression re paradigm (1), but adopt existing code to the schema-based implementation. My assumption is that it "can't be so hard given it has already been done once".

(3) Quite generally I believe that many access patterns will be streaming-based in the future, and that rather than considering streaming-compat a patch for SOMA, I'd consider viewing the lack of streaming-compat a deficiency of the DataFrameCurator that should be remedied in the upcoming months (not urgent); hopefully using similar or the same abstraction used for SOMA.

(4) Downloading metadata for 100M rows would amount to `100e6 * 50 * 4 / 1e9 = 20 GB` with a higher estimate for the number of cols and a lower estimate for the byte size per value. So, it's indeed large.

(5) It might be that for the task of curation we always want the full thing. 🤔 So this might make it special.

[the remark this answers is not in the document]
falexwolf: Pondering on this (5) does _not_ hold: we definitely don't want to load 100M string values in most cases; validation only establishes that the column is of type `str`, nothing more. While the latter can be verified from the `pyarrow.lib.Schema` object and takes microseconds once the schema is known, the full load of 100M string values will take seconds if not more.

So, I'd say one should follow the suggestion in (2). Further performance optimizations can be done at a later point. The scope of this PR was to "port the existing implementation to the new-style curators" not "to improve the existing implemenation" (unless that's simple, of course).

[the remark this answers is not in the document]
falexwolf: Pondering this again, if things really do become ugly, one could say that "the curator is only compatible with smaller-scale soma stores". There won't be a frequent need to run it on a gigantic store (Census is one example though). So, it comes down to how hard it is to port what Sergei did without making the code complicated. It might be that new abstractions would need to be introduced, and maybe that's out-of-scope for this PR.

docs/scripts/curate_soma_experiment.py
```diff
@@ -0,0 +1,42 @@
+import lamindb as ln
+import bionty as bt
+import tiledbsoma as soma
+import tiledbsoma.io
+
+adata = ln.core.datasets.small_dataset1(otype="AnnData")
+tiledbsoma.io.from_anndata("curate.tiledbsoma", adata, measurement_name="RNA")
```
falexwolf, line 7: This is a nitpick, but I don't find the choice for the folder name very intuitive. We're typically using `my_dataset.tiledbsoma` or something along these lines. I tried to standardize this with the idea of showcasing that "with lamindb you can store the same dataset in different formats".

docs/scripts/curate_soma_experiment.py
```diff
@@ -0,0 +1,42 @@
...
+soma_schema = ln.Schema(
+    name="soma_experiment_schema",
+    otype="tiledbsoma",
+    slots={
+        "obs": obs_schema,
+        "ms:RNA.T": var_schema,
+    },
+).save()
+
+curator = ln.curators.SomaExperimentCurator(experiment, soma_schema)
```
falexwolf, line 35: In most cases, this would be called on the URI, not on the `Experiment` stream object that you're passing. It also seems like an anti-pattern to just open the stream and not close it within a context manager.

So, I suggest to pass the URI aka folder path (be it local or on S3). I also believe that that's consistent with `ln.integrations.save_tiledbsoma_experiment()`. Tell me if I'm wrong!

Independent, given the name `SomaExperimentCurator` we should adopt the `.from_tiledbsoma()` and `save_tiledbsoma_experiment()` names (backward compat).

Either is's all called `Tiledbsoma` and we allow for additional logic that specifies that we're only looking for the `Experiment` slot (that would need to be an argument). Or we call everything `SomaExperiment`.

If there won't ever be a need to curate non-Experiment SOMA stores, the later is preferrable. Otherwise the former might be preferrable.

falexwolf: Either way naming has to be consistent and the docs should cross-reference these three pieces of API logic.

lamindb/core/storage/_valid_suffixes.py
```diff
@@ -6,6 +6,7 @@
 VALID_COMPOSITE_SUFFIXES.update(
     {
         ".vitessce.json",
+        "spatialdata.zarr",
```
falexwolf, line 9: Given there was official agreement against this suffix from the spatialdata team, we removed it here.

So, unless this changed, I'd keep it removed.

lamindb/core/storage/_zarr.py
```diff
@@ -64,15 +65,13 @@ def identify_zarr_type(
...
-    # unlike ".anndata.zarr" in VALID_SUFFIXES
-    # suffixes = UPath(storepath).suffixes
-    # if ".spatialdata" in suffixes:
-    #     return "spatialdata"
-    # elif ".anndata" in suffixes:
-    #     return "anndata"
+    suffixes = UPath(storepath).suffixes
+    if ".anndata" in suffixes:
+        return "anndata"
+    elif ".mudata" in suffixes:
```
falexwolf, line 70: This isn't a thing, is it?

lamindb/core/storage/_zarr.py
```diff
@@ -64,15 +65,13 @@ def identify_zarr_type(
...
+        return "mudata"
+    elif ".spatialdata" in suffixes:
```
falexwolf, line 72: Mark uses this, so I think it's a valid thing to check for independent of whether lamindb supports it as a legit value for `artifact.suffix` (I think it shouldn't unless spatiadata team supports it).

lamindb/curators/core.py
```diff
@@ -320,31 +320,35 @@ def save_artifact(
...
-                )
-            elif data_is_spatialdata(self._dataset):
-                self._artifact = Artifact.from_spatialdata(
-                    self._dataset,
-                    key=key,
-                    description=description,
-                    revises=revises,
-                    run=run,
-                )
+            type_mapping = [
```
falexwolf, line 325: Pretty!

lamindb/curators/core.py
```diff
@@ -320,31 +320,35 @@ def save_artifact(
...
+                (
+                    lambda data: data_is_scversedatastructure(data, "AnnData"),
```
falexwolf, line 327: Can this be called `dataset` rather than `data` for consistency with how we refer to the this object?

lamindb/curators/core.py
```diff
@@ -461,9 +465,9 @@ def __init__(
                         )
                     else:
                         dtype = None
-                    pandera_columns[feature.name] = pandera.Column(
+                    pandera_columns[feature.name] = pa.Column(
```
falexwolf: Please keep `pandera`. `pa` means `pyarrow`. This was a conscious choice.

lamindb/curators/core.py
```diff
@@ -851,6 +865,92 @@ def __init__(
         self._columns_field = self._var_fields
 
 
+class SomaExperimentCurator(SlotsCurator):
+    """Curator for `TileDB-SOMA`.
```
falexwolf, line 907: Should be called `SomaExperiment`.

falexwolf: Or explicit `tiledbsoma.Experiment`.

lamindb/curators/core.py
```diff
@@ -851,6 +853,88 @@ def __init__(
...
+
+    def __init__(
+        self,
+        dataset: SOMAExperiment | Artifact,
+        schema: Schema,
+    ) -> None:
+        super().__init__(dataset=dataset, schema=schema)
+        if not data_is_soma_experiment(self._dataset):
+            raise InvalidArgument("dataset must be SOMAExperiment-like.")
+        if schema.otype != "tiledbsoma":
```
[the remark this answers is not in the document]
falexwolf: Yes, we should if we generally decide for the `SomaExperiment` naming convention across the API, but this is a data migration that's beyond the scope of this PR. Within this PR for the `otype`, let's stick with `tiledbsoma` because this is what the instances have right now as `otype`.

It also depends on what `artifact.open()` needs to know. If it suffices to just know that it's a `tiledbsoma` store, then it'd be fine to keep naming it `tiledbsoma`. This would also mean that potentially our previous naming convention `Tiledbsoma` would be a good choice overall.

lamindb/curators/core.py
```diff
@@ -851,6 +865,92 @@ def __init__(
...
+            raise InvalidArgument("Schema otype must be 'tiledbsoma'.")
+
+        for slot, slot_schema in schema.slots.items():
+            if slot.startswith("ms:"):
```
falexwolf, line 935: Just double checking that the slot is actually called `ms` and not `measurement`. ☺️

lamindb/curators/core.py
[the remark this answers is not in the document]
falexwolf: Having seen the entire implementation I fear that significant new abstractions would be needed to implement this with streaming. I think that's beyond the scope of this PR even though it somewhat introduces a regression of the previous functionality.

It's more important to have a consolidated codebase in which data structures use the same abstractions so that we can focus on quality in the upcoming weeks and months (maintainability, performance etc.).

lamindb/curators/core.py
```diff
@@ -851,6 +865,92 @@ def __init__(
...
+                self._slots[slot] = DataFrameCurator(
+                    schema_dataset,
+                    slot_schema,
+                )
+
+            if modality_slot == "var" and schema.slots[slot].itype not in {
+                None,
+                "Feature",
+            }:
+                logger.warning(
```
falexwolf, line 976: I wouldn't add the "backward compat" here because there never was a schema-based soma curator. Hence we don't need to support this pattern.

lamindb/curators/core.py
```diff
@@ -851,6 +865,92 @@ def __init__(
...
+                    "auto-transposed `var` for backward compat, please indicate transposition in the schema definition by calling out `.T`: slots={'var.T': itype=bt.Gene.ensembl_gene_id}"
+                )
+
+            _assign_var_fields_categoricals_multimodal(
+                modality=slot,  # not using "ms" here as it would always be the same for all modalities
```
falexwolf, line 981: What does the comment mean?

I still totally don't get what the `_assign_var...` thing does but that's from @sunnyosun and for another PR. 😆

lamindb/models/artifact.py
```diff
@@ -523,45 +523,59 @@ def log_storage_hint(
...
+                return (
+                    identify_zarr_type(
+                        data_path if expected_ds == "AnnData" else data,
+                        check=True if expected_ds == "AnnData" else False,
+                    )
+                    == data_type
+                )
             else:
-                logger.warning("We do not check if cloud zarr is AnnData or not")
+                logger.warning(f"We do not check if cloud zarr is {expected_ds} or not")
```
falexwolf, line 568: Please lower-case the warning message for consistency with other logging messages

lamindb/models/artifact.py
falexwolf: The signature should be consistent with `integrations.save_tiledbsoma_experiment()`. How is it handled there? It should run into the same issue. Background: Are we sure that it's always `file://`?

lamindb/curators/core.py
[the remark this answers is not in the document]
Zethson: I'll have a look and will outline how much work it would be so that we can make an informed decision.

docs/scripts/curate_soma_experiment.py
Zethson: > In most cases, this would be called on the URI, not on the Experiment stream object that you're passing. It also seems like an anti-pattern to just open the stream and not close it within a context manager. So, I suggest to pass the URI aka folder path (be it local or on S3). I also believe that that's consistent with `ln.integrations.save_tiledbsoma_experiment()`. Tell me if I'm wrong!

Generally agreed. Nevertheless, I think we should support both as users might already have an Experiment open. I'll adapt the example though so that this pattern is clearer.

I'll think about the naming...

falexwolf:
My comments still hold, but none of them are a major deal. I'm merging now to have this in the release. Please follow-up with another PR.

lamindb/core/storage/_zarr.py
Zethson: You never know what people do but it's for consistencies sake. We also can't cover this properly with a test unless we have that. I am happy to remove it.

Oh, and people do everything.

lamindb/curators/core.py
Zethson: Whereas MuData stores modalities in modality keys ("RNA:", "ATAC:", ...), SomeExperiment stores all of them under "measurement". To reuse the same code, we don't pass "measurement" here because it's basically a constant but the slot which resolves to "RNA", "ATAC", ...

lamindb/models/artifact.py
Zethson: In the case of `save_tiledbsoma_experiment` we are constructing the path ourselves so there is no `file://` prefix.

PR #2771, opened 2025-05-15
✨ Add `is_run_input` to `Artifact.get()` and `Collection.get()`

Koncopd:
Allows to track artifacts and collection by passing `is_run_input` to `Artifact.get()` and `Collection.get()`

Examples:
```python
ln.Artifact.get(key="adata.h5ad", is_run_input=True)
ln.Collection.get(key="mycollection", is_run_input=True)
```


lamindb/models/artifact.py
```diff
@@ -2706,6 +2709,9 @@ def _track_run_input(
     is_run_input: bool | Run | None = None,
     run: Run | None = None,
 ):
+    if is_run_input is False:
+        return
```
Koncopd, line 2713: @falexwolf I think it should just return if `False` instead of pulling `run` etc. Am i correct?

falexwolf: I think so, yes!

lamindb/models/artifact.py
```diff
@@ -1518,12 +1518,15 @@ def _cache_path(self) -> UPath:
     def get(
         cls,
         idlike: int | str | None = None,
+        *,
+        is_run_input: bool | None = False,
```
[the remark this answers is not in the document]
Koncopd: I have changed `None` to `Run`. I think this is correct as we allow passing runs there.

falexwolf (approved):

Issue #2649, opened 2025-04-09
Improve error message when path suffix and key do not match


Zethson:
### Add a description

```python
ln.Artifact("noxfile.py",
            key="code/noxfile").save()
```

leads to

```
---------------------------------------------------------------------------
InvalidArgument                           Traceback (most recent call last)
Cell In[3], line 1
----> 1 ln.Artifact("noxfile.py",
      2             key="code/noxfile").save()

File ~/code/lamindb/lamindb/models/artifact.py:1364, in Artifact.__init__(self, *args, **kwargs)
   1362     is_automanaged_path = False
   1363 provisional_uid, revises = create_uid(revises=revises, version=version)
-> 1364 kwargs_or_artifact, privates = get_artifact_kwargs_from_data(
   1365     data=data,
   1366     key=key,
   1367     run=run,
   1368     format=format,
   1369     provisional_uid=provisional_uid,
   1370     version=version,
   1371     default_storage=default_storage,
   1372     using_key=using_key,
   1373     skip_check_exists=skip_check_exists,
   1374 )
   1376 # an object with the same hash already exists
   1377 if isinstance(kwargs_or_artifact, Artifact):

File ~/code/lamindb/lamindb/models/artifact.py:383, in get_artifact_kwargs_from_data(data, key, run, format, provisional_uid, version, default_storage, using_key, is_replace, skip_check_exists)
    380 from lamindb import settings
    382 run = get_run(run)
--> 383 memory_rep, path, suffix, storage, use_existing_storage_key = process_data(
    384     provisional_uid,
    385     data,
    386     format,
    387     key,
    388     default_storage,
    389     using_key,
    390     skip_check_exists,
    391     is_replace=is_replace,
    392 )
    393 stat_or_artifact = get_stat_or_artifact(
    394     path=path,
    395     key=key,
    396     instance=using_key,
    397     is_replace=is_replace,
    398 )
    399 if isinstance(stat_or_artifact, Artifact):

File ~/code/lamindb/lamindb/models/artifact.py:260, in process_data(provisional_uid, data, format, key, default_storage, using_key, skip_existence_check, is_replace)
    258     else:
    259         message = f"The suffix '{key_suffix}' of the provided key is inconsistent, it should be '{suffix}'"
--> 260     raise InvalidArgument(message)
    262 # in case we have an in-memory representation, we need to write it to disk
    263 if memory_rep is not None:

[...]

PR #2780, opened 2025-05-20
🚸 Improve suffix mismatch error message

Zethson:
fixes https://github.com/laminlabs/lamindb/issues/2649

Before

After

### DIFF
--- a/lamindb/models/artifact.py
+++ b/lamindb/models/artifact.py
@@ -261,9 +261,9 @@ def process_data(
     if key_suffix is not None and key_suffix != suffix and not is_replace:
         # consciously omitting a trailing period
         if isinstance(data, (str, Path, UPath)):  # UPathStr, spelled out
-            message = f"The suffix '{suffix}' of the provided path is inconsistent, it should be '{key_suffix}'"
+            message = f"Mismatch between data path suffix '{suffix}' and key suffix '{key_suffix}'."
         else:
-            message = f"The suffix '{key_suffix}' of the provided key is inconsistent, it should be '{suffix}'"
+            message = f"Mismatch between key suffix '{key_suffix}' and data suffix '{suffix}'."
         raise InvalidArgument(message)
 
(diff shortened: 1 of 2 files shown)


lamindb/models/artifact.py
```diff
@@ -261,9 +261,9 @@ def process_data(
     if key_suffix is not None and key_suffix != suffix and not is_replace:
         # consciously omitting a trailing period
         if isinstance(data, (str, Path, UPath)):  # UPathStr, spelled out
-            message = f"The suffix '{suffix}' of the provided path is inconsistent, it should be '{key_suffix}'"
+            message = f"Mismatch between data path suffix '{suffix}' and key suffix '{key_suffix}'."
```
falexwolf, line 264: I think this is less easy to understand than the previous message both because it doesn't tell the user how to resolve and because it uses a non-established concept "data path".

How about the following:
```
f"The suffix '{suffix}' of the passed path is inconsistent with what it should be according to the target key: '{key_suffix}'"
```

Zethson: That's a bit verbose. I am also always in favor of suggesting a solution but here I thought the resolution would be obvious enough.

How about

`f"The suffix '{suffix}' must match target key suffix '{key_suffix}'"`

?

falexwolf: Yes your suggestion is better up to that it should be `match the target key` instead of `match target key`, I think. For consistency one could use `passed key` rather than `target key`, because some people might be confused with what "target" means. It's not consistently referred to this way.

falexwolf:
I believe there is an apostrophe missing without which it's hard to understand in all occurrences of:

the passed path's suffix
the passed key's suffix

Zethson:
I had actually removed them deliberately as I had the impression that people sometimes omit them in errors to keep them simpler. However, I am adding them now for correctness.

falexwolf (approved):

PR #2783, opened 2025-05-21
✨ Add `WriteLog` replay

alexras:
This diff implements a "write log replayer" that can apply individual write log records (specified as `WriteLog` model instances) to the current instance's database. As a side-effect of this change, I had to make more of the system aware of individual columns' types, and I've moved a couple things shared by both the trigger installer and write log replayer into utility functions.


lamindb/models/writelog.py
```diff
@@ -52,8 +52,7 @@ class WriteLog(models.Model):
     created_by_uid = models.CharField(max_length=8, default=DEFAULT_CREATED_BY_UID)
     branch_code = models.IntegerField(default=DEFAULT_BRANCH_CODE)
     run_uid = models.CharField(max_length=16, default=DEFAULT_RUN_UID)
-    # Many-to-many tables don't have row UIDs, so this needs to be nullable.
-    record_uid = models.JSONField(null=True)
+    record_uid = models.JSONField()
```
falexwolf, line 55: I'm only seeing this now, but I think this should likely be a `CharField` rather than a `JSONField`, should it?

lamindb/core/writelog/_db_metadata_wrapper.py
```diff
@@ -67,14 +70,45 @@ def get_many_to_many_db_tables(self) -> set[str]:
...
                 TableUID(
                     source_table_name=table,
-                    uid_columns=["value", "created_at"],
+                    uid_columns=self._get_columns_by_name(
+                        table, ["value", "created_at"], cursor
+                    ),
+                    key_constraint=None,
+                )
+            ]
+        elif table == "lamindb_param":
```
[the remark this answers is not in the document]
falexwolf: Can be removed in the next PR.

lamindb/models/writelog.py
falexwolf: I tried this in the next PR and it seems there is a good reason for that this is a JSON field. Many tests break if turning it into CHARField. I think this is because you're modeling multiple uids as a list within the JSON field.

https://github.com/laminlabs/lamindb/pull/2798

alexras: The write log record's UID takes one of two forms:

* For tables that are many to many, it's a list of 3-tuples specifying the table's foreign-key relationships by UID
* For all other tables, it's an ordered list of values of the table's UID fields.

Because of that it's easier if those fields are JSONFields rather than CharFields.

PR #2784, opened 2025-05-21
🚸 Clearer error in `parse_cat_dtype` if cat dtype contains a module name and the module is not found

Koncopd:
If dtype is `cat[bionty.CellType]`, for example, and `bionty` is not installed, a very obscure import error appears.

See also https://laminlabs.slack.com/archives/C07H7FQT9DH/p1747150860447589?thread_ts=1747148431.902579&cid=C07H7FQT9DH


falexwolf (approved):
Good!

PR #2789, opened 2025-05-23
🎨 A couple minor additions to write logs

alexras:
When the write log was first established, I neglected to increase the width of its `seqno` field to 64-bits. This diff fixes that problem.

I've also added a test that verifies that write log records aren't written by triggers when the write log lock is locked.


falexwolf (approved):
Great! I also renamed `seqno` to `id` for consistency with the PK name on all other tables. This was another one of the little things we talked about the other day. ☺️

alexras:
> Great! I also renamed seqno to id for consistency with the PK name on all other tables

Thanks!

PR #2796, opened 2025-05-26
✨ Implement a `WriteLog` serde

alexras:
This diff adds a serializer/deserializer for `WriteLog` objects that allows them to be passed between instances as JSON. 

Needs:

- https://github.com/laminlabs/lamindb/pull/2783


falexwolf (approved):

falexwolf:
I had to add one more commit to the other PR because the migration state was inconsistent. This is why it offered to update this PR.

Maybe I can change the base now.

falexwolf:
This is the diff wrt `alexras/write-log-replay`

image

image

When I change the base to `main`, git seems to get confused. So, I'll do this now and then resolve conflicts by inspecting these "actual diffs".

falexwolf:
Ok, conflicts are resolved. The diff is correct.

PR #2798, opened 2025-05-27
♻️ Minor simplifications to `WriteLog` schema

falexwolf:
Minor simplifications: 

- two uid columns are now foreign keys (`Branch` and `Space` can be easily synchronized, `Branch` is needed for RLS which only understands the FK)
- simpler names for two tables
- removed unnecessary and duplicated code

### DIFF
--- a/lamindb/core/writelog/_trigger_installer.py
+++ b/lamindb/core/writelog/_trigger_installer.py
@@ -548,7 +548,7 @@ class PostgresHistoryRecordingFunctionBuilder:
             self.declare_variable(
                 variable_name,
                 "smallint",
-                f"SELECT id FROM lamindb_writelogtablestate WHERE table_name = '{table}' LIMIT 1",  # noqa: S608
+                f"SELECT id FROM lamindb_tablestate WHERE table_name = '{table}' LIMIT 1",  # noqa: S608
             )
 
         return variable_name


lamindb/core/writelog/_db_metadata_wrapper.py
```diff
@@ -99,29 +99,11 @@ def _get_columns_by_name(
...
-            column_list.append(column)
-
-        return column_list
-
     def get_uid_columns(self, table: str, cursor: CursorWrapper) -> UIDColumns:
         """Get the UID columns for a given table."""
         if table == "lamindb_featurevalue":
-            # TODO: update this to feature + hash instead of value + created_at
+            # TODO: update this to feature uid + hash instead of value + created_at
+            # feature_id is a foreign key on the featurevalue table, so this should be easy
```
falexwolf, line 106: This is the one thing that I can't figure out but I hope is simple for you given you solved the much harder case of the ManyToMany a few lines below, @alexras!

Can you please update the `TableUID` object so that it reflects `feature.uid` and `featurevalue.hash`. `lamindb_featurevalue.feature` is an FK to `lamindb_feature`.

lamindb/core/writelog/_db_metadata_wrapper.py
```diff
@@ -131,16 +113,6 @@ def get_uid_columns(self, table: str, cursor: CursorWrapper) -> UIDColumns:
                     key_constraint=None,
                 )
             ]
-        elif table == "lamindb_param":
```
falexwolf, line 134: This is no longer necessary as this table doesn't exist anymore.

falexwolf:
I hope that after this, we won't need schema changes anymore. 😅

alexras (approved):

PR #2799, opened 2025-05-28
♻️ Support combined UID format for `lamindb_featurevalue` in `WriteLog`

alexras:
`lamindb_featurevalue` is unique among our current tables because its UID is composed of both its own fields and the fields of one of its foreign-keyed tables. This requires some special handling both in cursor installation and replay.


falexwolf (approved):
Great!

PR #2803, opened 2025-05-28
🔈 Fix Pandera warning

### LINKED ISSUE #2777: Pandera FutureWarnings

Zethson:
Fixes #2777 

Extracted from https://github.com/laminlabs/lamindb/pull/2769

### DIFF
--- a/lamindb/curators/core.py
+++ b/lamindb/curators/core.py
@@ -20,7 +20,7 @@ from typing import TYPE_CHECKING, Any, Callable
 import lamindb_setup as ln_setup
 import numpy as np
 import pandas as pd
-import pandera
+import pandera.pandas as pa
 from lamin_utils import colors, logger
 from lamindb_setup.core._docs import doc_args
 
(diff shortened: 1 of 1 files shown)


lamindb/curators/core.py
```diff
@@ -20,7 +20,7 @@
 import lamindb_setup as ln_setup
 import numpy as np
 import pandas as pd
-import pandera
+import pandera.pandas as pa
```
falexwolf, line 23: As I pointed out in the other PR, AFAIK, `pa` is typically used for `pyarrow`. We'll merge as is now, but I think we should use something that is less confusing. If `pandera` doesn't work anymore maybe `ppd` or something like this.

falexwolf (approved):

Issue #2788, opened 2025-05-23
Rename warning should maybe only include the uid and not the whole track command


Zethson:
### Add a description

I ran

```python
ln.track(project=tutorial_proj)
```

and was met with

> • recommendation: to identify the notebook across renames, pass the uid: ln.track("4Ng4oyRJwmVg", project="Project(uid='lv6nhPrY2vqo', name='tutorial', is_type=False, space_id=1, created_by_id=1, created_at=2025-05-23 14:56:04 UTC)")

However, I don't think that the full project should be printed here.

Zethson:
@falexwolf what do you think?

Zethson:
Thanks @Koncopd

PR #2805, opened 2025-05-29
🔊 Fix project hint in ln.track()

Koncopd:
Fix https://github.com/laminlabs/lamindb/issues/2788

Now this shows project name if the argument is an instance of `ln.Project`.


Koncopd:
@falexwolf is it fine now?

PR #2816, opened 2025-06-01
🎨 Improve SomaExperimentCurator

Zethson:
Addresses review comments in https://github.com/laminlabs/lamindb/pull/2769

### DIFF
--- a/docs/scripts/curate_soma_experiment.py
+++ b/docs/scripts/curate_soma_experiment.py
@@ -28,7 +28,7 @@ soma_schema = ln.Schema(
     otype="tiledbsoma",
     slots={
         "obs": obs_schema,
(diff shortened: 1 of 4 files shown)


docs/scripts/curate_soma_experiment.py
```diff
@@ -5,7 +5,6 @@
 
 adata = ln.core.datasets.small_dataset1(otype="AnnData")
```
falexwolf: We're using the `ln.core.datasets.mini_immuno.get_dataset1()` accessor now. ☺️ -- The `small_dataset1()` function is for backward compat only.

falexwolf (approved):
Please just use the non-legacy convention for the mini immuno dataset! ☺️

PR #2819, opened 2025-06-02
🚸 Use branch and space from CLI

falexwolf:


tests/permissions/test_permissions.py
```diff
@@ -211,11 +213,27 @@ def test_token_reset():
...
+        capture_output=True,
+    )
+    print(result.stdout.decode())
+    print(result.stderr.decode())
+    assert "key='mytest'" in result.stdout.decode()
+    assert "storage path:" in result.stdout.decode()
+    assert result.returncode == 0
+
+    print(ln.setup.settings.instance.slug)
+    print(ln.Artifact.using("laminlabs/lamin-dev").filter().df())
```
falexwolf: @Koncopd, is there anything in this test setup that would prevent this query from returning anything?

Koncopd: not sure, need to check.

Koncopd: how do you see it? your tests are failing at a different place in the CI.

falexwolf: I see it when running tests locally.

Koncopd: Could you try checking it in the CI l also. `ln.Artifact.using("laminlabs/lamin-dev")` is tested in unit-tests, so it should be something very specific.

falexwolf: Debugging this I find that it works as expected until I start to run this script: https://github.com/laminlabs/lamindb/blob/main/tests/permissions/scripts/check_lamin_dev.py

After having run the script, the `.using()` pattern doesn't yield the result anymore. I keep investigating.

Koncopd: It does  `ln.connect("laminlabs/lamin-dev")`, maybe this is the problem?

falexwolf: Yeah, but it's a different python process, so I thought this would be OK. I'll figure things out.

falexwolf: I understand what's happening now. I will resolve in another PR because it's related to this issue: https://github.com/laminlabs/pfizer-lamin-usage/issues/223

Koncopd: Wha is the problem?

tests/permissions/test_permissions.py
```diff
@@ -211,11 +213,31 @@ def test_token_reset():
...
+    result = subprocess.run(  # noqa: S602
+        "lamin save .gitignore --key mytest --space 'Our test space for CI'",
+        shell=True,
+        capture_output=True,
+    )
+    print(result.stdout.decode())
+    print(result.stderr.decode())
+    assert "key='mytest'" in result.stdout.decode()
+    assert "storage path:" in result.stdout.decode()
+    assert result.returncode == 0
```
Koncopd, line 240: please use `check=True` instead in `subprocess.run`.

falexwolf: But will I still be able to print the logs then? If `check=True` it errors before printing the logs, doesn't it?

falexwolf: I'm consciously not using `check=True` for this reason (tell me if I'm wrong).

tests/permissions/test_permissions.py
falexwolf: Will explain in another PR!

tests/permissions/test_permissions.py
Koncopd: it outputs the traceback i think.

falexwolf: I'll investigate. I've been using this pattern all over the place and can change it everywhere if I get the info that I need to debug also with `check=True`. But I don't want to do it in this PR.

Koncopd: but also fine as it is i guess.

PR #2822, opened 2025-06-05
🏷️ Mark as py.typed to allow mypy inspection

ap--:
Prevents all types to show up as Any when running mypy against scripts using lamindb


Koncopd:
Thank you, @ap-- , i will check a bit later.

falexwolf:
Thanks, @ap--! I added you as a collaborator to the repo so that you don't need to use a fork anymore and tests will run automatically.

Merging!

PR #2823, opened 2025-06-05
🏷️ BaseSQLRecord.save should return self type

ap--:
`BaseSQLRecord.save()` should return Self type.

This PR uses the TypeVar T defined above which is bound by SQLRecord (==Self).

This fixes issues with typing of for example `ULabel(...).save()`


falexwolf:
Thank you, @ap--! Merging.

Issue #2606, opened 2025-03-28
Progress bar too verbose in non-interactive settings


falexwolf:
I've now seen this multiple times. We should think about a different default behavior. It makes logs unreadable.

```
→ connected lamindb: testuser1/test-docs
→ created Transform('hMrsNMEnyoEd0000'), started new Run('UViScVnU...') at 2025-03-28 12:26:46 UTC
... synchronizing 02faf712-92d4-4589-bec7-13105059cf86.h5ad:  0.0%
... synchronizing 02faf712-92d4-4589-bec7-13105059cf86.h5ad:  0.1%
... synchronizing 02faf712-92d4-4589-bec7-13105059cf86.h5ad:  0.2%
... synchronizing 02faf712-92d4-4589-bec7-13105059cf86.h5ad:  0.5%
... synchronizing 02faf712-92d4-4589-bec7-13105059cf86.h5ad:  0.6%
... synchronizing 02faf712-92d4-4589-bec7-13105059cf86.h5ad:  0.9%
... synchronizing 02faf712-92d4-4589-bec7-13105059cf86.h5ad:  1.1%
... synchronizing 02faf712-92d4-4589-bec7-13105059cf86.h5ad:  1.4%
... synchronizing 02faf712-92d4-4589-bec7-13105059cf86.h5ad:  1.6%
... synchronizing 02faf712-92d4-4589-bec7-13105059cf86.h5ad:  2.1%
... synchronizing 02faf712-92d4-4589-bec7-13105059cf86.h5ad:  2.1%
... synchronizing 02faf712-92d4-4589-bec7-13105059cf86.h5ad:  2.5%
... synchronizing 02faf712-92d4-4589-bec7-13105059cf86.h5ad:  3.0%
... synchronizing 02faf712-92d4-4589-bec7-13105059cf86.h5ad:  3.0%
... synchronizing 02faf712-92d4-4589-bec7-13105059cf86.h5ad:  3.2%
... synchronizing 02faf712-92d4-4589-bec7-13105059cf86.h5ad:  3.7%
... synchronizing 02faf712-92d4-4589-bec7-13105059cf86.h5ad:  4.2%
... synchronizing 02faf712-92d4-4589-bec7-13105059cf86.h5ad:  4.3%
... synchronizing 02faf712-92d4-4589-bec7-13105059cf86.h5ad:  4.8%
... synchronizing 02faf712-92d4-4589-bec7-13105059cf86.h5ad:  5.2%
[...]

PR #2824, opened 2025-06-06
🔊 Correct treatment of carriage returns in logs

Koncopd:
Fix https://github.com/laminlabs/lamindb/issues/2606

This PR fixes logging strings with carriage returns, it only writes the last non-empty text segment so that carriage return is obeyed.


falexwolf (approved):
Great!

PR #2825, opened 2025-06-08
🧑‍💻 Support import of model classes before connecting to an instance

ap--:
Hi :wave:

This adds support for importing lamindb model classes before running `ln.connect(...)` via lazy import proxies for all objects that are only available after connect. This plays much nicer with modern Python code style, where all imports are on the top of the file. After the user connects to an instance, instantiating the proxy classes returns instances of the corresponding models and calling classmethods on the proxy delegates to the model class.

### Notes

This needs tests, since I would bet that there are some edge cases where it breaks down. My guess is that auto-completion in jupyter might behave differently, but it needs to be checked.

I also noticed that `lamindb` imports a lot of submodules eagerly once connect happens, and I am not sure why you'd want that. I wonder if this enabled a better experience for users in jupyter notebooks and that's why you did it that way. It would be great to understand the reasoning behind it.

In addition I changed the submodule import behavior in this PR slightly, and made the import of submodules "base", "errors", and "setup" also lazy (the three submodules that are eagerly imported before connect). I also made the module level `__getattr__` actually error with the correct Exception in case a user wants to access something that doesn't exists. For example `import lamindb as ln; ln.blah` will now correctly raise an `AttributeError`.

I'll continue with this a bit next week.

### DIFF
--- a/lamindb/__init__.py
+++ b/lamindb/__init__.py
@@ -107,32 +107,99 @@ Backwards compatibility.
 
 """
 
-# ruff: noqa: I001
+from __future__ import annotations
+
+import importlib
+import warnings
+from typing import TYPE_CHECKING
+
+from lamindb_setup import connect
+from lamindb_setup._check_setup import _check_instance_setup
+from lamindb_setup.core.upath import UPath
+
 # denote a release candidate for 0.1.0 with 0.1rc1, 0.1a1, 0.1b1, etc.
 __version__ = "1.6.1"
 
-import warnings
+_loaded_submodules_pre_connect = [
+    "base",
+    "errors",
+    "setup",
+]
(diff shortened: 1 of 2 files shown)


ap--:
The failing tests are currently due to my code changes. I'll debug what I am missing.

ap--:
_some notes for next week_

It should probably be revisited why importlib.reload is needed at all in `lamindb`. My guess is some of the issues I am seeing are due to the caveats around https://docs.python.org/3/library/importlib.html#importlib.reload and "from imports".

If it's required and can't be avoided it might be useful to adopt a ruff style for imports that prevents named from imports from within lamindb.

ap--:
Closing in favour of #2849

PR #2828, opened 2025-06-11
✨ Add `dtype` `path`

sunnyosun:
```python

df = pd.DataFrame(
    {
        "sample": ["Sample_X", "Sample_Y", "Sample_Y"],
        "fastq_1": [
            "https://raw.githubusercontent.com/nf-core/test-datasets/scrnaseq/testdata/cellranger/Sample_X_S1_L001_R1_001.fastq.gz",
            "https://raw.githubusercontent.com/nf-core/test-datasets/scrnaseq/testdata/cellranger/Sample_Y_S1_L001_R1_001.fastq.gz",
            "https://raw.githubusercontent.com/nf-core/test-datasets/scrnaseq/testdata/cellranger/Sample_Y_S1_L002_R1_001.fastq.gz",
        ],
        "fastq_2": [
            "https://raw.githubusercontent.com/nf-core/test-datasets/scrnaseq/testdata/cellranger/Sample_X_S1_L001_R2_001.fastq.gz",
            "https://raw.githubusercontent.com/nf-core/test-datasets/scrnaseq/testdata/cellranger/Sample_Y_S1_L001_R2_001.fastq.gz",
            "https://raw.githubusercontent.com/nf-core/test-datasets/scrnaseq/testdata/cellranger/Sample_Y_S1_L002_R2_001.fastq.gz",
        ],
        "expected_cells": [5000, 5000, 5000],
    }
)

nextflow_schema = ln.Schema(
    name="nf-core/scrnaseq standard",
    features=[
        ln.Feature(
            name="sample",
            dtype="str",
            nullable=False,
        ).save(),
        ln.Feature(
            name="fastq_1",
            dtype="path",
            nullable=False,
        ).save(),
        ln.Feature(
            name="fastq_2",
            dtype="path",
        ).save(),
        ln.Feature(
            name="expected_cells",
            dtype=int,
        ).save(),
    ],
).save()

curator = ln.curators.DataFrameCurator(df, schema=nextflow_schema)
curator.validate()
```

Note that the `path` dtype is validated as `str`, as we can't guarantee the path is valid in the environment that registers the schema. For instance, if the file path is used in a pipeline in a specific environment, it will only be accessible there.

The main reason of introducing `path` is to allow users to select file paths in the UI.


lamindb/base/types.py
```diff
@@ -51,6 +51,7 @@
     "datetime",  # datetime
     "dict",  # dictionary
     "object",  # this is a pandas input dtype, we're only using it for complicated types, not for strings
+    "path",  # path
```
falexwolf: Can you also add documentation, please?

Given this is absent in pandera & pandas and only present in pydantic with a different convention, you have to explain how it's exactly defined in LaminDB.

image

You might need to re-organize the table a bit.

falexwolf: Also, this here is confusing as it doesn't exist in pandas or pandera. It only exists in pydantic with this syntax

image

sunnyosun:
> This is good up to documentation (adding an example to the `Feature` page) and the question: where is it validated that the path is path-like? I mean why might not actually want to do that because it could be very expensive, but in that case it should be clearly documented under the dtype definition.

I explained in the description that it's validated as "str".

PR #2835, opened 2025-06-13
🐛 Have transfer comply with `settings.annotation.n_max_records` and fix a bug related to repeated schema transfer

falexwolf:
Fixes the two problems mentioned above.


falexwolf:
I will comprehensively refactor `transfer` and `FeatureManager` next.

PR #2839, opened 2025-06-14
♻️ Refactor storage deletion

falexwolf:


lamindb/models/core.py
```diff
@@ -125,3 +128,29 @@ def path(self) -> Path | UPath:
 
         access_token = self._access_token if hasattr(self, "_access_token") else None
         return create_path(self.root, access_token=access_token)
+
+    def delete(self) -> None:
+        """Delete the storage location.
+
+        This errors in case the storage location is not empty.
+        """
+        from lamindb_setup.core._hub_core import delete_storage_record
```
Koncopd, line 137: Is there really a reason to import things here rather than at the top?

falexwolf: No, not in this case, sorry. I will refactor in another PR.

falexwolf: Done: https://github.com/laminlabs/lamindb/pull/2841/commits/55e82bac7ff46f7051cd7a6dbf50d59e852396bc

PR #2841, opened 2025-06-14
🚸 Enable users to more conveniently create storage locations, make storage management more transparent

falexwolf:
## API design improvements

Before this PR, users could only _implicitly create_ storage locations by _switching_ storage locations:

```python
ln.settings.storage = "s3://my-bucket/my-new-location"
```

While this continues to work (and errors in problematic cases), we can now advocate an explicit call of the `Storage()` constructor:

```python
ln.Storage(root="s3://my-bucket/my-new-location").save()
```

Because we also just introduced a comprehensive `storage.delete()` experience, users can now safely manage the full life cycle of a storage location.

- https://github.com/laminlabs/lamindb/pull/2839

## Governance improvements

LaminDB now also throws an error in case a new _managed_ storage location would be implicitly registered when registering an existing artifact path. This is because inadvertent, non-thoughtful registration of managed storage locations can deadlock an organization. A user accidentally registers a whole bucket or drive and then nobody else can register storage locations in that bucket or drive anymore (because we don't allow nested storage locations, even though, of course, we could consider it in the future, but it's complicated). This is a feature because it creates a nice data flow DAG on the storage location level.

It will continue to silently create referenced storage locations in the current instance for convenience.

## Materials

Needs:

- https://github.com/laminlabs/lamindb-setup/pull/1058
- https://github.com/laminlabs/lamindb-setup/pull/1059


lamindb/models/artifact.py
```diff
@@ -180,10 +178,18 @@ def process_pathlike(
                     new_root = list(filepath.parents)[-1]
                 # do not register remote storage locations on hub if the current instance
                 # is not managed on the hub
-                storage_settings, _ = init_storage(
-                    new_root, prevent_register_hub=not setup_settings.instance.is_on_hub
-                )
-                storage_record = register_storage_in_instance(storage_settings)
+                if not os.getenv("LAMIN_TESTING") == "true":
+                    response = input(
```
Koncopd: This is not really convenient in my opinion, especially if you want to register a lot of different artifacts in a script. I think there should be an option to force `yes` automatically.

Koncopd: I think there should **always** be a way to avoid input, just not turned on by default.

falexwolf: I hadn't read this comment, but I replaced it with throwing a hard `UnknownStorageLocation` error in the next commit.

I'm coming from the worry that inadvertent, non-thoughtful registration of storage locations can readily deadlock an organization. A user accidentally registers a whole bucket or a whole drive/root on a server and then nobody else can register new storage locations anymore (because we don't allow nested storage locations, even though, of course, we could consider it in the future, but it's complicated).

Instead, I think, it's in fact a nice UX to be thoughtful about which storage locations I want to manage with LaminDB, and then make a simple API call: `ln.Storage(root="s3://my-bucket/my-new-location").save()`

I think this is particularly helpful for the upcoming "citizen registrations of data in various locations" in a new org customer.

Koncopd: It was just so convenient to create an artifact and register a storage automatically. This PR feels like a very big degradation of usability.

falexwolf: Ok, I think I have a compromise. There is no downside of ad-hoc registering non-managed storage locations, because these don't lead to a deadlock situation.

For managed storage locations, the user should be aware. They will likely even check the infrastructure dashboad of the organization.

Koncopd: I am not sure this is a good idea either, we will create a lot of confusion if different types of storages behave very differently.

falexwolf: But they already behave very differently. And the deadlocking creates a lot of confusion if it's not explicit. Think back to how we resolved the deadlocking for all the CI storage locations.

We can't avoid

1. the constraints for managed storage locations
2. the distinction between managed, referenced, and non-managed storage locations

If we could, that'd be interesting to consider, but I'm not even sure it's better because LaminDB forces you to create a data flow DAG on the level of storage locations through these concepts & constraints (data can only flow one way). And I believe this is a lot better than the generically mutable storage locations that most organizations have (buckets, server storage).

Koncopd: But then maybe it should be checked in `init_storage` then, after managed storage check.

Koncopd: But i will review the code logic later.

lamindb/models/artifact.py
```diff
@@ -171,19 +169,22 @@ def process_pathlike(
...
-                storage_settings, _ = init_storage(
-                    new_root, prevent_register_hub=not setup_settings.instance.is_on_hub
-                )
-                storage_record = register_storage_in_instance(storage_settings)
+                    new_root = list(filepath.parents)[-1].as_posix().rstrip("/")
+                storage_record = Storage(root=new_root).save()
+                if storage_record.instance_uid == setup_settings.instance.uid:
+                    # we don't want to inadvertently create managed storage locations
+                    # hence, we revert the creation and throw an error
+                    storage_record.delete()
```
Koncopd, line 183: What i don't like here is this create and then delete logic if it is not correct, taking into account that storage creation is a pretty heavy operation now.

Koncopd: We have similar logic (create then delete if wrong) in `init_storage` and i think it should also be reworked.

falexwolf: I agree that this is bad and it all goes back to that it's so hard to know in what category a storage location falls:

- testing write access needs a request to S3, trying to write something
- testing "storage location already managed by another instance" needs a hub request

This is robust and tested in the storage creation logic, but we don't yet have good abstractions to call this outside of creating a storage location.

I think this should be refactored but this is independent from the objective of this PR.

Koncopd:
Now you get errors because `check_path_in_existing_storage` finds an existing storage in the hub, calls `Storage(...).save()` to create it in the instance and triggers the new logic of creating the same storage in the hub inside `Storage.save`.

Koncopd:
It feels like it gets very complicated.

Koncopd:
https://github.com/laminlabs/lamindb/blob/b2dff3f758d69caf16c7aeec73816afadd4e4707/lamindb/models/artifact.py#L356
here it should avoid saving to the hub

falexwolf:
Re your hypothesis: I don't think you're right, but I might be wrong.

I unfortunately missed one aspect before but I actually believe this was a bug in lamindb-setup, so I made another PR:

- https://github.com/laminlabs/lamindb-setup/pull/1059

Re overall complexity: I think it gets simpler because it carves out more explicit patterns. The code organization here would benefit much from integrating lamindb-setup into lamindb, because then one could consistently leverage the `Storage()` interface. But of course that'd be a huge change so I keep two equivalent patterns:

```# lamindb-setup
ssettings, hub_record_status = init_storage(
    root=root, instance_id=settings.instance._id, register_hub=True
)
register_storage_in_instance(ssettings)
```

```# lamindb
ln.Storage(root=root).save()
```

The later is an abstraction over these two calls that we otherwise have to call subsequently everywhere.

All this being said: design choices within lamindb-setup can be much improved. We'll do a refactor sooner or later that makes everything much simpler and clearer. The patterns have stabilized for a very long time now so that we don't have to anticipate new "kinds of storage locations" or registration patterns etc., which is great.

Koncopd:
```
lamindb/models/artifact.py:1406: in __init__
    kwargs_or_artifact, privates = get_artifact_kwargs_from_data(
lamindb/models/artifact.py:397: in get_artifact_kwargs_from_data
    memory_rep, path, suffix, storage, use_existing_storage_key = process_data(
lamindb/models/artifact.py:237: in process_data
    storage, use_existing_storage_key = process_pathlike(
lamindb/models/artifact.py:158: in process_pathlike
    result = check_path_in_existing_storage(
lamindb/models/artifact.py:357: in check_path_in_existing_storage
    return Storage(**result).save()
           ^^^^^^^^^^^^^^^^^
lamindb/models/storage.py:143: in __init__
    ssettings, _ = init_storage(
/opt/hostedtoolcache/Python/3.13.4/x64/lib/python3.13/site-packages/lamindb_setup/core/_settings_storage.py:161: in init_storage
    hub_record_status = init_storage_hub(
/opt/hostedtoolcache/Python/3.13.4/x64/lib/python3.13/site-packages/lamindb_setup/core/_hub_core.py:157: in init_storage_hub
    return call_with_fallback_auth(
/opt/hostedtoolcache/Python/3.13.4/x64/lib/python3.13/site-packages/lamindb_setup/core/_hub_client.py:157: in call_with_fallback_auth
    raise e
/opt/hostedtoolcache/Python/3.13.4/x64/lib/python3.13/site-packages/lamindb_setup/core/_hub_client.py:143: in call_with_fallback_auth
    result = callable(**kwargs, client=client)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/opt/hostedtoolcache/Python/3.13.4/x64/lib/python3.13/site-packages/lamindb_setup/core/_hub_core.py:186: in _init_storage_hub
    if _select_storage(ssettings, update_uid=True, client=client):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

ssettings = StorageSettings(root='s3://lamindata', uid='06cugfEp9dh2')
update_uid = True
client = <supabase._sync.client.SyncClient object at 0x7f0698ba2810>

    def _select_storage(
        ssettings: StorageSettings, update_uid: bool, client: Client
    ) -> bool:
        root = ssettings.root_as_str
        response = client.table("storage").select("*").eq("root", root).execute()
        if not response.data:
            return False
        else:
            existing_storage = response.data[0]
            if existing_storage["instance_id"] is not None:
                if ssettings._instance_id is not None:
                    # consider storage settings that are meant to be managed by an instance
                    if UUID(existing_storage["instance_id"]) != ssettings._instance_id:
                        # everything is alright if the instance_id matches
                        # we're probably just switching storage locations
                        # below can be turned into a warning and then delegate the error
                        # to a unique constraint violation below
>                       raise ValueError(
                            f"Storage root {root} is already managed by instance {existing_storage['instance_id']}."
                        )
E                       ValueError: Storage root s3://lamindata is already managed by instance 037ba1e0-8d80-4f91-a902-75a47735076a.

/opt/hostedtoolcache/Python/3.13.4/x64/lib/python3.13/site-packages/lamindb_setup/core/_hub_core.py:[113](https://github.com/laminlabs/lamindb/actions/runs/15656076203/job/44107324020?pr=2841#step:15:114): ValueError
```
This looks exactly what i described.

Koncopd:
>I unfortunately missed one aspect before but I actually believe this was a bug in lamindb-setup, so I made another PR:

I am not sure what is the bug, we will have to go through this together. I have to ask you to wait before merging it, i don't think there are any bugs there )maybe not right.)

But the thing that you see in the CI now is the problem that i described, i am pretty sure.

Koncopd:
`_skip_preparation` in the referenced place should solve the problem.

falexwolf:
image

This is another architecturally weird design choice in lamindb-setup. It's just a consequence of ad-hoc programming.

I might be missing something again, but if this now goes through things have become quite a bit simpler and with a bit of renaming one can create a nice abstraction for "select storage from hub to tell me which instance manages this storage" and other things.

The problem is that the code was written at a time when concepts like "this instance manages this storage location" were not existent. So, of course the code couldn't be well organized and ended up with the complexity it has today.

Koncopd:
Maybe I am not following, but some changes look very suspicious to me,  please don’t merge this until Monday, let’s go through the changes and the code together then.

falexwolf:
Yes, everything passes now with considerable disentanglement in lamindb-setup.

I would love to merge it for the improvements about CI infra in general that are also in this PR. I want to work on the whole transfer architecture tomorrow and this will be a huge new PR.

So, ideally I could merge and we have a thorough discussion on Monday morning. And if I really did make a mistake, then I'll make another PR to fix it. Of course this here won't go in a release 💯 without proper discussion before.

Btw, I expanded the tests and all previous tests are still passing. So, I wasn't cheating if you believe that. 😅

Koncopd:
I still think the above comment about `_skip_preparation` in `check_path_in_existing_storage` is correct.

And the condition cuts in `lamindb-setup` look pretty random to me.

I am really afraid we are breaking some logic and it will be hard to figure out where when we face random problems. This is why I think we should be really cautious with such changes, we really had problems persisting for a lot of time until we fixed recently.

falexwolf:
> I still think the above comment about _skip_preparation in check_path_in_existing_storage is correct.

If you mean that you'd want to add `_skip_preparation=True`, then you'd actually fundamentally change logic from the previous implementation. Because the previous implementation always made the explicit call to:

```
  storage_settings, _ = init_storage(
      new_root, prevent_register_hub=not setup_settings.instance.is_on_hub
  )
```

And of course that's now happening within the `Storage` constructor.

> And the condition cuts in lamindb-setup look pretty random to me.

Let's discuss them in depth on Monday morning!

falexwolf:
I'm just keeping track of issues -- we'll discuss them all on Monday.

I've run the new code on the `introduction.ipynb` notebook, and what has happened there so far is that a local instance that's not registered on the hub but has (somewhat accidentally and unrelatedly) write access to `s3://lamindata` implicitly created a managed storage location, despite that storage location already being managed by the `lamindata` instance.

Now this fails. I think that's a good thing because the previous situation was indeed dangerous because -- while an edge case that only appears when instances are not registered on the hub -- it allowed that two lamindb instances would both write to the same storage location. There is a finite probability that this could happen in test runs based on test instances that are accidentally pointed to production storage locations (which is exactly what happened on `lamin-docs` CI just now).

This is the error I'm currently getting in:

- https://github.com/laminlabs/lamindb/pull/2842

image

The solution is, IMO, to not be naive about storage locations that are already initialized: https://github.com/laminlabs/lamindb-setup/pull/1059#discussion_r2147340085

There is an alternative solution in which we _always_ making a request to the hub, even if the current instance itself is not registered. But this seems inconsistent and of course it'd miss the case in which the other instance is also not on the hub.

PR #2849, opened 2025-06-16
🚸 Lazy import proxy for `lamindb.models`

ap--:
Hi Lamins,

In case the default connect to dummy instances is problematic, here's the POC PR for using the lazy import proxy mechanism throughout the lamindb tools.

It requires another PR in laminlabs/lamindb-setup#1064 and laminlabs/bionty#265 (and would need others like wetlab, clinicore, etc... to be updated too)

This replaces #2825

### DIFF
--- a/lamindb/__init__.py
+++ b/lamindb/__init__.py
@@ -107,6 +107,14 @@ Backwards compatibility.
 
 """
 
+from lamindb_setup.lazy_import import enable_lazy_imports
+from lamindb_setup._check_setup import _check_instance_setup
+
+enable_lazy_imports(
+    "lamindb.models",
+    on_load=lambda mod, attr: _check_instance_setup(from_module="lamindb"),
+)
+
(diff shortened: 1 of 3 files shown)


falexwolf:
I think we can close this now with:

- https://github.com/laminlabs/lamindb/pull/2851

This was a great way of working around the Django models, but 2851 should ultimately be the simpler solution. If there are problems with 2851, we can come back here.

falexwolf:
Re-opening as an alternative way to work around the current Django reset problems in case Django is set up twice.

One alternative is:

- https://github.com/laminlabs/lamindb-setup/pull/1231

falexwolf:
Added commits to this branch to avoid force pushing here after resolving conflicts locally:

- https://github.com/laminlabs/lamindb/pull/3476

falexwolf:
Superseded by:

- https://github.com/laminlabs/lamindb/pull/3580

PR #2850, opened 2025-06-17
Alwaysconnect

falexwolf:

PR #2851, opened 2025-06-17
🚸 Always import the entire public API & enable switching instances in the same Python session

falexwolf:
## Summary

This PR removes the conditioning the import of wide parts of the lamindb API on the dynamic result of `_check_instance_setup()`: the complete API is now _always_ available upon `import lamindb`.

This enables static type checking and eliminates setup quirks: developers no longer have to be aware of the setup cascade. LaminDB now behaves like a regular static python package that can be imported anywhere and anytime.

The reason for why this wasn't the case before are Django models: Django models require that `django.setup()` is called prior to importing them (they establish reverse relationships with their attributes dynamically, and this depends on the modules/apps configured). Now we account for this requirement by using a "mock" in-memory database to enable setting up Django in all cases.

When a default instance is configured in the environment and `auto-connect` is on lamindb will still connect to the user-configured instance.

## [experimental] Re-connecting to other instances in the same Python process

This PR also introduces the experimental ability of re-connecting to other instances in the same Python process.

It is experimental because Django mutates globally scoped objects during `setup()`: re-connecting an instance requires re-setting-up Django and clearing all Django objects from the Python system cache. This PR comes with the corresponding logic that achieves this but there will be edge cases in which this logic isn't comprehensive enough.

Hence, `lamindb` will print this warning:

image

It will throw a `CannotSwitchDefaultInstance` error if the user attempts to `init` a new instance or if `ln.track()` is live.

image

See https://ecd81c4c.lamindb.pages.dev/faq/setup for the full documentation.

### Upon `ln.track()`

```python
>>> import lamindb as ln
! not connected, call: ln.connect('account/name')
>>> ln.track()
  File "/Users/falexwolf/repos/laminhub/rest-hub/sub/lamindb/lamindb/core/_context.py", line 355, in _track
    instance_settings = ln_setup.settings.instance
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/falexwolf/repos/laminhub/rest-hub/sub/lamindb/sub/lamindb-setup/lamindb_setup/core/_settings.py", line 152, in instance
    self._instance_settings = load_instance_settings()
                              ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/falexwolf/repos/laminhub/rest-hub/sub/lamindb/sub/lamindb-setup/lamindb_setup/core/_settings_load.py", line 42, in load_instance_settings
    raise CurrentInstanceNotConfigured
lamindb_setup.errors.CurrentInstanceNotConfigured: No instance is connected! Call
- CLI:     lamin connect / lamin init
- Python:  ln.connect()  / ln.setup.init()
- R:       ln$connect()  / ln$setup$init()
```

### Upon query

We do not yet have a solution yet for when the user attempts queries -- the challenge is that we need to pass a valid db connection string to Django and then Django would need to be intercepted later. 

The user sees a warning if they import lamindb without a default instance configured.

image

If the user ignores that warning, they will run into the following error.

<details><summary>Current error behavior</summary>
<p>
>>> ln.Artifact.df()
  File "/Users/falexwolf/miniconda3/envs/py312/lib/python3.12/site-packages/django/db/backends/utils.py", line 105, in _execute
    return self.cursor.execute(sql, params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/falexwolf/miniconda3/envs/py312/lib/python3.12/site-packages/django/db/backends/sqlite3/base.py", line 354, in execute
    return super().execute(query, params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
sqlite3.OperationalError: no such table: lamindb_artifact

The above exception was the direct cause of the following exception:

  File "/Users/falexwolf/repos/laminhub/rest-hub/sub/lamindb/lamindb/models/sqlrecord.py", line 505, in df
    return query_set[:limit].df(include=include, features=features)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/falexwolf/repos/laminhub/rest-hub/sub/lamindb/lamindb/models/query_set.py", line 650, in df
    df = pd.DataFrame(queryset.values(*field_names, *list(annotate_kwargs.keys())))
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/falexwolf/miniconda3/envs/py312/lib/python3.12/site-packages/pandas/core/frame.py", line 843, in __init__
    data = list(data)
           ^^^^^^^^^^
  File "/Users/falexwolf/miniconda3/envs/py312/lib/python3.12/site-packages/django/db/models/query.py", line 381, in __iter__
    self._fetch_all()
  File "/Users/falexwolf/miniconda3/envs/py312/lib/python3.12/site-packages/django/db/models/query.py", line 1909, in _fetch_all
    self._result_cache = list(self._iterable_class(self))
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/falexwolf/miniconda3/envs/py312/lib/python3.12/site-packages/django/db/models/query.py", line 213, in __iter__
    for row in compiler.results_iter(
               ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/falexwolf/miniconda3/envs/py312/lib/python3.12/site-packages/django/db/models/sql/compiler.py", line 1531, in results_iter
    results = self.execute_sql(
              ^^^^^^^^^^^^^^^^^
  File "/Users/falexwolf/miniconda3/envs/py312/lib/python3.12/site-packages/django/db/models/sql/compiler.py", line 1580, in execute_sql
    cursor.execute(sql, params)
  File "/Users/falexwolf/miniconda3/envs/py312/lib/python3.12/site-packages/django/db/backends/utils.py", line 79, in execute
    return self._execute_with_wrappers(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/falexwolf/miniconda3/envs/py312/lib/python3.12/site-packages/django/db/backends/utils.py", line 92, in _execute_with_wrappers
    return executor(sql, params, many, context)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/falexwolf/miniconda3/envs/py312/lib/python3.12/site-packages/django/db/backends/utils.py", line 100, in _execute
    with self.db.wrap_database_errors:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/falexwolf/miniconda3/envs/py312/lib/python3.12/site-packages/django/db/utils.py", line 91, in __exit__
    raise dj_exc_value.with_traceback(traceback) from exc_value
  File "/Users/falexwolf/miniconda3/envs/py312/lib/python3.12/site-packages/django/db/backends/utils.py", line 105, in _execute
    return self.cursor.execute(sql, params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/falexwolf/miniconda3/envs/py312/lib/python3.12/site-packages/django/db/backends/sqlite3/base.py", line 354, in execute
    return super().execute(query, params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
django.db.utils.OperationalError: no such table: lamindb_artifact
</p>
</details> 

### Upon toggling auto-connect

Behavior upon trying to toggle auto-connect:
```bash
% lamin settings set auto-connect false
! setting auto_connect to `False` no longer has an effect and the setting will likely be removed in the future; since lamindb 1.7, auto_connect `True` no longer clashes with connecting in a Python session
```

## Materials

Credit goes to @Koncopd through the below change that enables switching instances in the same python session:

- https://github.com/laminlabs/lamindb-setup/pull/1046

The other important change that enabled removing the gated models import is here:

- https://github.com/laminlabs/lamindb-setup/pull/1063

This enables a few simplifications in the CLI:

- https://github.com/laminlabs/lamin-cli/pull/138

Supersedes:

- https://github.com/laminlabs/lamindb/pull/2849
- https://github.com/laminlabs/lamindb/pull/2850


falexwolf:
Ok, now tests are working.

tests/storage/conftest.py
```diff
@@ -51,7 +51,7 @@ def pytest_sessionstart():
         quit()
     total_time_elapsed = perf_counter() - t_execute_start
     print(f"time to setup the instance: {total_time_elapsed:.1f}s")
-    assert ln.Storage.filter(root="s3://lamindb-ci/test-data").one_or_none() is not None
+    # assert ln.Storage.filter(root="s3://lamindb-ci/test-data").one_or_none() is not None
```
Koncopd: Why is it commented out?

lamindb/models/artifact.py
```diff
@@ -1090,6 +1090,7 @@ class Artifact(SQLRecord, IsVersioned, TracksRun, TracksUpdates):
 
     class Meta(SQLRecord.Meta, IsVersioned.Meta, TracksRun.Meta, TracksUpdates.Meta):
         abstract = False
+        app_label = "lamindb"
```
Koncopd, line 1113: Is this really needed?

Koncopd: I think the classes should hold references to the module where they are defined.

Koncopd: `ln.Artifact.__module__` -> `'lamindb.models.artifact'`, so this info should be there already without any additional variables.

Koncopd: Also i can't find where it is used.

Koncopd: Aha, ok, it is needed by `django`, interesting.

tests/storage/conftest.py
falexwolf: There indeed was an issue here that I hadn't yet caught. Now dealt with here: https://github.com/laminlabs/lamindb-setup/pull/1063/commits/7523b970ea031d211854f2cd6ece9697239680cb

This is now maybe the remaining issue: some modules that might import lamindb somewhere. But we'll gather some experience for this and I think it's less bad than the constant errors you obtain when you try to normally develop with lamindb without all these changes of this PR.

PR #2852, opened 2025-06-17
🎨 Support nested cat dtypes

sunnyosun:
Handles nested dtypes like `cat[ULabel[Sample[BioSample]]]`.


falexwolf (approved):
Looks great to me!

PR #2862, opened 2025-06-25
🚸 Mark failed upload of artifacts via `artifact._is_saved_to_storage_location`

Koncopd:
Now artifacts have `._is_saved_to_storage_location` property which is `True` only if upload or local saving was completed successfully. 

`artifact._is_saved_to_storage_location` is meaningful only after calling `.save()`. `None` means no writing was necessary, `True` - that it was written correctly. `False` shows that there was a problem with writing.

```python
artifact.save() # or ln.save([artifact])
assert artifact._is_saved_to_storage_location
```

### DIFF
--- a/lamindb/models/artifact.py
+++ b/lamindb/models/artifact.py
@@ -79,6 +79,7 @@ from ._relations import (
 from .feature import Feature, FeatureValue
 from .has_parents import view_lineage
 from .run import Run, TracksRun, TracksUpdates, User
+from .save import check_and_attempt_clearing, check_and_attempt_upload
 from .schema import Schema
 from .sqlrecord import (
     BaseSQLRecord,
@@ -2675,8 +2676,6 @@ class Artifact(SQLRecord, IsVersioned, TracksRun, TracksUpdates):
             self._save_complete = False  # will be updated to True at the end
 
         self._save_skip_storage(**kwargs)
-
-        from .save import check_and_attempt_clearing, check_and_attempt_upload
 
         using_key = None
         if "using" in kwargs:
--- a/lamindb/models/artifact.py
+++ b/lamindb/models/artifact.py
@@ -1103,6 +1103,9 @@ class Artifact(SQLRecord, IsVersioned, TracksRun, TracksUpdates):
             ),
         ]
 
+    _aux_fields: dict[str, tuple[str, type]] = {
+        "0": ("_save_complete", bool),
+    }
     _len_full_uid: int = 20
     _len_stem_uid: int = 16
 
@@ -2624,6 +2627,18 @@ class Artifact(SQLRecord, IsVersioned, TracksRun, TracksUpdates):
                 if delete_msg != "did-not-delete":
                     logger.success(f"deleted {colors.yellow(f'{path}')}")
 
+    @property
+    def _save_complete(self) -> bool | None:
+        if self._aux is not None:
+            return self._aux.get("auf", {}).get("0", None)
+        else:
+            return None
+
+    @_save_complete.setter
+    def _save_complete(self, value: bool) -> None:
+        self._aux = self._aux or {}
+        self._aux.setdefault("af", {})["0"] = value
+
     def save(self, upload: bool | None = None, **kwargs) -> Artifact:
         """Save to database & storage.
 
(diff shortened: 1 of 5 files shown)


lamindb/models/save.py
```diff
@@ -105,6 +105,8 @@ def save(
     if artifacts:
         with transaction.atomic():
             for record in artifacts:
+                # will swtich to True after the successful upload / saving
+                record._save_complete = False
```
falexwolf: Why is this necessary, the default should be `False`?

lamindb/models/artifact.py
```diff
@@ -1103,6 +1104,9 @@ class Meta(SQLRecord.Meta, IsVersioned.Meta, TracksRun.Meta, TracksUpdates.Meta)
             ),
         ]
 
+    _aux_fields: dict[str, tuple[str, type]] = {
```
falexwolf, line 1107: We need to set the default value to `False` in the Artifact `__init__`.

lamindb/models/save.py
```diff
@@ -333,7 +337,10 @@ def store_artifacts(
         if exception is not None:
             break
         stored_artifacts += [artifact]
-        # if check_and_attempt_upload was successfull
+        # update to show successful saving
+        artifact._save_complete = True
```
falexwolf: Why can't we handle toggling within `Artifact.save()`? This seems the wrong place here. 🤔

tests/storage/test_artifact_zarr.py
```diff
@@ -30,10 +30,13 @@ def test_zarr_upload_cache(get_small_adata):
     get_small_adata.write_zarr(zarr_path)
 
     artifact = ln.Artifact(zarr_path, key="test_adata.zarr")
+    assert artifact._save_complete is None
```
falexwolf: This should be `False`, not `None`.

lamindb/models/artifact.py
```diff
@@ -2654,9 +2670,12 @@ def save(self, upload: bool | None = None, **kwargs) -> Artifact:
             # ensure that the artifact is uploaded
             self._to_store = True
 
-        self._save_skip_storage(**kwargs)
+        # _save_complete indicates whether the saving / upload process is successfull
+        flag_complete = hasattr(self, "_local_filepath")
+        if flag_complete:
+            self._save_complete = False  # will be updated to True at the end
```
falexwolf: This should be done in the `__init__` as a default value.

lamindb/models/artifact.py
falexwolf: I think it needs to be called after super().__init__(), parallel to here

https://github.com/laminlabs/lamindb/blob/cf5ede63b2abc15e8960830b0afd0a6f81d31636/lamindb/models/feature.py#L636-L638

falexwolf:
Great!

I made a few comments that should be easy to address. Overall it's nice that this is so simple.

I'm not sure about the name `save_complete` as opposed to `upload_complete` or e.g. `is_saved_to_storage_location` or something more explicit.

I think I'd go for something more explicit because "save" is so overloaded and we're talking here specifically about the storage component.

lamindb/models/save.py
Koncopd: I don't agree, it should not be `False`, i don't want this to be `False` for already existing artifacts, i also don't want to set it to `False` when we just update artifact fields without uploading anything.

lamindb/models/save.py
Koncopd: This is for bulk saving, it doesn't use `Artifact.save()`.

lamindb/models/artifact.py
Koncopd: See the answer to the first comment.

tests/storage/test_artifact_zarr.py
Koncopd: See the answer to the first comment.

lamindb/models/artifact.py
Koncopd: See the answer to the first comment.

lamindb/models/save.py
falexwolf: I'll call you.

falexwolf: Ok.

lamindb/models/artifact.py
falexwolf: Ok.

Koncopd:
Renamed to `._is_saved_to_storage_location`.

PR #2874, opened 2025-06-30
✨ Add Settings `__repr__`

Zethson:
By user request.

| Before | After |
| ------ | ----- |

### DIFF
--- a/lamindb/core/_settings.py
+++ b/lamindb/core/_settings.py
@@ -48,7 +48,7 @@ class Settings:
         verbosity_color = colors.yellow if self.verbosity == "warning" else colors.green
         verbosity_str = verbosity_color(self.verbosity)
 
-        storage_root = str(self._storage_settings.root)
+        storage_root = self._storage_settings.root_as_str
         storage_str = colors.italic(storage_root)
 
         instance_str = colors.italic(self.instance_uid)


lamindb/core/_settings.py
```diff
@@ -43,6 +43,36 @@ def __init__(self):
         logger.set_verbosity(self._verbosity_int)
         self._sync_git_repo: str | None = None
 
+    def __repr__(self) -> str:  # pragma: no cover
+        cls_name = colors.green(self.__class__.__name__)
+        verbosity_color = colors.yellow if self.verbosity == "warning" else colors.green
+        verbosity_str = verbosity_color(self.verbosity)
+
+        storage_root = str(self._storage_settings.root)
+        storage_str = colors.italic(storage_root)
```
Koncopd, line 52: I am not sure we need colors here.

Koncopd: @falexwolf what do you think?

falexwolf: I think it'd be better if it was consistent with the setup settings (so either no colors here or also colors there).

I also wanted to add the setup settings to the settings under `.setup` at some point.

But I also don't think we have to overthink it.

Zethson: I think the general colors and the way I used them in this PR is pretty consistent with other models. So I'd be happy to follow up up with adapting the setup settings.

OK with you @Koncopd ?

lamindb/core/_settings.py
```diff
@@ -43,6 +43,36 @@ def __init__(self):
         logger.set_verbosity(self._verbosity_int)
         self._sync_git_repo: str | None = None
 
+    def __repr__(self) -> str:  # pragma: no cover
+        cls_name = colors.green(self.__class__.__name__)
+        verbosity_color = colors.yellow if self.verbosity == "warning" else colors.green
+        verbosity_str = verbosity_color(self.verbosity)
+
+        storage_root = str(self._storage_settings.root)
```
Koncopd:
```suggestion
        storage_root = self._storage_settings.root_as_str
```
This is better because it avoids triggering access-aws.

lamindb/core/_settings.py
Koncopd: Yes

Koncopd (approved):

Issue #2585, opened 2025-03-21
✨ Rewrite `CellxgeneCatManager` and `PertCatManager` using schemas


sunnyosun:
### Add a description

Need to figure out how to specify the sources etc.

PR #2878, opened 2025-07-02
✨Add Schema based CELLxGENE Curator

Zethson:
Partially fixes https://github.com/laminlabs/lamindb/issues/2585


lamindb/curators/__init__.py
```diff
@@ -8,6 +8,7 @@
    MuDataCurator
    SpatialDataCurator
    TiledbsomaExperimentCurator
+   CxGCurator
```
falexwolf: I don't think we want a Curator anymore, unless I'm wrong. The idea would be to just provide a schema, no? There wouldn't be any specific curator code for CxG, would it?

Zethson: No, the Curator needs custom code for missing ontology values (such as "normal"), missing defaults, handling names vs ontology_id, ...

I currently don't think that we can hack all of that into a Schema (some things we can).

I'll first add tests for the low level functionality that I changed/fixed and then I'll open a Slack thread to discuss this topic. Yesterday, I actually had moved a CxG generator function to `lamindb.examples.anndata` just to move it back. I understand what you're looking for.

Zethson: The function that gets the Schema can probably do some of that. I'll revisit this ASAP but first tests!

sunnyosun: (Sorry that comment was from me)
It should still not have its Curator code, because the code is just to prepare the instance. We need a different strategy to add control terms; or we can just guide users to prepare the instance with a proper notebook.

sunnyosun: We want to have a few schemas available, like cxg, perturbation. Maybe even under `lamindb.schemas.` or something.

sunnyosun:
Is it still possible to move this fix to a new PR?

"Fixes a very nasty bug where Schemas that were already saved earlier were returned with changed _aux which led to import field such as the index (via _index_feature_uid) got removed"

I think it will make both the fix discussion and the cxg schema discussion more organized in their own PRs.

You can re-point this PR to the fix PR.

lamindb/curators/_cellxgene_schemas/schema_versions.csv
```diff
@@ -41,3 +41,14 @@ schema_version,entity,organism,source,version
 5.2.0,Tissue,all,uberon,2024-08-07
 5.2.0,Gene,human,ensembl,release-110
 5.2.0,Gene,mouse,ensembl,release-110
+5.3.0,CellType,all,cl,2025-02-13
```
sunnyosun, line 44: I still think the ontology versions information should be parsed from https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/5.3.0/schema.md directly. Otherwise, we are duplicating the information and also increasing the maintenance of this file.

Zethson: I don't know how to do that reliably... This is a rather free style doc.

sunnyosun: Could you please look into it? I mean single-cell-curation must be using this information for their validator somehow. I think this session is also rather a structured table:
Screenshot 2025-07-03 at 17 28 52

But single-cell-curation should have relevant code already.

Zethson: I found https://github.com/chanzuckerberg/cellxgene-ontology-guide/blob/main/ontology-assets/ontology_info.json but it's only for 5.3.0+. It also lacks Genes just like the table in what you've shown. They have a more comprehensive set up to get gene versions.

I asked whether they'd be willing to add gene versions.

A disadvantage of allowing any version to pulled also causes issues because we have to ensure that all required versions are available in Bionty. Yes, people can technically pull any, but not for all ontologies.

lamindb/curators/__init__.py
falexwolf: I think it is legitimate to have some logic in code and others in a schema. Code and schema are quite different. Evidently the former can't be easily queried for because it's not structured and the later can. Not _everything_ has to fit into a schema. Some could -- as Sunny says -- be part of a "curation notebook" which is likely better a curation script which is likely better some kind of configurable thing in lamindb (a "curator", just that this doesn't have so much to do with what we currently brand as curator which is just a thing that validates the schema).

lamindb/curators/_cellxgene_schemas/schema_versions.csv
Zethson: Active discussion here: https://github.com/chanzuckerberg/cellxgene-ontology-guide/issues/292

tests/curators/test_cxg_curator.py
```diff
@@ -1,36 +1,89 @@
...
+        ).delete()
+    for name in ["tissue", "organoid", "cell culture"]:
+        ln.ULabel.get(name=name, description="From CellxGene schema.").delete()
+    ln.ULabel.get(name="TissueType", is_type=True).delete()
+    for name in ["cell", "nucleus", "na"]:
+        ln.ULabel.get(name=name, description="From CellxGene schema.").delete()
+    ln.ULabel.get(
+        name="SuspensionType",
+        is_type=True,
+    ).delete()
```
sunnyosun: Why 30 lines of code are needed for cleanup? It makes more sense to clean up everything in the instance.

If you want to test certain labels are added, they should be part of the test with proper checks, not here.

lamindb/examples/cellxgene.py
```diff
@@ -0,0 +1,246 @@
...
+    var_schema = Schema(
+        name=f"var of CELLxGENE version {schema_version}",
+        index=Feature(
+            name="var_index",
+            dtype=bt.Gene.ensembl_gene_id,
+            cat_filters={"source": sources["var_index"]},
+        ).save(),
+        itype=Feature,
+        dtype="DataFrame",
+        minimal_set=True,
```
sunnyosun: Why is this minimal_set=True? It means all Features are required?

tests/curators/test_cxg_curator.py
```diff
@@ -1,36 +1,89 @@
...
-        adata,
-        defaults=ln.curators._legacy.CellxGeneAnnDataCatManager.cxg_categoricals_defaults,
-        schema_version=schema_version,
+    schema = ln.examples.cellxgene.get_cxg_schema(
+        schema_version="5.2.0", field_types=["name", "ontology_id"]
     )
-    assert not curator.validate()
+    adata = ln.core.datasets.small_dataset3_cellxgene()
+
+    curator = ln.curators.AnnDataCurator(adata, schema)
```
sunnyosun, line 34: Could you please add more tests for cases that should error? And examples in the description?

Here it only checks var_index. But with minimal_set=True for the var_schema, it should have errored when features are not present in the var columns due to the minimal_set=True.

lamindb/examples/cellxgene.py
```diff
@@ -0,0 +1,246 @@
...
+
+from lamindb.base.types import FieldAttr
+from lamindb.models import Feature, Schema, SQLRecord, ULabel
+from lamindb.models._from_values import _format_values
+
+CELLxGENESchemaVersions = Literal["4.0.0", "5.0.0", "5.1.0", "5.2.0", "5.3.0"]
+FieldType = Literal["ontology_id", "name"]
+
+
+def add_cxg_defaults_to_obs(adata: AnnData) -> None:
```
sunnyosun: This should be properly done with `feature.default_value`, not a custom function.

lamindb/examples/cellxgene.py
```diff
@@ -0,0 +1,246 @@
...
+    }
+    for name, default in cxg_categoricals_defaults.items():
+        if (
+            name not in adata.obs.columns
+            and f"{name}_ontology_term_id" not in adata.obs.columns
+        ):
+            adata.obs[name] = default
+
+
+def save_cxg_defaults() -> None:
```
sunnyosun: Not sure the UX of this function, there's no hint that users should call this function, right?

It'd be good to have it automatically checked in `get_cxg_schema`, if these labels are not added, add them, otherwise don't do anything. This can be a private function.

lamindb/examples/cellxgene.py
```diff
@@ -0,0 +1,246 @@
...
+            source = bt.Source.filter(
+                organism=row.organism,
+                entity=f"bionty.{entity}",
+                name=row.source,
+                version=row.version,
+            ).one_or_none()
+            # if the source was not found, we register it from bionty-assets
+            if source is None:
+                source = getattr(bt, entity).add_source(
+                    bt.Source.using("laminlabs/bionty-assets")
```
sunnyosun: One shouldn't rely on `laminlabs/bionty-assets` as `add_source` can work with any sources, and you don't need them to be pre-curated in bionty-assets.

Simply do `add_source(source=<str>, version=..., organism=...)`.

sunnyosun:
The curator tests will only be run if curator code is changed, I think the cxg test should be run when the examples/cellxgene.py is changed, so maybe move it into it's own folder and tie it to it's own file.

sunnyosun:
A general comment of having so much custom stuff in cellxgene.py is that users will have a hard time to use this as part of their own custom schema (like cellxgene + additional features).

It's fine if you can't come up with anything better, but to me still not the best UX. I would personally prefer to not have any custom code for cellxgene, and only show the schema as an example. Of course this might requires more powerful curators in general, but we should aim for that.

Maintain all these custom code will also became a burden if their default value changes in a certain version in the future.

My preference (as mentioned before) would be a proper guide notebook with custom functions to setup an instance, but not part of our code base.

lamindb/examples/cellxgene.py
Zethson: Yeah and we require a `var_index`, right? Is it wrong to set it to True here?

lamindb/examples/cellxgene.py
Zethson: Yeah, I totally agree that there's no hint that users should call this function. This would be a part of the guide.

We'd be overloading `get_cxg_schema` with unexpected functionality and just getting a Schema would alter the database. I'm torn here but if we think that's more user friendly to just do it, I'm fine with it.

lamindb/examples/cellxgene.py
sunnyosun: I think this means all features in the Feature registry are required as columns of var. Please test it out on your end, maybe I'm wrong.

I'm not sure how var_index is controlled here, it should be required already when you specify index.

lamindb/examples/cellxgene.py
Zethson: `feature.default_value` doesn't add missing columns, right? We can argue that it's fine UX if users don't get these columns in the AnnData object but we just keep the default value annotation. Fine with me.

sunnyosun: The missing columns should have a proper error from the curator if minimal_set=True is set, and users can add themselves, I don't think we need to do everything for them.

lamindb/examples/cellxgene.py
Zethson: Indeed, it doesn't make a difference here. Removed.

lamindb/examples/cellxgene.py
Zethson: We can also have this be a parameter of `get_cxg_schema`?
@falexwolf what do you think? (adding you here because Sunny wanted you to also have a look at some comments here)

falexwolf: I think not integrating is better because `save_...` should be called one time and `get_...` can be called many times.

So, I think it's good as is.

falexwolf:
Given this shows up at the top here and people may come here:

image

Can this PR description be updated so that it features some screenshots of docs? And the changelog entry so that it points to a guide on the docs?

I think that for almost everybody this section here is nuance/technical. It could be at the bottom. At the top it could explain how to curate against CxG in simple terms ideally using a docs screenshot. ☺️ 

image

PR #2888, opened 2025-07-04
✨ Respect column ordering & categorical dtypes in records export, respect nested sub types in curation, enable annotating artifacts by records, enable tracking lineage of records

falexwolf:
This all works in LaminDB but is right now mostly for enabling the UI.

- https://github.com/laminlabs/laminhub/pull/2875


lamindb/curators/core.py
```diff
@@ -1143,9 +1143,11 @@ def _add_validated(self) -> tuple[list, list]:
         # inspect the default instance and save validated records from public
         if self._subtype_str != "" and "=" not in self._subtype_str:
             related_name = registry._meta.get_field("type").remote_field.related_name
-            self._subtype_query_set = getattr(
-                registry.get(name=self._subtype_str), related_name
-            ).all()
+            type_record = registry.get(name=self._subtype_str)
+            if registry.__name__ == "Record":
```
falexwolf, line 1147: This should be `issubclass(HasType)` once that's introduced.

PR #2889, opened 2025-07-04
🐛 Fix updates of re-loaded `Schema._aux`

### LINKED ISSUE #2887: Nested `_aux` fields are not updated correctly when the Schema was already saved
### Add a description

I have observed that the `_aux` field of nested Schemas get overwritten when a Schema was found and is supposed to be returned. Such cases enter:

```python
if schema is not None:
            logger.important(f"returning existing schema with same hash: {schema}")
            init_self_from_db(self, schema)
            update_attributes(self, validated_kwargs)
            return None
```

The culprit is the `update_attributes` function which simply `setattr(record, key, value)` but does not merge the `_aux` keys. A special case where we merge `_aux` keys resolves this issue but it might be too hacky of a solution.

 To reproduce:

```python
!lamin init --storage run-tests --modules bionty

import lamindb as ln
import bionty as bt

var_schema = ln.Schema(
        name=f"CELLxGENE var",
        index=ln.Feature(
            name="var_index",
            dtype=bt.Gene.ensembl_gene_id,
            cat_filters={"source": bt.Source.get("4UGNz3fr")},
        ).save(),
        itype=ln.Feature,
        dtype="DataFrame",
        minimal_set=True,
        coerce_dtype=True,
    ).save()

full_cxg_schema = ln.Schema(
        name=f"CELLxGENE AnnData schema",
        otype="AnnData",
        minimal_set=True,
        coerce_dtype=True,
        slots={"var": var_schema},
    ).save()

print(full_cxg_schema.slots["var"]._aux)
# gives us
# {'af': {'0': True, '2': False, '3': 'iUvpnSXD2s3N'}}

assert len(full_cxg_schema.slots["var"]._aux["af"].keys()) == 3

# NOW WE RUN IT AGAIN

var_schema = ln.Schema(
        name=f"CELLxGENE var",
        index=ln.Feature(
            name="var_index",
            dtype=bt.Gene.ensembl_gene_id,
            cat_filters={"source": bt.Source.get("4UGNz3fr")},
        ).save(),
        itype=ln.Feature,
        dtype="DataFrame",
        minimal_set=True,
        coerce_dtype=True,
    ).save()

full_cxg_schema = ln.Schema(
        name=f"CELLxGENE AnnData schema",
        otype="AnnData",
        minimal_set=True,
        coerce_dtype=True,
        slots={"var": var_schema},
    ).save()

print(full_cxg_schema.slots["var"]._aux)
# give us
# {'af': {'0': True}}


# and therefore the assert fails
assert len(full_cxg_schema.slots["var"]._aux["af"].keys()) == 3
```

Zethson:
Fixes #2887 

- now correctly preserves the `_aux` fields when a Schema that was already saved is returned

Implementation wise, there's a way to reduce the test to ensure that it becomes more compact and readable. However, I opted for a more comprehensive solution that covers more cases.


falexwolf:
This looks interesting! Can you describe the problem a little bit more verbosely?

I spent quite some time having schema mutations behave decently (and consciously allowing them). So I want to carefully understand if this introduces side effects.

falexwolf:
I will look in depth in about an hour.

falexwolf:
Studying this now.

falexwolf:
Ok. What you're doing here is "keeping around parts of _aux from the record loaded from the db" while "adding new parts of _aux from the record that was newly defined".

Let's look at your example:
```python
import lamindb as ln
import bionty as bt

var_schema = ln.Schema(
        name=f"CELLxGENE var",
        index=ln.Feature(
            name="var_index",
            dtype=bt.Gene.ensembl_gene_id,
            cat_filters={"source": bt.Source.get("4UGNz3fr")},
        ).save(),
        itype=ln.Feature,
        dtype="DataFrame",
        minimal_set=True,
        coerce_dtype=True,
    ).save()

full_cxg_schema = ln.Schema(
        name=f"CELLxGENE AnnData schema",
        otype="AnnData",
        minimal_set=True,
        coerce_dtype=True,
        slots={"var": var_schema},
    ).save()

print(full_cxg_schema.slots["var"]._aux)
# gives us
# {'af': {'0': True, '2': False, '3': 'iUvpnSXD2s3N'}}

assert len(full_cxg_schema.slots["var"]._aux["af"].keys()) == 3

# NOW WE RUN IT AGAIN

var_schema = ln.Schema(
        name=f"CELLxGENE var",
        index=ln.Feature(
            name="var_index",
            dtype=bt.Gene.ensembl_gene_id,
            cat_filters={"source": bt.Source.get("4UGNz3fr")},
        ).save(),
        itype=ln.Feature,
        dtype="DataFrame",
        minimal_set=True,
        coerce_dtype=True,
    ).save()

full_cxg_schema = ln.Schema(
        name=f"CELLxGENE AnnData schema",
        otype="AnnData",
        minimal_set=True,
        coerce_dtype=True,
        slots={"var": var_schema},
    ).save()

print(full_cxg_schema.slots["var"]._aux)
# give us
# {'af': {'0': True}}


# and therefore the assert fails
assert len(full_cxg_schema.slots["var"]._aux["af"].keys()) == 3
```

Clearly `full_cxg_schema.slots["var"]._aux` should have all 3 auxiliary fields as in the case where the code ran first. This is a bug.

I'd try to understand why the `_aux` field content of the `var_schema` that gets loaded from the database does not match the `_aux` field content of the code second run of 

```
var_schema = ln.Schema(
        name=f"CELLxGENE var",
        index=ln.Feature(
            name="var_index",
            dtype=bt.Gene.ensembl_gene_id,
            cat_filters={"source": bt.Source.get("4UGNz3fr")},
        ).save(),
        itype=ln.Feature,
        dtype="DataFrame",
        minimal_set=True,
        coerce_dtype=True,
).save()
```

The thing is: the same code runs twice it should hence produce the _same_ result or -- in rare integrity related cases like inadvertent mutations of `dtype` -- should in fact maintain the state of the database.

But here it just looks that for whatever reason the `_aux` content of the `in-memory` schema object doesn't match what's in the database.

And the bug is, IMO, that the aux field state gets set _after_ doing the hash lookup and calling

```
init_self_from_db(self, schema)
update_attributes(...)
```

In particular these lines are problematic:

https://github.com/laminlabs/lamindb/blob/cffe65ab4624b62ffcb57159c24a258290ff1883/lamindb/models/schema.py#L571-L573
https://github.com/laminlabs/lamindb/blob/cffe65ab4624b62ffcb57159c24a258290ff1883/lamindb/models/schema.py#L590-L594

What happens is comparing apples with oranges: the object from the database happens to be correct, and the object in memory _would_ be correct if all logic setting it state had already run at the point when `update_attributes(...)` is called. But  it was not yet called!

So, fixing the bug, IMO, means ensuring that the state of the in memory object mirrors the user input (= the user code meant to set the state of the object).

The solution here is, I fear, dangerous because it doesn't address the root of the problem and introduces the new problem that there might be inadvertent spillage from a database state of the record that gets not cleared by user code. While we likely want the database state to supersede the user instruction in the case of `uid`, `dtype`, and `otype` because updating them would deeply mess with integrity, for everything else, we want a controlled update that enables the user to e.g. add new optional columns to an existing schema:

https://github.com/laminlabs/lamindb/blob/cffe65ab4624b62ffcb57159c24a258290ff1883/lamindb/models/sqlrecord.py#L168

A guided mutation of a schema is a complicated topic that should be based on "allowed rules" (like adding an optional column is allowed, adding a required column is problematic if datasets have been validated with it). Writing down these rules should be done using the `Schema` API, not JSON language. But this is a topic for the future.

Long story short: Do you agree that the issue should rather be fixed by introducing state consistency between database and freshly-created in-memory record rather than mixing their states? And if so, do you want to give it a shot in a new PR?

Zethson:
Thanks!

Yes, your investigation is similar to mine and we came to the same conclusion.

> Do you agree that the issue should rather be fixed by introducing state consistency between database and freshly-created in-memory record rather than mixing their states? 

I had drafted a long answer here but I think just excluding `"_aux"` from the updates should be good enough. The `"_aux"` are already a part of the hash calculation. So if they match, we can just skip over updating them and reuse them, right?

Side question:
```python
    _aux_fields: dict[str, tuple[str, type]] = {
        "0": ("coerce_dtype", bool),
        "1": ("optionals", list[str]),
        "2": ("flexible", bool),
        "3": ("index_feature_uid", str),
    }
```

These number keys are used throughout the code. This is probably hard to improve now but why numbered keys? This is pretty unreadable and confusing in my opinion.

falexwolf:
> I had drafted a long answer here but I think just excluding "_aux" from the updates should be good enough. The "_aux" are already a part of the hash calculation. So if they match, we can just skip over updating them and reuse them, right?

Yes, we can skip. But that's the current behavior on main and it leads to the bug because the state of `_aux` is incorrect at point of the comparison. https://github.com/laminlabs/lamindb/blob/f27e3730b10fa94510a0057b8c195c017dcf32c2/lamindb/models/sqlrecord.py#L167

Zethson:
No, that's not the current on main?

With this simple change:

https://github.com/laminlabs/lamindb/blob/fix/schema_aux/lamindb/models/sqlrecord.py#L168

it's fine and the test passes. But maybe I misunderstand?

falexwolf:
> Side question: ...
> These number keys are used throughout the code. This is probably hard to improve now but why numbered keys? This is pretty unreadable and confusing in my opinion.

Yes, absolutely unreadable and that's why we should use the typed Python API (`schema.flexible` etc.) rather than writing logic on the JSON level. 

> This is probably hard to improve now but why numbered keys?

This was a conscious design decision: because this stuff gets persisted in a JSON field it's almost impossible to migrate it. So, if you rename `coerce_dtype` to `coerce` or other things that aren't so unlikely then you're kinda screwed. I mean you could write a data migration but then you still have the problem that all users would need to update their clients exactly when that data migration happened, which is impossible without throwing hard errors and causing big disruption.

The other reason is the way Postgres models JSON. Even binary JSON is first coded as a text field and then somehow binarized (I don't know how). But every character you add to keys lets indexes grow, queries slow, and is a burden on the database. It's a LOT less efficient than dealing with typed fields.

falexwolf:
> No, that's not the current on main?

It is, see the permalink [I shared](https://github.com/laminlabs/lamindb/pull/2889#issuecomment-3036542089), which clearly shows `getattr(record, key) != value`.

> With this simple change:
> https://github.com/laminlabs/lamindb/blob/fix/schema_aux/lamindb/models/sqlrecord.py#L168
> it's fine and the test passes.

What you're showing me does not implement what you said you would implement, does it?

You said:

> So if they match, we can just skip over updating them and reuse them, right?

This implies to me a check that _if they match_ you skip, which is exactly what we have on main via `getattr(record, key) != value`.

What you're showing is that you simply ignore the state comparison thereby missing a legitimate update of the state analogous to how we do it for all other attributes except `uid`, `dtype`, and `otype`.

This means that if a user wants to set a new `default_value` that setting would _no_ longer propagate because you're always skipping the update.

That's why I think we should do what I suggested [here](https://github.com/laminlabs/lamindb/pull/2889#issuecomment-3036247420): address the root cause of the issue and starting to compare apples with apples.

Zethson:
Sorry, you're right. I had made an assumption that was off. Not sure how I suddenly went there.

In any case, here's a proper attempt (I think). In https://github.com/laminlabs/lamindb/pull/2889/commits/e4e481812fec5e8e71fa44e30917210efd37b49a we're now adding the full `_aux` and therefore the comparison later when the attributes are updated correctly skips it in such cases. Now we're comparing apples and apples right?

Hope I got it correct now...

falexwolf:
Great! This fixes the bug correctly.

The last remaining questions are stylistic.

This introduces violation of DRY. Because you're now correctly setting the `_aux` state before doing the hash lookup, you don't need the setters further down in the constructor anymore.

These are superfluous now, aren't they? If you agree, can you simply remove them, please?

* https://github.com/laminlabs/lamindb/blob/d0ba1d763855c7ccb41bb372359480b5eaf2c4f2/lamindb/models/schema.py#L573
* https://github.com/laminlabs/lamindb/blob/d0ba1d763855c7ccb41bb372359480b5eaf2c4f2/lamindb/models/schema.py#L591-L594

The other stylistic question is the manual interaction with the coding inside `_aux`. As you pointed out this is unreadable, and by duplicating it across the properties and the `validate_kwargs()` function we again violate DRY and somewhat dangerous code.

image

But given how long this took already, I'd say it's fine to postpone that refactor. One will need to introduce an abstraction beyond the properties, because the properties cannot be used inside `validate_kwargs()`. That requires some thought to organize it well.

I also remember that I had issues with using the property setters in the constructor before calling `super().__init__()`. This is the reason why some of them ended up below it. I don't understand why there are issues with this, but that'd be yet another thing to investigate. Not necessary now.

Issue #2662, opened 2025-04-11
Add Artifact kind to `.describe()`


Zethson:
### Add a description

![Image](https://github.com/user-attachments/assets/cab7b7b8-b9f4-4146-b249-3a0f0a13c2ba)

We should denote the Artifact kind in the `.describe()` output. 

Brought up by @falexwolf who should have created this issue given that this specific MLFlow run was `fearless wolf`!
https://laminlabs.slack.com/archives/C03Q5TXF797/p1744296943374149?thread_ts=1744294530.554479&cid=C03Q5TXF797

PR #2894, opened 2025-07-05
🚸 Introduce two-column layout in `Artifact.describe()` and add information like `space`, `branch`, and `kind`

falexwolf:
Before | After
--- | ---
image | image

Partially supersedes:

- https://github.com/laminlabs/lamindb/pull/2764

Fixes

- https://github.com/laminlabs/lamindb/issues/2662
- https://github.com/laminlabs/pfizer-lamin-usage/issues/202

### DIFF
--- a/lamindb/models/_describe.py
+++ b/lamindb/models/_describe.py
@@ -27,7 +27,7 @@ def highlight_time(iso: str):
         # raises ValueError: Invalid isoformat string: '<django.db.models.expressions.DatabaseDefault object at 0x1128ac440>'
         # but can't be caught with `isinstance(iso, DatabaseDefault)` for unkown reasons
         return Text("timestamp of unsaved record not available", style="dim")
-    return Text(res, style="dim")
+    return Text(res)
 
 
 # Define consistent column widths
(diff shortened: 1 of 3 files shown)


Zethson:
BTW, why are we showing some things as string like output and some not?

`.transform 'None'` looks weird.

You can also argue that it's weird that the `path` is not a string but they key is.

I think this is just for pretty output and not to copy & paste things. I think I would prefer it if we didn't output anything in quotes. Am I missing something?

falexwolf:
> BTW, why are we showing some things as string like output and some not?

Right, that should be a careful choice. The idea is to quote things that might be ambiguous to read or needed for copy & paste. For a while we had `SQLRecord.__repr__()` print lots of non-quoted stuff and that led to hiccups.

> .transform 'None' looks weird.

Yeah, that was a bug because there was no check for whether description is None. `key` is always present on a transform.

> You can also argue that it's weird that the path is not a string but they key is.

Agreed.

> I think this is just for pretty output and not to copy & paste things. I think I would prefer it if we didn't output anything in quotes. Am I missing something?

Indeed the danger of confusion/ambiguity and the need for copy & paste might be less than I thought. I think we can scratch most of these quotes, good point!

falexwolf:
Fixed:

image

- https://github.com/laminlabs/lamindb/pull/2897

PR #2904, opened 2025-07-06
🎨 Warn when no values are validated

sunnyosun:
Before this PR, the `CatManager` errored if zero valid categories were found, even if `maximal_set=False`, which should allow arbitrarily many invalid categories.

This previous behavior causes issues in cases in which users have datasets that are annotated by draft metadata.

If a user wants the error, they should instead set `maximal_set=True` for that schema. For example, if you are validating gene ids, you want to make sure at least one ID is validated, otherwise you might be validating against a wrong field.

Removed this hard erroring and replaced it with a warning instead.

### DIFF
--- a/lamindb/curators/core.py
+++ b/lamindb/curators/core.py
@@ -559,6 +559,8 @@ class DataFrameCurator(Curator):
                 ordered=schema.ordered_set,
                 index=index,
             )
+            print("pandera_columns", pandera_columns)
+            print("index", index)
         # in the DataFrameCatManager, we use the
         # actual columns of the dataset, not the pandera columns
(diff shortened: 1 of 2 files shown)


sunnyosun:
Now running into weird bugs that I can't reproduce with local tests.

Koncopd:
`unit-storage` failures are probably to running in parallel with other PRs or with local tests.

Koncopd:
oh, no, probably not that, but i have an idea.

Koncopd:
`s3://lamindb-ci/test-settings-switch-storage` is still marked by mistake, i will fix that.

PR #2917, opened 2025-07-09
🚸 Improve the experience of working in `keep-artifacts-local` mode

falexwolf:
# The `Storage` page

Before | After
--- | ---
image | 
image |

# The `keep-artifacts-local` page

Before | After
--- | ---
image | 
image | 
image |
image |

# Needs

- https://github.com/laminlabs/lamindb-setup/pull/1087

### DIFF
--- a/docs/faq/keep-artifacts-local.ipynb
+++ b/docs/faq/keep-artifacts-local.ipynb
@@ -109,7 +109,7 @@
    },
    "outputs": [],
    "source": [
-    "ln.Storage(root=\"./our_local_storage\", host=\"institute-abc-server1\").save()"
+    "ln.Storage(root=\"./our_local_storage\", host=\"abc-institute-drive1\").save()"
    ]
   },
   {
@@ -278,7 +278,7 @@
    },
    "outputs": [],
    "source": [
-    "ln.Storage(root=\"./our_local_storage2\", host=\"institute-abc-server1\").save()\n",
+    "ln.Storage(root=\"./our_local_storage2\", host=\"abc-institute-drive1\").save()\n",
     "ln.settings.local_storage = \"./our_local_storage2\"  # switch to the new storage location"
    ]
   },
--- a/docs/faq/keep-artifacts-local.ipynb
+++ b/docs/faq/keep-artifacts-local.ipynb
@@ -57,7 +57,9 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "In real use cases, you checkmark the \"keep artifacts local\" box on the instance settings tab on LaminHub. For testing, we toggle it through the following private attribute."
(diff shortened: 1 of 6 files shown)


Koncopd:
```python
ln.settings.storage_local = "./my_storage_local"
ValueError: `host` must be set for local storage locations that are registered on the hub
```
and the rest.
This looks really annoying to me.

falexwolf:
> This looks really annoying to me.

This call won't be allowed anymore which is consistent with the whole storage UX refactor. A hard error is much better than an accidental registration.

`ln.settings.storage = ...` like calls should no longer create things.

falexwolf:
This PR is not done, I have tons of local changes. No review, yet, please!

Koncopd:
Ok!

falexwolf:
> ln.settings.storage_local = "./my_storage_local"
> ValueError: `host` must be set for local storage locations that are registered on the hub
> and the rest.
> This looks really annoying to me.

I think the API for this was never well designed. The whole "switching storage locations" based on settings is probably not the best design in the world. I still don't know a much better one though. The `switch()` function could be one given that people are used to it from git.

I'm still slightly confused about where exactly the `host` argument is required and where not. Clearly, scanning the available local locations on any given compute environment works without it by looping over available roots in the instance db and matching the `storage_uid.txt` file.

But also clearly, the fact that there is no uniqueness guarantee for a storage location in the hub table other than indirectly through the storage registration mechanism, in which the `storage_uid.txt` is checked, worries me.

Finally, it's simply a UX benefit to have a host identifier for any given local storage location.

If it's really just about UX, we could relax some of the requirements of passing `host`. In particular for switching locations on a given host environment, users shouldn't be required to pass it.

Writing this I have a UX idea: one could write the host identifier into the shared compute environment root using `platformdir`. Then users wouldn't need to populate it themselves.

One user does that in the beginning and then this host has an identifier that everybody else can use.

Zethson:
> I think the API for this was never well designed. The whole "switching storage locations" based on settings is probably not the best design in the world. I still don't know a much better one though. The switch() function could be one given that people are used to it from git.

In git that's probably a remote? If we want to keep the analogy.

falexwolf:
> In git that's probably a remote? If we want to keep the analogy.

The problem is that it's indeed different. git doesn't allow you to keep things local _and_ work with a remote, either you have a remote or you don't. I mean the only way is to keep certain branches local but that doesn't seem the right thing to do for production data.

So, in lamindb, we want a way to keep certain production artifacts local without moving them onto a particular branch.

Also, lamindb allows you to track data in many local and remote storage locations, so you want to be able to switch between them.

This PR here is self-contained and improves things a lot but I think what would make sense now is to introduce `lamin switch --storage /abc/shared/` to be able to switch on the command line, and not just via `ln.settings.local_storage`.

Issue #2594, opened 2025-03-24
Improve instructions for


Zethson:
### Add a description

```python
artifact = ln.Artifact.using("scportrait/examples").filter(key__icontains=row["image_path"]).one()
artifact.cell_lines.add(bt.CellLine.filter(name=row.cell_line).one())
```

errors with

```
ValueError: Cannot add "CellLine(uid='4oLoqaUP', name='U2OS', space_id=1, created_by_id=1, run_id=2, created_at=2025-01-26 23:31:54 UTC)": instance is on database "scportrait/examples", value is on database "default"
```

We should tell people to save the artifact first to transfer it to the current instance. Then it can be labeled.

PR #2924, opened 2025-07-10
🚸 Improve UX for labeling unsaved records from other instances

Zethson:
Fixes https://github.com/laminlabs/lamindb/issues/2594

| Before | After |
| ------ | ----- |

### DIFF
--- a/lamindb/models/_django.py
+++ b/lamindb/models/_django.py
@@ -50,8 +50,8 @@ def patch_many_to_many_descriptor() -> None:
                     source_db = manager.instance._state.db
 
                     raise ValueError(
-                        f"Cannot label an object from instance '{source_db}'. "
-                        f"Please save the object first to your instance using '.save()'."
+                        f"Cannot label an record from instance '{source_db}'. "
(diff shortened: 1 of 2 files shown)


falexwolf (approved):
Great!

Just: We don't use the word "object" for records anywhere else. Please update to "record" for consistency

image

Zethson:
Yeah, I intentionally avoided `record` because `ln.Record` exists. But I guess if it's lowercase, it should be fine. Adapted.

falexwolf:
> Yeah, I intentionally avoided record because ln.Record exists. But I guess if it's lowercase, it should be fine. Adapted.

Yes, I agree that ever since introducing `SQLRecord` and `Record` side-by-side, "record" is ambiguous. The thing is that in most cases that ambiguity shouldn't matter to users. The artifact record is still the artifact SQLRecord object. You don't have to specify that it's a "SQL record" as opposed to a "flexible record".

I anticipate that when people talk about rigid pre-defined things like artifacts, transforms, and cell types, they say artifacts, transforms, and cell types. But when they talk about flexible things they say records.

PR #2925, opened 2025-07-10
🚸 Improve error message when attempting to curate unsaved Schema

Zethson:
| Before | After |
| ------ | ----- |

### DIFF
--- a/lamindb/models/_django.py
+++ b/lamindb/models/_django.py
@@ -21,12 +21,10 @@ if TYPE_CHECKING:
 def patch_many_to_many_descriptor() -> None:
     """Patches Django's `ManyToManyDescriptor.__get__` method to suggest better errors when saving relationships of an unsaved model.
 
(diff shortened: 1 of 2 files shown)


lamindb/curators/core.py
```diff
@@ -175,9 +175,18 @@ class Curator:
         - :class:`~lamindb.curators.AnnDataCurator`
         - :class:`~lamindb.curators.MuDataCurator`
         - :class:`~lamindb.curators.SpatialDataCurator`
+        - :class:`~lamindb.curators.TiledbsomaExperimentCurator`
     """
 
     def __init__(self, dataset: Any, schema: Schema | None = None):
+        if not isinstance(schema, Schema):
```
Zethson, line 182: This check also helps with the issue of people passing tuples of schemas etc. This is already tested but this makes it explicit.

Issue #2551, opened 2025-03-11
Better UX for ValidationError: expected series 'XYZ' to have type category, got object


Zethson:
### Add a description

We need to suggest `coerce_dtype` here. People don't understand this error.

To reproduce:
```python
df = pd.DataFrame(
    columns=["cell_type"],
    data=["t cell", "b cell"]
)
schema = ln.Schema(
    features=[ln.Feature(name="cell_type", dtype=ln.ULabel).save()]
).save()
curator = ln.curators.DataFrameCurator(df, schema)
curator.validate()
```

PR #2926, opened 2025-07-10
🚸 Improve data type issue errors

Zethson:
Fixes https://github.com/laminlabs/lamindb/issues/2551

Not my favorite solution but I tried a few:
1. Fully circumventing the Pandera error is possible but it might be helpful. Especially when users don't want to use `coerce_dtype`, I don't think that we should swallow it.
2. Instead of modifying the traceback, we could omit a logger message. However, the traceback is so long, I don't think that people will even see the logger message
3. Instead of appending to the same line, I had used a few new lines. The issue is that the vscode renderer then moves the essential error lines up and they're hidden.

| Before | After |
| ------ | ----- |

### DIFF
--- a/lamindb/curators/core.py
+++ b/lamindb/curators/core.py
@@ -179,6 +179,9 @@ class Curator:
     """
 
     def __init__(self, dataset: Any, schema: Schema | None = None):
+        if not isinstance(schema, Schema):
+            raise InvalidArgument("schema argument must be a Schema record.")
+
         if schema.pk is None:
             raise ValueError(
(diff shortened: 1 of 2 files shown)


falexwolf (approved):
Good!

The arg is called `coerce_dtype` and it'd be preferable if people defined it on the Feature rather than the schema-level. So, I'd make a note about this.

> ▶
I like this, we should think about using it more consistently

> to automatically convert types."
I'd say "to attempt coercing/converting values during validation to the pre-defined dtype".

PR #2928, opened 2025-07-11
🐛 Fix deletion of cloud transfer local instance

Zethson:
See https://c222b12c.lamindb.pages.dev/storage/transfer-local-to-cloud

Done via Github editor.


Koncopd:
Thanks, this makes sense. On the other hand it is not a problem, because the instance if fully local and is not registered in the hub (i hope).

Issue #2931, opened 2025-07-11
`track()` errors for anonymous users


Zethson:
### Add a description

https://github.com/theislab/single-cell-best-practices/issues/384

To reproduce:

```python
!lamin connect theislab/sc-best-practices
import lamindb as ln
ln.track()
```

leads to

```
---------------------------------------------------------------------------
DoesNotExist                              Traceback (most recent call last)
File ~/code/lamindb/lamindb/base/users.py:16, in current_user_id.<locals>.query_user_id()
     15 try:
---> 16     user_id = User.objects.get(uid=settings.user.uid).id
     17 except User.DoesNotExist:

File ~/miniforge3/envs/lamindb/lib/python3.12/site-packages/django/db/models/manager.py:87, in BaseManager._get_queryset_methods.<locals>.create_method.<locals>.manager_method(self, *args, **kwargs)
     85 @wraps(method)
     86 def manager_method(self, *args, **kwargs):
---> 87     return getattr(self.get_queryset(), name)(*args, **kwargs)

File ~/miniforge3/envs/lamindb/lib/python3.12/site-packages/django/db/models/query.py:649, in QuerySet.get(self, *args, **kwargs)
    648 if not num:
--> 649     raise self.model.DoesNotExist(
    650         "%s matching query does not exist." % self.model._meta.object_name
    651     )
    652 raise self.model.MultipleObjectsReturned(
    653     "get() returned more than one %s -- it returned %s!"
    654     % (
   (...)
    657     )
    658 )

DoesNotExist: User matching query does not exist.

During handling of the above exception, another exception occurred:

DoesNotExist                              Traceback (most recent call last)
Cell In[3], line 1
----> 1 ln.track()

File ~/code/lamindb/lamindb/core/_context.py:448, in Context._track(self, transform, project, space, branch, params, new_run, path)
    446     if description is None:
    447         description = self._path.name
--> 448     self._create_or_load_transform(
    449         description=description,
    450         transform_ref=transform_ref,
    451         transform_ref_type=transform_ref_type,
    452         transform_type=transform_type,  # type: ignore
    453     )
    454 else:
    455     if transform.type in {"notebook", "script"}:

File ~/code/lamindb/lamindb/core/_context.py:827, in Context._create_or_load_transform(self, description, transform_ref, transform_ref_type, transform_type)
    825 if transform is None:
    826     assert key is not None  # noqa: S101
--> 827     transform = Transform(  # type: ignore
    828         uid=self.uid,
[...]

PR #2934, opened 2025-07-11
🐛 Properly ignore `ln.track()` and tracking warnings on read-only connections

Koncopd:
Fix https://github.com/laminlabs/lamindb/issues/2931


Zethson:
Could we have a test for anonymous access or the different right combinations, please?

Issue #2918, opened 2025-07-09
Should not check hash when creating a Schema type


sunnyosun:
### Add a description

<img width="711" height="674" alt="Image" src="https://github.com/user-attachments/assets/b150a14d-c63d-4dec-b24d-7608fe8d478e" />

PR #2937, opened 2025-07-14
🐛 Fix creation of a `Schema` with `is_type=True`

sunnyosun:
Resolves: https://github.com/laminlabs/lamindb/issues/2918


falexwolf (approved):
Looks good!

PR #2938, opened 2025-07-15
♻️ More efficient tests cleanup

sunnyosun:


tests/curators/test_cxg_curator.py
```diff
@@ -3,81 +3,46 @@
...
-
+    # fix typo in obs column
     adata.obs["tissue"] = adata.obs["tissue"].cat.rename_categories({"lungg": "lung"})
-
-    curator.validate()
-
     artifact = curator.save_artifact(key="examples/dataset-curated-against-cxg.h5ad")
 
-    # Clean up
-    artifact.delete(permanent=True)
```
Zethson, line 65: I understand that this is more efficient because we only query and delete once. However, then this test isn't self contained anymore. If you were to put the second test above this one, we'd have leaking objects, right?

I thought we opt for self-contained tests that also clean up after themselves?

sunnyosun: why not? A cleanup is done for all records in the registry a test touches, it has been done like this for all other tests.

If all tests are done this way, there should be no leakage of objects.

Your code was also cleaning up everything, just one by one:
Screenshot 2025-07-15 at 11 45 11

I'll move the schema into a fixture.

Zethson: Ohhhh, I had missed that you integrated `test_missing_obs_cols` into `test_cxg_curator`. 

I think it's better if we have smaller tests where it's clear from the function title what they test but I'm fine with that.

PR #2943, opened 2025-07-15
📝 Add an actual README

falexwolf:


README.md
```diff
@@ -4,4 +4,79 @@
...
+
+<!-- py-quickstart.py -->
+
+Here's how to create an artifact while tracking source code, run environment, run logs, and inputs and outputs of a script or notebook.
+
+```python
+import lamindb as ln
+
+ln.track()  # track the run of a script or notebook
+open("sample.fastq", "w").write("@r1\nACGT\n+\nIIII\n")
```
Zethson: Looks scary. Maybe a fasta is good enough and friendlier.

README.md
```diff
@@ -4,4 +4,79 @@
 
 # LaminDB - A data framework for biology
 
-Read the [docs](https://docs.lamin.ai).
+LaminDB is an open-source data framework to enable learning at scale in computational biology.
+It lets you track data transformations, curate datasets, manage metadata, and query a built-in database for biological entities & data structures.
```
Zethson, line 8: Should we have some bullet points of features below? 

What you're proposing is great to give people a glimpse of what we have but we're never selling that LaminDB is SOOO much more. If we can find a way to do that without overwhelming people, it might not be bad.

Zethson (commented):
Generally great!

I will die on that hill, but I still find it hard to discern Lamin concepts and objects from everything else when everything is lowercase.

README.md
falexwolf: I had the same thought but thought `.fasta` is so small scale that people will often want to put this into a table instead of its own file. But we also use it in the redun example so I guess it'd be fine.

This is how it'd look for fasta

```
open("sample.fasta", "w").write(">seq1\nACGT\n")
```

README.md
falexwolf: Yes, I agree.

How about we render the lamindb specs at the bottom? Need to be updated anyway and could be part of this repo then.

image

And in lamin-docs we can import from the readme.

falexwolf: Well, all the links won't work in the readme unless we replace them with hardcoded URLs, which is maybe fine these days because URL structure really doesn't change much anymore.

falexwolf: One could also call it "LaminDB features" again.

Zethson: I was actually thinking of for something much much more condensed and high level. 5 bullet points max.

falexwolf: I see two problems with that:

- I'm always worried about duplicating the same information in slightly different ways in different places, and we've also gotten user feedback about this
- 5 bullets is the same depth than the 2 sentence prose that we already have at the top, meaning we'd just cast prose to bullets

So, I think either we take the one & only official "LaminDB features" list or we don't add a features list.

README.md
falexwolf: Used that! Sunny also thought fasta would be good.

README.md
falexwolf: Will take care of this in another PR because it involves moving things in lamin-docs.

PR #2957, opened 2025-07-17
📝 Refactor curate guide

sunnyosun:
Before | After
--- | ---
Screenshot 2025-07-21 at 11 22 25 | Screenshot 2025-07-21 at 11 22 46
Screenshot 2025-07-21 at 11 23 38 | Screenshot 2025-07-21 at 11 23 59
NA | Screenshot 2025-07-21 at 11 26 13
NA | Screenshot 2025-07-21 at 11 26 41
Screenshot 2025-07-21 at 11 27 15 | Screenshot 2025-07-21 at 11 27 33
NA | Screenshot 2025-07-21 at 11 28 16
NA | Screenshot 2025-07-21 at 11 28 45
NA | Screenshot 2025-07-21 at 11 29 04


falexwolf:
I'm sure this is great! Just, it's impossible to see what and why it changed.

Could you write a brief summary of rationales and add Before and After screenshots?

falexwolf:
image

Two comments here:

1. I'd not say "messy, real-world data" because the guide is full of synthetic examples. I also don't think we have to stress "messiness" because validation is also for internal data.
2. I think the emojis break style because we don't use them anywhere else. If we want to consistently adopt them, sure, but otherwise I think it's prettier without.
3. The sentences in the enumeration are so long that maybe a period at the end wouldn't hurt now.

falexwolf:
image

One could add: "similar to `pydantic.Model` for dictionaries, and `pandera.Schema`, and `pyarrow.lib.Schema` for tables, but supporting more complicated data structures."

falexwolf:
I'd say:

image

What Features (dimensions) exist in your dataset

falexwolf:
image

I'd say gene-derived features.

PR #2964, opened 2025-07-21
🎨 Polish curate guide

Zethson:
Minor wording & style fixes.

- Markdown sentences should always be on their own line for easier diffs and reviews
- `feature` is such an overloaded term. Sometimes a sphinx ref might help. It's not always obvious when to add it and when not. I conservatively added one more which I think helps.
- Fixed casing of some words

### DIFF
--- a/docs/curate.ipynb
+++ b/docs/curate.ipynb
@@ -57,13 +57,14 @@
    "source": [
     "## Schema design patterns\n",
     "\n",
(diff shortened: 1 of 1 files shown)


sunnyosun (approved):

falexwolf:
This is all great with the exception of one change, IMO.

Here the added Sphinx ref does not make sense, IMO.

image

Feature was defined in the first bullet through the sphinx ref. All following bullets make reference to that definition when using the term "features" so that the user doesn't have to deal with the bold (visually noisy) Sphinx ref again.

_Just_ changing "feature" to the Sphinx ref in the last bullet is both inconsistent (why not in the 2nd and 3rd bullet) and introduces that visual noise that was meant to be avoided.

But this is minor, so, let's just keep it as is.

PR #2972, opened 2025-07-23
🚸 Adapt batch registration logging to be debug level

Zethson:
The batch registration logging can clutter stdout quite a bit.

Advantages:
- Transparent behavior
- Kind of acts like a progress bar when registering large files

Disadvantages:
- Clutters stdout
- Is not a progress bar and takes up several lines

To me, the disadvantages outweigh the advantages. Therefore, I suggest that we make the logging messages debug level messages instead.

### DIFF
--- a/lamindb/models/save.py
+++ b/lamindb/models/save.py
@@ -51,7 +51,7 @@ def save(
             unique or another constraint. However, it won't inplace update the id
             fields of records. If you need records with ids, you need to query
             them from the database.
-        batch_size: Number of records to process in each batch. Defaults to 10000.
+        batch_size: Number of records to process in each batch.
             Large batch sizes can improve performance but may lead to memory issues.
 
     Examples:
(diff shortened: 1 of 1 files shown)


sunnyosun:
I disagree, due to the previous server traffic issues, it acts as a warning for users to know they are adding 1M records and have a chance to interrupt. If they don't see it, they will wait forever for the run to finish. This was a conscious decision to be on the warning level.

Zethson:
OK, then how about we keep these

```python
logger.warning(
                f"Starting bulk_create for {total_records} {model_name} records in batches of {batch_size}"
            )
```

at warning level but move the "processing batch xyz" messages to debug?

sunnyosun:
I also don't see a point of moving info logging to debug, since info is by default not shown and I often set verbosity to "hint" for exploratory notebooks (also suggested for users).

Zethson:
> I also don't see a point of moving info logging to debug, since info is by default not shown and I often set verbosity to "hint" for exploratory notebooks.

I don't think that these 

```python
logger.info(
                    f"Processing batch {batch_num}/{total_batches} for {model_name}: {len(batch)} records"
                )
```

are useful on the info level. We can also move them to info.

Or do nothing at all if you have a strong opinion here.

sunnyosun:
> > I also don't see a point of moving info logging to debug, since info is by default not shown and I often set verbosity to "hint" for exploratory notebooks.
> 
> I don't think that these
> 
> ```python
> logger.info(
>                     f"Processing batch {batch_num}/{total_batches} for {model_name}: {len(batch)} records"
>                 )
> ```
> 
> are useful on the info level. We can also move them to info.
> 
> Or do nothing at all if you have a strong opinion here.

It's important when I see things are progressing, otherwise it's a call that takes minutes without feedback. I just don't see a significant advantage to move them to debug since it's already hidden.

falexwolf:
I think these are only printed these days in case of >10k records. I made a PR some time ago that led to this no longer being printed all the time.

That's a rare operation and quite a load on the db, so I'd also like to see what's happening. We had an issue where somebody tried to ingest 2M at a time without realizing it, causing problems.

Hence, I think we want to keep the current way.

Let me know if I'm getting something wrong!

Issue #2974, opened 2025-07-23
`collection.describe()` returns empty string


Zethson:
### Add a description

Noticed while working on CELLxGENE registration. For example:

```python
ln.Collection.using("laminlabs/cellxgene").get("VnqB6Y76ePP6wuaFzNyR").describe()
```

returns an empty string.

PR #2975, opened 2025-07-23
✨ Add Collection.describe()

Zethson:
Fixes https://github.com/laminlabs/lamindb/issues/2974

| Before | After |
| ------ | ----- |

## Technical details

- Adds Collection specific fetch and describe code
- Adds more thorough tests for `Artifact.describe()` and `Collection.describe()`

### DIFF
--- a/lamindb/models/_describe.py
+++ b/lamindb/models/_describe.py
@@ -127,61 +127,161 @@ def format_bytes(bytes_value):
         return f"{bytes_value / (1024**4):.1f} TB"
 
 
-def describe_dataset(
-    obj: Artifact | Collection,
+def describe_artifact_general(
+    self: Artifact,
     tree: Tree | None = None,
     foreign_key_data: dict[str, dict[str, int | str]] | None = None,
 ) -> Tree:
-    from lamindb.models import Artifact
-
     if tree is None:
-        tree = describe_header(obj)
+        tree = describe_header(self)
 
+    # add general information (order is the same as in API docs)
     general = tree.add(Text("General", style="bold bright_cyan"))
 
-    if obj.key:
-        general.add(Text.assemble(("key: ", "dim"), (f"{obj.key}", "cyan3")))
-    if obj.description:
+    if self.key:
+        general.add(Text.assemble(("key: ", "dim"), (f"{self.key}", "cyan3")))
+    if self.description:
         general.add(
             Text.assemble(
                 ("description: ", "dim"),
-                f"{obj.description}",
+                f"{self.description}",
             )
         )
 
+    # Two column items (short content)
     two_column_items = []
-    two_column_items.append(Text.assemble(("uid: ", "dim"), f"{obj.uid}"))
 
-    if isinstance(obj, Artifact):
-        two_column_items.append(Text.assemble(("hash: ", "dim"), f"{obj.hash}"))
-        two_column_items.append(
-            Text.assemble(("size: ", "dim"), f"{format_bytes(obj.size)}")
-        )
-
+    two_column_items.append(Text.assemble(("uid: ", "dim"), f"{self.uid}"))
+    two_column_items.append(Text.assemble(("hash: ", "dim"), f"{self.hash}"))
+    two_column_items.append(
+        Text.assemble(("size: ", "dim"), f"{format_bytes(self.size)}")
+    )
     transform_key = (
         foreign_key_data["run"]["transform_key"]
-        if foreign_key_data and isinstance(obj, Artifact)
-        else foreign_key_data["transform"]["name"]
-        if foreign_key_data and "transform" in foreign_key_data
-        else obj.transform.key
-        if hasattr(obj, "transform") and obj.transform and isinstance(obj, Artifact)
-        else obj.transform.name
-        if hasattr(obj, "transform") and obj.transform
+        if foreign_key_data
+        else self.transform.key
+        if self.transform
         else None
     )
-    if transform_key:
+    two_column_items.append(
+        Text.assemble(
+            ("transform: ", "dim"),
+            (f"{transform_key}", "cyan3"),
+        )
+    )
+    space_name = (
+        foreign_key_data["space"]["name"] if foreign_key_data else self.space.name
+    )
+    two_column_items.append(Text.assemble(("space: ", "dim"), space_name))
+    branch_name = (
+        foreign_key_data["branch"]["name"] if foreign_key_data else self.space.name
+    )
+    two_column_items.append(Text.assemble(("branch: ", "dim"), branch_name))
+    # actually not name field here, but handle
+    created_by_handle = (
+        foreign_key_data["branch"]["name"]
+        if foreign_key_data
+        else self.created_by.handle
+    )
+    two_column_items.append(
+        Text.assemble(
+            ("created_by: ", "dim"),
+            (created_by_handle),
+        )
+    )
+    two_column_items.append(
+        Text.assemble(("created_at: ", "dim"), highlight_time(str(self.created_at)))
+    )
+    if self.n_files:
+        two_column_items.append(Text.assemble(("n_files: ", "dim"), f"{self.n_files}"))
+    if self.n_observations:
+        two_column_items.append(
+            Text.assemble(("n_observations: ", "dim"), f"{self.n_observations}")
+        )
+    if self.version:
+        two_column_items.append(Text.assemble(("version: ", "dim"), f"{self.version}"))
+
+    # Add two-column items in pairs
+    for i in range(0, len(two_column_items), 2):
+        if i + 1 < len(two_column_items):
+            # Two items side by side
+            left_item = two_column_items[i]
+            right_item = two_column_items[i + 1]
+
+            # Create padded version by calculating the plain text length
+            left_plain_text = (
+                left_item.plain if hasattr(left_item, "plain") else str(left_item)
+            )
+            padding_needed = max(0, 35 - len(left_plain_text))
+            padding = " " * padding_needed
+
+            general.add(Text.assemble(left_item, padding, right_item))
+        else:
+            # Single item (odd number)
+            general.add(two_column_items[i])
+
+    storage_root = (
+        foreign_key_data["storage"]["name"] if foreign_key_data else self.storage.root
+    )
+    storage_key = self.key if self.key else self.uid
+    if not self.key and self.overwrite_versions > 0:
+        storage_key = storage_key[:-4]
+    general.add(
+        Text.assemble(
+            ("storage path: ", "dim"),
+            (storage_root, "cyan3"),
+            f"/{storage_key}",
+        )
+    )
+    return tree
+
+
+def describe_collection_general(
+    self: Collection,
+    tree: Tree | None = None,
+    foreign_key_data: dict[str, dict[str, int | str]] | None = None,
+) -> Tree:
+    if tree is None:
+        tree = describe_header(self)
+
+    # add general information (order is the same as in API docs)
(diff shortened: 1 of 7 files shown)


lamindb/models/_describe.py
```diff
@@ -212,7 +212,7 @@ def describe_artifact_general(
             left_plain_text = (
                 left_item.plain if hasattr(left_item, "plain") else str(left_item)
             )
-            padding_needed = max(0, 35 - len(left_plain_text))
+            padding_needed = max(0, 45 - len(left_plain_text))
```
Zethson: 35 looks better but I ran into formatting issues for long transform names. 40 was borderline not enough. Thoughts? There might be other ways to solve this issue of course but this is the "quick fix".

lamindb/models/_describe.py
```diff
@@ -127,81 +127,113 @@ def format_bytes(bytes_value):
         return f"{bytes_value / (1024**4):.1f} TB"
 
 
-def describe_artifact_general(
-    self: Artifact,
+def describe_dataset(
```
Zethson: I'm open to better names that catch both Artifact and Collection but nothing else. I couldn't really think of something better. I can then also name the parameter `obj` just below accordingly to make it even clearer.

lamindb/models/_describe.py
```diff
@@ -212,25 +244,30 @@ def describe_artifact_general(
             left_plain_text = (
                 left_item.plain if hasattr(left_item, "plain") else str(left_item)
             )
-            padding_needed = max(0, 35 - len(left_plain_text))
+            padding_needed = max(0, 45 - len(left_plain_text))
```
Zethson, line 334: 35 looks better but I ran into formatting issues for long transform names. 40 was borderline not enough. Thoughts? There might be other ways to solve this issue of course but this is the "quick fix".

Zethson: Actually, I think after refactoring I can probably revert this to 35.

lamindb/models/_describe.py
```diff
@@ -127,81 +127,113 @@ def format_bytes(bytes_value):
         return f"{bytes_value / (1024**4):.1f} TB"
 
 
-def describe_artifact_general(
-    self: Artifact,
+def describe_dataset(
+    obj: Artifact | Collection,
```
falexwolf: I think you're reverting quite a bit of my change from a few days ago.

image

- https://github.com/laminlabs/lamindb/pull/2953

It used to accept both Artifact & Collection, but that led to all kinds of problems.

Rationale was: Given collection doesn't have its own metadata and in that sense is like every other sqlrecord I thought we need a proper solution for artifact, and then a generic solution for all other sqlrecords.

I'm looking through the code now but - for this reason - "philosophically" and operationally this doesn't seem great. 😅

Zethson: I had a custom implementation for Collection but then opted to merge Artifact and Collection because the code overlapped a lot. I can disentangle that.

> then a generic solution for all other sqlrecords.

wasn't called correctly in this case because it returned empty strings. I can see whether there's a way to fix that but having an implementation here that outputs a similar design as `Artifact` is also great.

Zethson: > I think you're reverting quite a bit of my change from a few days ago.

Evidently bad. After your general feedback, I'll look again to ensure that I don't do that.

falexwolf: Got distracted, now back here.

lamindb/models/_describe.py
```diff
@@ -127,81 +127,113 @@ def format_bytes(bytes_value):
...
     tree: Tree | None = None,
     foreign_key_data: dict[str, dict[str, int | str]] | None = None,
```
falexwolf: The main thing is that this dict is quite hard to abstract across all sqlrecords at the current state of the code base. But it's absolutely needed for performance reasons.

lamindb/models/_describe.py
```diff
@@ -127,81 +127,113 @@ def format_bytes(bytes_value):
...
     transform_key = (
         foreign_key_data["run"]["transform_key"]
-        if foreign_key_data
-        else self.transform.key
-        if self.transform
+        if foreign_key_data and isinstance(obj, Artifact)
+        else foreign_key_data["transform"]["name"]
+        if foreign_key_data and "transform" in foreign_key_data
+        else obj.transform.key
+        if hasattr(obj, "transform") and obj.transform and isinstance(obj, Artifact)
```
falexwolf: A main goal of the previous refactor was to get rid of the non-idiomatic `hasattr(obj, "transform")` calls. Imagine you'd have to do this for every type of sqlrecord.

And in addition to this being non-idiomatic it conceals performance considerations; Django makes database requests upon these checks. I think devs can learn this when they see `record.created_by` (because this is well known for Django), but for `hasattr(record, created_by)` it really surprised even me (shouldn't have) after multiple years of Django.

Zethson: Good to know. Implies to me that it'll be difficult to have a single implementation then and maybe having two is the better choice.

lamindb/models/_describe.py
falexwolf: I went through all changes to `lamindb/models/_describe.py` and I think they're adding lots of complexity for little value (just `collection.describe()` doesn't bring us much) and at the cost of much danger: the performance aspect of the code will be very hard to keep under control and we might readily run into 8 sec `artifact.describe()` calls again.

So, I'd keep artifact isolated for now. I don't mind another function for collection that would hopefully leverage a smart abstraction re the rich formatting so that we don't duplicate everything.

Destination has to be `sqlrecord.describe()`, but I don't think many people care about this right now.

lamindb/models/collection.py
```diff
@@ -232,7 +227,7 @@ class Meta(SQLRecord.Meta, IsVersioned.Meta, TracksRun.Meta, TracksUpdates.Meta)
     @overload
     def __init__(
         self,
-        artifacts: list[Artifact],
+        artifacts: Artifact | Sequence[Artifact],
```
falexwolf: I don't think we necessarily support `Sequence`.

lamindb/models/artifact.py
```diff
@@ -730,20 +733,27 @@ def _describe_postgres(self):  # for Artifact & Collection
...
-        tree = describe_artifact_general(self, foreign_key_data=related_data["fk"])
+        tree = describe_dataset(self, foreign_key_data=related_data["fk"])
         return describe_features(
             self,
             tree=tree,
             related_data=related_data,
             with_labels=True,
         )
+    elif model_name == "Collection":
+        result = get_artifact_with_related(self, include_fk=True, include_m2m=True)
```
falexwolf: Already the function name implies that this is a shady thing to do. I'm sure we can work out a generally useful version of prefech related (and it's not so far from it), but right now there is a LOT of artifact specific code in it.

So, without extensive testing of the content of that `result`, this is dangerous.

falexwolf:
We also have a custom implementation of schema.describe() that should stylistically not be duplicated (two-column layout)

image

FYI @sunnyosun

lamindb/models/collection.py
Zethson: I think `Sequence` should be correct here and not cause issues.

```python
artifacts: Artifact | Iterable[Artifact] = (
            kwargs.pop("artifacts") if len(args) == 0 else args[0]
        )
```

used to be in the code. I can also type it as a more defensive Iterable.

falexwolf: Yes, `Iterable` is wrong. But `Sequence` might be incorrect too. We've only tested this with lists.

Zethson:
@falexwolf I think we should make a decision on how much time and effort I should spend on this.

1. I think that `Collection.describe()` is kind of important. It just returning an empty string feels bad.
2. The PR currently "makes it work" but there's lots of code duplication and some code is in the wrong places. It should be easy to merge though and won't lead to regressions.
3. I can spend substantially more time to disentangle the concerns (too much code lives in `Artifact` even before this PR) and DRY things. Probably 0.5 to 1 day of work.
4. I can disentangle and DRY everything in a bigger refactor and add a `SQLRecord.describe()`. My guess is that having a `Collection.describe()` which looks similar to `Artifact.describe()` is still a good idea. Therefore, I wouldn't know whether we'd keep such an implementation anyways. My estimate is that this would be 2 days of work.

I am leaning towards 2 and following this up with 4 in a month or so when more urgent customer related requests are done.

What do you think?

sunnyosun:
I'm not sure `collection.describe()` is so important since it only displays the basic metadata that one can also see from the record representation?

The artifact and schema describe are important because they fetch many deeply linked information that users otherwise difficult to see without queries.

I'm also not a fan of so much code duplication here since artifact has the same session to display simple metadata.

sunnyosun:
> @falexwolf I think we should make a decision on how much time and effort I should spend on this.
> 
> 1. I think that `Collection.describe()` is kind of important. It just returning an empty string feels bad.
> 2. The PR currently "makes it work" but there's lots of code duplication and some code is in the wrong places. It should be easy to merge though and won't lead to regressions.
> 3. I can spend substantially more time to disentangle the concerns (too much code lives in `Artifact` even before this PR) and DRY things. Probably 0.5 to 1 day of work.
> 4. I can disentangle and DRY everything in a bigger refactor and add a `SQLRecord.describe()`. My guess is that having a `Collection.describe()` which looks similar to `Artifact.describe()` is still a good idea. Therefore, I wouldn't know whether we'd keep such an implementation anyways. My estimate is that this would be 2 days of work.
> 
> I am leaning towards 2 and following this up with 4 in a month or so when more urgent customer related requests are done.
> 
> What do you think?

I don't think merging a PR that known to decrease the code base quality is a good idea, I'd not vote for 2. Instead I'll suggest not merge this PR.

falexwolf:
Sunny wrote at the same time 😆 

> I'm not sure collection.describe() is so important

I agree it's not important because Collection is like every other SQLRecord.

> I'm also not a fan of so much code duplication here

Agreed that the duplication isn't great but I still think here it could be fine because we're also duplicating it for schema and the way it's done here seems fine code quality wise. I see a rather straight forward way to refactoring this in the future.

The original version of this PR where `describe_artifact_general()` was reused for collection was a No-Go for me because it was terribly complex etc. (I commented on this).

falexwolf:
@sunnyosun, returning an empty string in `collection.describe()` is evidently not good. So that should be resolved -- I thought that it'd only entail fixing something in the header but now we have a path to study how we want to converge towards a `sqlrecord.describe()` in the future. So, that's helpful, isn't it?

sunnyosun:
> I agree it's not important because Collection is like every other SQLRecord.

Then why do we need to introduce collection.describe() if user can just read the SQLRecord representation? I don't see a benefit of 200 lines of code that doesn't give additional information.

The main reason that artifact.describe() and schema.describe() exist is not because of the general session, but the remaining sessions that are otherwise a few queries deep down.

If we want to introduce a general `sqlrecord.describe()`, that I agree! And I also think the schema.general session should be refactored to reused the same code as artifact.describe. But now this PR adds a 3rd non-reuseable piece of code to do the same thing as artifact and schema.

Zethson:
> If we want to introduce a general sqlrecord.describe(), that I agree! And I also think the schema.general session should be refactored to reused the same code as artifact.describe. But now this PR adds a 3rd non-reuseable piece of code to do the same thing as artifact and schema.

OK so you're suggesting that we go for option 4 (of what I outlined above).

We can also keep this PR on hold and after my more urgent priorities, I can go all in and refactor this end-to-end. The only cost is that `collection.describe()` has unexpected behavior until I get to that.

I'm fine with whatever both of you agree on :smile:

sunnyosun:
OK, then merge this so collection.describe() works.

Issue #2979, opened 2025-07-24
Support more organisms for CxG Schema


Zethson:
### Add a description

- [ ] Sars-2
- [ ] Synthetic construct
- [ ] ???

sunnyosun:
Any custom source should have been well supported via:

```python

source = bt.Gene.add_source("custom", version="1.0", organism="new-org", df=df)
```

See examples: https://docs.lamin.ai/bionty.core.biorecord#bionty.core.BioRecord.add_source

It will behave like a static reference.

I don't think we need any custom code for this in lamindb, because it will increase maintenance once the organism list expands. But rather, point users to how to register the source in the cxg curate guide.

Zethson:
Correct but the Literal here https://github.com/laminlabs/lamindb/blob/main/lamindb/examples/cellxgene/_cellxgene.py#L127 should have all supported CxG organisms. That's what this issue is referring to.

PR #2980, opened 2025-07-24
✨ Support more CxG Organisms & ✨ Allow AnnDataAccessor for curators

Zethson:
Fixes https://github.com/laminlabs/lamindb/issues/2979

- Adds all supported organisms by CELLxGENE to the typehint
- `is_scverse_datastructure` now recognizes AnnDataAccessor objects as AnnData

### DIFF
--- a/lamindb/examples/cellxgene/_cellxgene.py
+++ b/lamindb/examples/cellxgene/_cellxgene.py
@@ -8,6 +8,16 @@ from lamindb.models import Feature, Schema, SQLRecord, ULabel
 from lamindb.models._from_values import _format_values
 
 CELLxGENESchemaVersions = Literal["4.0.0", "5.0.0", "5.1.0", "5.2.0", "5.3.0"]
+CELLxGENESupportedOrganisms = Literal[
+    "human",
+    "mouse",
+    "SARS-COV-2",
+    "synthetic construct",
+    "zebra danio",
+    "rhesus macaquedomestic pig",
+    "chimpanzee",
+    "white-tufted-ear marmoset",
+]
 FieldType = Literal["ontology_id", "name"]
 
 
@@ -124,7 +134,7 @@ def get_cxg_schema(
     schema_version: CELLxGENESchemaVersions,
     *,
     field_types: FieldType | Collection[FieldType] = "ontology_id",
-    organism: Literal["human", "mouse"] = "human",
+    organism: CELLxGENESupportedOrganisms = "human",
 ) -> Schema:
     """Generates a :class:`~lamindb.Schema` for a specific CELLxGENE schema version.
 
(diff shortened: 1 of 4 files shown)


lamindb/curators/core.py
```diff
@@ -1040,9 +1062,12 @@ def __init__(
         self._maximal_set = maximal_set
 
         self._all_filters = {"source": self._source, "organism": self._organism}
+
         if self._subtype_str and "=" in self._subtype_str:
             self._all_filters.update(
-                resolve_relation_filters(parse_filter_string(self._subtype_str), self)  # type: ignore
```
Zethson, line 1047: Now uses the right code path but shouldn't have any user facing changes. `resolve_relation_filters` also expects a SQLRecord and not a CatVector.

lamindb/examples/cellxgene/_cellxgene.py
```diff
@@ -143,6 +169,11 @@ def get_cxg_schema(
 
     from lamindb.models import Feature, Schema, ULabel
 
+    # Attempt to find the Schema early as building the Schema is expensive when looped
```
Zethson: There's probably a way to reconstruct/find the schema by querying the component schemas but I think that this is kind of Occam's razor.

lamindb/examples/cellxgene/_cellxgene.py
```diff
@@ -45,16 +54,19 @@ def save_cxg_defaults() -> None:
...
+    for biorecord, name in zip(
         [
+            bt.Ethnicity,
             bt.Ethnicity,
             bt.DevelopmentalStage,
             bt.Phenotype,
             bt.CellType,
         ],
-        ["na", "unknown", "unknown", "unknown"],
+        ["na", "unknown", "unknown", "unknown", "unknown"],
```
Zethson: According to the spec, it should only be "na" but they have both. We also had both before. So this change restores the behavior.

Zethson:
Sorry @sunnyosun this PR covers two things again but that's because I had used it for CxG registration where I needed both (to test things locally). I can split it up into two PRs but since the CxG changes aren't that big here, I hope it's fine. If not, please let me know and I can disentangle it.

tests/curators/test_curators_general.py
```diff
@@ -658,3 +658,21 @@ def test_index_feature_exclusion_from_categoricals(df):
     # clean up
     schema.delete(permanent=True)
     ln.Feature.filter().delete()
+
+
+def test_artifact_standardize_errors(df):
```
sunnyosun, line 663: Why is this test needed if we already have the same check in test_cat_managers.py above?

sunnyosun (approved):
Looks good otherwise, thank you!

tests/curators/test_curators_general.py
Zethson: I assume that you're referring to https://github.com/laminlabs/lamindb/blob/dc94714221ba33fc7ff307889c0974e0f299c0de/tests/curators/test_cat_managers.py#L263 which tests legacy API. I think that a clearer new and better scoped test is cleaner. Note that it also uses a different (old) exception message.

PR #2987, opened 2025-07-27
⚡️ Generate GIN indexes

falexwolf:
All credit to @chaichontat.

- https://github.com/laminlabs/bionty/pull/307
- https://github.com/laminlabs/lamindb-setup/pull/1114

~Remaining issue is that this unnecessarily runs upon import time.~ Given this is a general problem with Django we'll find another PR to reduce this nuisance to the import experience. 

<details><summary>No impact on import time</summary>
<p>

image

image

</p>
</details> 

Squashing also remains to be done. But given this makes it a bit more complicated we'll wait until all comprehensive global search tests on the hub ran through.

To enable:

- https://github.com/laminlabs/laminhub/pull/3084


chaichontat:
The index name cannot exceed 30 chars, you'll probably have to remove some parts of the name.

You'll also need this as part of the migration.

```
from django.contrib.postgres.operations import TrigramExtension
from django.db import connection, migrations


def apply_trigram_extension(apps, schema_editor):
    if connection.vendor == "postgresql":
        TrigramExtension().database_forwards(apps, schema_editor, None, None)


def revert_trigram_extension(apps, schema_editor):
    if connection.vendor == "postgresql":
        TrigramExtension().database_backwards(apps, schema_editor, None, None)


class Migration(migrations.Migration):
    dependencies = []

    operations = [
        migrations.RunPython(apply_trigram_extension, revert_trigram_extension)
    ]

```

falexwolf:
I see, maybe I'm exceeding the 30 char limit. 🤔

falexwolf:
No, there is another problem beyond the 30 chars and installing the extension

PR #2992, opened 2025-07-29
🚚 Move datasets from core to examples

falexwolf:


Koncopd:
You can ignore the permissions test for now, i will fix it in a separate PR.

Issue #2951, opened 2025-07-16
CxG Schema 6.0.0 support


Zethson:
### Add a description

Things I learned so far:

1. New versions:
```
6.0.0,CellType,all,cl,2025-04-10
6.0.0,ExperimentalFactor,all,efo,3.78.0
6.0.0,Ethnicity,human,hancestro,3.0
6.0.0,DevelopmentalStage,human,hsapdv,2025-01-23
6.0.0,DevelopmentalStage,mouse,mmusdv,2025-01-23
6.0.0,Disease,all,mondo,2025-05-06
6.0.0,Organism,all,ncbitaxon,2024-11-25
6.0.0,Phenotype,all,pato,2025-05-14
6.0.0,Tissue,all,uberon,2025-05-28
6.0.0,Gene,human,ensembl,release-110
6.0.0,Gene,mouse,ensembl,release-110
```
2. `organism_ontology_term_id` moved to `uns` -> requires code changes to check `uns` instead.
3. Changelog is here: https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/6.0.0/schema.md#schema-v600

Zethson:
As of 6.0.0, disease now also allows multiple values. They've already been using it for a while but it wasn't in the spec.

PR #3000, opened 2025-07-30
✨ Add support for CELLxGENE Schema 6.0.0

Zethson:
Fixes https://github.com/laminlabs/lamindb/issues/2951

- Adds support for CELLxGENE Schema 6.0.0

### DIFF
--- a/tests/curators/test_cxg_curator.py
+++ b/tests/curators/test_cxg_curator.py
@@ -116,7 +116,7 @@ def test_cxg_curator_6_spatial(cxg_schema_factory):
 
 def test_invalid_field_type():
     with pytest.raises(ValueError) as e:
-        ln.examples.cellxgene.get_cxg_schema(
+        ln.examples.cellxgene.create_cellxgene_schema(
             schema_version="5.3.0", field_types=["ensembl_gene_ids"]
         )
     assert "Invalid field_types" in str(e.value)
@@ -124,5 +124,5 @@ def test_invalid_field_type():
 
 def test_invalid_schema_Version():
     with pytest.raises(ValueError) as e:
-        ln.examples.cellxgene.get_cxg_schema(schema_version="200.0.0")
+        ln.examples.cellxgene.create_cellxgene_schema(schema_version="200.0.0")
     assert "Invalid schema_version" in str(e.value)
--- a/lamindb/examples/cellxgene/__init__.py
+++ b/lamindb/examples/cellxgene/__init__.py
@@ -8,4 +8,9 @@
 
 """
 
-from ._cellxgene import create_cellxgene_schema, save_cellxgene_defaults
+from ._cellxgene import (
+    create_cellxgene_schema,
+    get_cxg_schema,
+    save_cellxgene_defaults,
+    save_cxg_defaults,
+)


lamindb/examples/cellxgene/_cellxgene.py
```diff
@@ -260,12 +274,59 @@ class CategorySpec(NamedTuple):
...
+            k: v for k, v in categoricals.items() if k in organism_fields
+        }
+
+        uns_features = [
+            Feature(
+                name=field,
+                dtype=uns_categoricals[field],
+                cat_filters={"source": sources[field]},
+                default_value=categoricals_to_spec[field].default,
+            ).save()
```
falexwolf, line 312: I didn't realize this before, but given the creation of things in the database the name `create_cxg_schema()` (or actually no need to abbreviate: `create_cellxgene_schema()` or `get_or_create_cellxgene_schema()`) instead of `get_cxg_schema()` might have been more appropriate.

If you're happy with the current name I suggest adding information to the doc string that this function is going to "clutter" the database with a bunch of features, labels, and schemas. ☺️

lamindb/examples/cellxgene/_cellxgene.py
```diff
@@ -260,12 +274,59 @@ class CategorySpec(NamedTuple):
...
+                name=f"scalefactors of spatial {spatial_library_id}",
+                features=[
+                    Feature(name="spot_diameter_fullres", dtype=float).save(),
+                    Feature(name="tissue_hires_scalef", dtype=float).save(),
+                ],
+            ).save()
+
+            spatial_schema = Schema(
+                name="spatial metadata",
+                features=[Feature(name="is_single", dtype=bool).save()],
```
falexwolf: What does this feature mean? Can we populate `description`?

lamindb/examples/cellxgene/_cellxgene.py
```diff
@@ -260,12 +274,59 @@ class CategorySpec(NamedTuple):
...
+            scalefactors_schema = Schema(
+                name=f"scalefactors of spatial {spatial_library_id}",
+                features=[
+                    Feature(name="spot_diameter_fullres", dtype=float).save(),
+                    Feature(name="tissue_hires_scalef", dtype=float).save(),
+                ],
+            ).save()
+
+            spatial_schema = Schema(
+                name="spatial metadata",
```
falexwolf: I think it's important to add `CxG` or `CellXGene` to the name because otherwise it will clash with user data due to auto-lookup upon name equality.

falexwolf: We should have only a single way to refer to cellxgene and that could be `cellxgene`, `cxg`, `CxG`, or `CellXGene`, I think.

lamindb/examples/cellxgene/_cellxgene.py
```diff
@@ -260,12 +274,59 @@ class CategorySpec(NamedTuple):
...
+                features=[Feature(name="is_single", dtype=bool).save()],
+            ).save()
+
+            slots["uns:spatial"] = spatial_schema
+            slots[f"uns:spatial:{spatial_library_id}:scalefactors"] = (
+                scalefactors_schema
+            )
+
     full_cxg_schema = Schema(
         name=f"AnnData of CELLxGENE version {schema_version} for {organism} of {', '.join(field_types) if isinstance(field_types, list) else field_types}",
```
falexwolf, line 353: Let's pay attention to whether `CELLxGENE` is consistent with all other references to cellxgene. There should be one and only one way to refer to it. ☺️

tests/curators/test_cxg_curator.py
```diff
@@ -1,21 +1,22 @@
...
 
-@pytest.fixture(scope="module")
-def cxg_schema():
-    ln.examples.cellxgene.save_cxg_defaults()
-    schema = ln.examples.cellxgene.get_cxg_schema(
-        schema_version="5.2.0", field_types=["name", "ontology_id"]
-    )
+@pytest.fixture
+def cxg_schema_factory():
+    def _create_schema(version: CELLxGENESchemaVersions, **kwargs):
```
falexwolf: I'd not prefix a locally scoped function with a prefix unless there is adjacent logic that's public within that scope.

lamindb/examples/cellxgene/_cellxgene.py
Zethson: `CELLxGENE` is how CZI named it and I've therefore decided to settle on that. Also renamed methods to have the long form `cellxgene` but lower cased of course.

falexwolf: Sounds great!

PR #3007, opened 2025-08-04
🐛 Annotate Artifacts passed to Curator with Schema

Zethson:
Earlier, Artifacts that were passed to a Curator were not annotated with the Schema. This PR ensures that the `artifact.schema` slot is always annotated.

Further fixes an issue where `key` was accessed despite it not being set during `save_artifact` for DataFrameCurator.

Internal context: https://laminlabs.slack.com/archives/C03Q5TXF797/p1754323073662859?thread_ts=1754320108.456249&cid=C03Q5TXF797

### DIFF
--- a/lamindb/curators/core.py
+++ b/lamindb/curators/core.py
@@ -686,10 +686,11 @@ class DataFrameCurator(Curator):
                 description=description,
                 revises=revises,
                 run=run,
-                format=".csv" if key.endswith(".csv") else None,
+                format=".csv" if key is not None and key.endswith(".csv") else None,
             )
-            self._artifact.schema = self._schema
+
             self._artifact.save()
+        self._artifact.schema = self._schema
         return annotate_artifact(  # type: ignore
             self._artifact,
             cat_vectors=self.cat._cat_vectors,
--- a/lamindb/curators/core.py
+++ b/lamindb/curators/core.py
@@ -364,13 +364,13 @@ class SlotsCurator(Curator):
                     )
                     break
 
-            self._artifact.save()
         cat_vectors = {}
         for curator in self._slots.values():
             for key, cat_vector in curator.cat._cat_vectors.items():
                 cat_vectors[key] = cat_vector
 
         self._artifact.schema = self._schema
+        self._artifact.save()
         return annotate_artifact(  # type: ignore
             self._artifact,
             curator=self,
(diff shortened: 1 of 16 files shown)


lamindb/curators/core.py
```diff
@@ -364,12 +364,13 @@ def save_artifact(
...
                     break
 
-            self._artifact.schema = self._schema
             self._artifact.save()
         cat_vectors = {}
         for curator in self._slots.values():
             for key, cat_vector in curator.cat._cat_vectors.items():
                 cat_vectors[key] = cat_vector
+
+        self._artifact.schema = self._schema
```
sunnyosun, line 372: I don't think this works as the foreign key won't get saved without `self._artifact.save()`?

lamindb/curators/core.py
```diff
@@ -685,10 +686,11 @@ def save_artifact(
                 description=description,
                 revises=revises,
                 run=run,
-                format=".csv" if key.endswith(".csv") else None,
+                format=".csv" if key is not None and key.endswith(".csv") else None,
             )
-            self._artifact.schema = self._schema
+
             self._artifact.save()
+        self._artifact.schema = self._schema
```
sunnyosun, line 692: Same here, see comment above.

lamindb/curators/core.py
Zethson: Well that's an embarrassing oversight. I fooled myself with the test that still has the memory status even after having called `curator.save_artifact`.

tests/curators/test_curators_general.py
```diff
@@ -242,6 +242,24 @@ def test_schema_not_saved(df):
...
+    """A passed Artifact should be annotated with a Schema if successfully curated."""
+    af = ln.Artifact.from_df(df, key="test.parquet").save()
+    schema = ln.Schema(
+        name="sample schema",
+        features=[ln.Feature(name="sample_id", dtype="str").save()],
+    ).save()
+    curator = ln.curators.DataFrameCurator(af, schema)
+    curator.validate()
+    curator.save_artifact()
+    assert af.schema is not None
```
sunnyosun: this here doesn't do proper check of the FK in the DB. The artifact needs to be queried from the DB first before checking.

sunnyosun (approved):
Feel free to merge after fixing the tests!

PR #3015, opened 2025-08-07
:card_file_box: Add `type` record constraint

chaichontat:
A Record can be a type only if `is_type=True` and `schema=null`.


lamindb/migrations/0119_add_record_fk_constraint.py
```diff
@@ -0,0 +1,55 @@
...
+from django.db import migrations
+
+TABLE_NAME = "lamindb_record"
+
+CREATE_FUNCTION_SQL = f"""
+CREATE OR REPLACE FUNCTION is_valid_record_type(type_id INTEGER)
+RETURNS BOOLEAN AS $$
+BEGIN
+    IF type_id IS NULL THEN
+        RETURN TRUE;
```
falexwolf: Why does this return `True` if `type_id` is `NULL`?

falexwolf:
Apologies that I'm only seeing this now. Can you provide a bit more context by linking a Slack discussion or similar?

What do you mean when saying "A Record can be a type"?

Of course I understand "if is_type=True and schema=null." but why do we need a SQL-level function for that?

The SQL function also seems to do something slightly different but I also don't understand it entirely.

If there was a bit of context either in the code or the PR description it'd probably click for me. ☺️

chaichontat:
The old version was actually wrong, put in the correct version with comments.

# PostgreSQL Function: `is_valid_record_type`

## Why This Must Be a PostgreSQL Function

### The Technical Constraint
PostgreSQL CHECK constraints **cannot contain subqueries**. You cannot write:
```sql
-- THIS DOESN'T WORK IN POSTGRESQL
ALTER TABLE lamindb_record 
ADD CONSTRAINT check_type CHECK (
    type_id IS NULL OR 
    EXISTS (SELECT 1 FROM lamindb_record WHERE id = type_id AND is_type = TRUE)
);
-- ERROR: cannot use subquery in check constraint
```

### Why Other Approaches Don't Work

1. **Standard Foreign Key**: Only validates that the ID exists, cannot check that `is_type = TRUE` and `schema_id IS NULL`

2. **Django's CheckConstraint**: Limited to simple field comparisons within the same row, cannot perform lookups to other rows

3. **Composite Foreign Key**: Would require `(id, is_type, schema_id)` to be a unique key, which doesn't match the data model

4. **Triggers**: More complex to write and maintain than a function-based CHECK constraint

### The Solution: Function Wrapper
PostgreSQL allows CHECK constraints to call functions, and functions can contain subqueries. This is the standard PostgreSQL pattern for complex referential integrity:
```sql
CHECK (is_valid_record_type(type_id))  -- This works!
```

## Logic Breakdown

### Line 4-6: NULL Check
```sql
IF type_id IS NULL THEN
    RETURN TRUE;
END IF;
```
- **NULL values are considered valid** - No type record is valid.

### Line 8-13: Validation for Type Records
```sql
IF is_type = TRUE THEN
    RETURN EXISTS (
        SELECT 1 FROM lamindb_record
        WHERE id = type_id AND is_type = TRUE AND schema_id IS NULL
    );
END IF;
```
- **When current record is a type**: Type inheritance requires stricter validation
- **Must reference schema-less types**: Types can only inherit from base types (not schema-specific ones)

### Line 15-18: Validation for Regular Records
```sql
RETURN EXISTS (
    SELECT 1 FROM lamindb_record
    WHERE id = type_id AND is_type = TRUE
);
```
- **Regular records**: Can reference any type (schema-specific or not)
- **Only requirement**: The referenced record must be a type (`is_type = TRUE`)

## Usage
This function is used in a CHECK constraint to ensure referential integrity:
```sql
CHECK (is_valid_record_type(type_id, is_type))
```

The constraint validates different rules based on the record type:
- **Type records** → Can only inherit from schema-less types
- **Regular records** → Can use any type
- **All records** → NULL type_id is always valid

falexwolf:
I see, makes sense!

falexwolf:
I'll take care of fixing tests.

Koncopd:
Ok, i had to add some fixes to the function, now it is fine. Also i believe adding this constraint to existing db will try to validate all existing rows, we will need to check if it will be successful.

falexwolf:
Thank you, @Koncopd! Sorry it took so long, @chaichontat!

falexwolf:
This surprisingly broke a test in lamindb-setup.

@Koncopd, do you know what the ambiguity could be?

image

https://github.com/laminlabs/lamindb-setup/actions/runs/17511719732/job/49744160193

PR #3026, opened 2025-08-18
✨ Support more organisms in CELLxGENE

Zethson:
Fixes https://github.com/laminlabs/lamindb/issues/2979

- Adds all supported organisms by CELLxGENE to the Literal typehint
- Now uses the `scientific_name` as expected.

This PR is a breakout from https://github.com/laminlabs/lamindb/pull/2980

### DIFF
--- a/lamindb/examples/cellxgene/_cellxgene.py
+++ b/lamindb/examples/cellxgene/_cellxgene.py
@@ -12,6 +12,16 @@ if TYPE_CHECKING:
     from lamindb.models import Schema, SQLRecord
 
 CELLxGENESchemaVersions = Literal["4.0.0", "5.0.0", "5.1.0", "5.2.0", "5.3.0"]
+CELLxGENEOrganisms = Literal[
+    "human",
+    "mouse",
+    "zebra danio",
+    "rhesus macaquedomestic pig",
+    "chimpanzee",
+    "white-tufted-ear marmoset",
+    "sars-2",
+    "synthetic construct",
+]
 FieldType = Literal["ontology_id", "name"]
 
 
@@ -25,7 +35,6 @@ def save_cxg_defaults() -> None:
     - "unknown" entries for DevelopmentalStage, Phenotype, and CellType
     - "tissue", "organoid", and "cell culture" ULabels (tissue_type)
     - "cell", "nucleus", "na" ULabels (suspension_type)
-
     """
     import bionty as bt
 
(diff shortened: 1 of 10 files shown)


lamindb/examples/cellxgene/_cellxgene.py
```diff
@@ -45,16 +54,19 @@ def save_cxg_defaults() -> None:
     ).save()
 
     # na, unknown
-    for model, name in zip(
+    for biorecord, name in zip(
```
sunnyosun: I think it's more consistent to call it model, biorecord would be confusing as it sounds like it's a record, but it's the class.

lamindb/examples/cellxgene/_cellxgene.py
```diff
@@ -76,6 +88,20 @@ def save_cxg_defaults() -> None:
...
+    taxonomy_ids = [
+        "NCBITaxon:9606",  # Homo sapiens (Human)
+        "NCBITaxon:10090",  # Mus musculus (House mouse)
+        "NCBITaxon:9544",  # Macaca mulatta (Rhesus monkey)
+        "NCBITaxon:9825",  # Sus scrofa domesticus (Domestic pig)
+        "NCBITaxon:9598",  # Pan troglodytes (Chimpanzee)
+        "NCBITaxon:9483",  # Callithrix jacchus (White-tufted-ear marmoset)
+        "NCBITaxon:7955",  # Danio rerio (Zebrafish)
+    ]
+    ncbitaxon_source = bt.Source.filter(name="ncbitaxon").first()
```
sunnyosun: why first() instead of specifying the version?

lamindb/examples/cellxgene/_cellxgene.py
```diff
@@ -143,6 +169,11 @@ def get_cxg_schema(
 
     from lamindb.models import Feature, Schema, ULabel
 
+    # Attempt to find the Schema early as building the Schema is expensive when looped
+    full_cxg_schema_name = f"AnnData of CELLxGENE version {schema_version} for {organism} of {', '.join(field_types) if isinstance(field_types, list) else field_types}"
```
sunnyosun: I'm not sure it's the best idea to rely on name matching here, I think relying on the hash would be better. But I get it takes a bit more time (which isn't terrible) and gives some warnings.

sunnyosun:
Does it support mixed organisms? I don't think any datasets has only the synthetic construct genes, they are always mixed with human genes. So I'm not sure how necessary it is to support it alone.

PR #3029, opened 2025-08-19
✨ Unstructured slot validation of scverse datastructures

Zethson:
This PR extends `AnnDataCurator` and `MuDataCurator` to support validation of (nested) dictionaries in `.uns` slots, similar to how `SpatialDataCurator` handles `attrs`. Now, you can define schema slots targeting arbitrarily deep paths in `.uns` for both AnnData and MuData objects. It further improves error handling.

**1. Validate a nested dict in AnnData `.uns`:**

```python
import lamindb as ln

adata = ln.core.datasets.small_dataset1(otype="AnnData")
adata.uns["study_metadata"] = {"temperature": 21.6, "experiment_id": "EXP001"}

uns_schema = ln.Schema(
    features=[
        ln.Feature(name="temperature", dtype=float).save(),
        ln.Feature(name="experiment_id", dtype=str).save(),
    ],
).save()

anndata_schema = ln.Schema(
    otype="AnnData",
    slots={
        "uns:study_metadata": uns_schema,
    },
).save()

curator = ln.curators.AnnDataCurator(adata, anndata_schema)
curator.validate()
```

**2. Validate nested dicts in MuData and AnnData-in-MuData:**
```python
mdata = ln.core.datasets.mudata_papalexi21_subset()
mdata.uns["study_metadata"] = {"temperature": 21.6, "experiment_id": "EXP001"}
mdata["rna"].uns["site_metadata"] = {"pos": 99.9, "site_id": "SITE001"}
mdata["rna"].uns["study_metadata"] = {
    "temperature": 21.6,
    "experiment_id": "EXP001",
}

study_uns_schema = ln.Schema(
    features=[
        ln.Feature(name="temperature", dtype=float).save(),
        ln.Feature(name="experiment_id", dtype=str).save(),
    ],
).save()

site_uns_schema = ln.Schema(
    features=[
        ln.Feature(name="pos", dtype=float).save(),
        ln.Feature(name="site_id", dtype=str).save(),
    ]
).save()

mdata_schema = ln.Schema(
    otype="MuData",
    slots={
        "uns:study_metadata": study_uns_schema,
        "rna:uns:site_metadata": site_uns_schema,
    },
).save()

curator = ln.curators.MuDataCurator(mdata, mdata_schema)
```

This PR is a breakout from https://github.com/laminlabs/lamindb/pull/3000

### DIFF
--- a/lamindb/curators/core.py
+++ b/lamindb/curators/core.py
@@ -466,7 +466,7 @@ class DataFrameCurator(Curator):
 
     Example:
 
-        For simple example using a flexible schema, see :meth:`~lamindb.Artifact.from_df`.
+        For a simple example using a flexible schema, see :meth:`~lamindb.Artifact.from_df`.
 
         Here is an example that enforces a minimal set of columns in the dataframe.
 
@@ -720,20 +720,23 @@ class DataFrameCurator(Curator):
         )
 
 
-def _navigate_nested_dict(
-    dict_obj: dict[str, Any], keys: Iterable[str], slot: str, base_path: str
+def _resolve_schema_slot_path(
+    target_dict: dict[str, Any], slot_keys: Iterable[str], slot: str, base_path: str
 ) -> Any:
-    """Traverse nested dictionaries with sane error behavior.
+    """Resolve a schema slot path by traversing nested dictionary keys.
 
     Args:
-        dict_obj: The dictionary to traverse
-        keys: Keys to traverse
-        slot: Original slot string (for error messages)
-        base_path: Base path string (for error messages)
-    """
-    current = dict_obj
+        target_dict: Root dictionary to traverse
+        slot_keys: Sequence of keys defining the paths to traverse
+        slot_name: Schema slot identifier for error context
+        base_path: Base path string for error context
 
-    for key in keys:
+    Returns:
+        The value at the resolved path
+    """
+    current = target_dict
+
+    for key in slot_keys:
         base_path += f"['{key}']"
         try:
             current = current[key]
@@ -752,14 +755,21 @@ def _navigate_nested_dict(
 def _handle_dict_slots(
     dataset: ScverseDataStructures, slot: str
 ) -> tuple[pd.DataFrame | None, str | None, str | None]:
-    """Handle dict-based slot paths (uns/attrs) for all ScverseCurators.
+    """Handle dict-based slot paths (uns/attrs standalone or of modalities) for all ScverseCurators.
+
+    Supports two patterns:
+        - Direct dict access: "uns", "attrs", "uns:key1:key2", "attrs:key"
+        - Modality dict access: "modality:uns"
 
     Args:
         dataset: The scverse datastructure object
         slot: The slot path string to parse like 'uns:path:to'.
 
     Returns:
-        tuple: (DataFrame, key, sub_slot)
+        tuple: (dataframe, modality_key, remaining_slot_path)
+            - dataframe: Single-row DataFrame containing the resolved data
+            - modality_key: Modality identifier if slot targets modality dict, else None
+            - remaining_slot_path: The dict attribute and nested keys as string
     """
     path_parts = slot.split(":")
 
@@ -771,7 +781,9 @@ def _handle_dict_slots(
                 return pd.DataFrame([dict_attr]), None, path_parts[0]
 
             deeper_keys = path_parts[1:]
-            data = _navigate_nested_dict(dict_attr, deeper_keys, slot, path_parts[0])
+            data = _resolve_schema_slot_path(
+                dict_attr, deeper_keys, slot, path_parts[0]
+            )
             return pd.DataFrame([data]), None, ":".join(path_parts[1:])
 
     # Handle modality dict slots: "modality:uns", "modality:uns:key1:key2"
@@ -785,14 +797,17 @@ def _handle_dict_slots(
                     return pd.DataFrame([dict_attr]), modality, dict_name
 
                 deeper_keys = path_parts[2:]
-                data = _navigate_nested_dict(
+                data = _resolve_schema_slot_path(
                     dict_attr, deeper_keys, slot, f"{modality}.{dict_name}"
                 )
                 return pd.DataFrame([data]), modality, ":".join(path_parts[1:])
         except (KeyError, AttributeError):
             pass
     else:
-        raise InvalidArgument(f"Dict slot '{slot}' not found in dataset")
+        raise InvalidArgument(
+            f"Invalid dict slot pattern '{slot}'. Expected formats: "
+            f"'uns', 'attrs', 'uns:key', 'attrs:key', 'modality:uns'"
+        )
 
     return None, None, None
 
@@ -914,10 +929,11 @@ class MuDataCurator(SlotsCurator):
             raise InvalidArgument("Schema otype must be 'MuData'.")
 
         for slot, slot_schema in schema.slots.items():
+            # Handle slots: "mdata.uns", "modality:uns"
             if "uns" in slot:
                 df, modality, modality_slot = _handle_dict_slots(self._dataset, slot)
             else:
-                # Handle modality slots: "modality:obs", "modality:var"
+                # Handle slots: "modality:obs", "modality:var"
                 parts = slot.split(":")
                 if len(parts) == 2:
                     modality, modality_slot = parts
@@ -933,7 +949,7 @@ class MuDataCurator(SlotsCurator):
                             f"Attribute '{modality_slot}' not found on modality '{modality}'"
                         ) from None
                 else:
-                    # Global MuData slots: "obs", "var" (uns is a dictionary and gets handled above)
+                    # Handle slots: "mdata:obs", "mdata:var" (uns is a dictionary and gets handled above)
                     modality, modality_slot = None, slot
                     schema_dataset = self._dataset
                     df = getattr(schema_dataset, modality_slot.rstrip(".T"))
@@ -993,11 +1009,12 @@ class SpatialDataCurator(SlotsCurator):
             raise InvalidArgument("Schema otype must be 'SpatialData'.")
 
         for slot, slot_schema in schema.slots.items():
+            # Handle slots: "sdata:attrs"
             if slot.startswith("attrs"):
                 df, table_key, table_slot = _handle_dict_slots(self._dataset, slot)
             else:
                 parts = slot.split(":")
-                # Handle table slots: "tables:table_key:obs", "tables:table_key:var"
+                # Handle slots: "tables:table_key:obs", "tables:table_key:var"
                 if len(parts) == 3 and parts[0] == "tables":
                     table_key, table_slot = parts[1], parts[2]
                     try:
@@ -1005,14 +1022,14 @@ class SpatialDataCurator(SlotsCurator):
                         df = getattr(slot_object, table_slot.rstrip(".T"))
                     except KeyError:
                         raise InvalidArgument(
-                            f"Table '{table_key}' not found in dataset.tables"
+                            f"Table '{table_key}' not found in sdata.tables"
                         ) from None
                     except AttributeError:
                         raise InvalidArgument(
                             f"Attribute '{table_slot}' not found on table '{table_key}'"
                         ) from None
                 else:
-                    # Legacy single attrs handling for backward compatibility
+                    # Handle legacy single attrs for backward compatibility
                     if len(parts) == 1 and parts[0] != "attrs":
                         logger.warning(
                             f"please prefix slot {slot} with 'attrs:' going forward"
@@ -1023,7 +1040,7 @@ class SpatialDataCurator(SlotsCurator):
                             table_slot = slot
                         except KeyError:
                             raise InvalidArgument(
-                                f"Slot '{slot}' not found in SpatialData.attrs"
+                                f"Slot '{slot}' not found in sdata.attrs"
                             ) from None
                     else:
                         raise InvalidArgument(f"Unrecognized slot format: {slot}")
--- a/lamindb/curators/core.py
+++ b/lamindb/curators/core.py
@@ -393,15 +393,15 @@ class SlotsCurator(Curator):
         )
 
 
-def is_list_of_type(value, expected_type):
+def _is_list_of_type(value: Any, expected_type: Any) -> bool:
     """Helper function to check if a value is either of expected_type or a list of that type, or a mix of both in a nested structure."""
     if isinstance(value, Iterable) and not isinstance(value, (str, bytes)):
         # handle nested lists recursively
-        return all(is_list_of_type(item, expected_type) for item in value)
+        return all(_is_list_of_type(item, expected_type) for item in value)
     return isinstance(value, expected_type)
 
 
-def check_dtype(expected_type) -> Callable:
+def _check_dtype(expected_type: Any) -> Callable:
     """Creates a check function for Pandera that validates a column's dtype.
 
     Supports both standard dtype checking and mixed list/single values for the same type.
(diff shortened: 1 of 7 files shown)


lamindb/curators/core.py
```diff
@@ -393,15 +393,15 @@ def save_artifact(
         )
 
 
-def is_list_of_type(value, expected_type):
+def _is_list_of_type(value: Any, expected_type: Any) -> bool:
```
falexwolf: Why make it private given we're in an "internal module"? (The user API is re-exported)

I know we're not entirely consistent but I've tried to converge towards consistently not using underscores unless indicating private functionality relative to the internal API.

lamindb/curators/core.py
```diff
@@ -900,63 +1009,66 @@ def __init__(
...
-                    else getattr(slot_object, sub_slot)
-                )
-            elif len(split_result) == 1 or (
-                len(split_result) > 1 and split_result[0] == "attrs"
-            ):
-                table_key = None
-                if len(split_result) == 1:
-                    if split_result[0] != "attrs":
+                    # Handle legacy single attrs for backward compatibility
+                    if len(parts) == 1 and parts[0] != "attrs":
```
falexwolf, line 976: Why is a single `attrs` slot deprecated? Or am I misunderstanding this?

falexwolf: I thought `attrs` is handled in line 1014

tests/curators/test_curators_examples.py
```diff
@@ -563,33 +563,62 @@ def test_anndata_curator_varT_curation_legacy(ccaplog):
             varT_schema.delete(permanent=True)
 
 
-def test_soma_curator(curator_params: dict[str, str | FieldAttr]):
-    """Test SOMA curator implementation."""
+def test_anndata_curator_nested_uns():
```
falexwolf: Tell me if I'm off, but I think we have this test that iterates over different anndata object compositions. And one of the cases should be the case where it also validates `uns`.

If this was integrated into "one comprehensive anndata test" it'd be easier to ascertain correctness.

tests/curators/test_curators_examples.py
```diff
@@ -563,33 +563,62 @@ def test_anndata_curator_varT_curation_legacy(ccaplog):
...
-        "B cell",
-    }
-    assert set(artifact.features.get_values()["cell_type_by_model"]) == {
-        "T cell",
-        "B cell",
-    }
+    anndata_schema = ln.Schema(
+        otype="AnnData",
+        slots={
+            "uns:study_metadata": uns_schema,
```
falexwolf: Could we make the example parallel to what we have for spatialdata? So that it's really easy to see the parallelism?

tests/curators/test_curators_examples.py
```diff
@@ -563,33 +563,62 @@ def test_anndata_curator_varT_curation_legacy(ccaplog):
...
-    )
+    adata.uns["study_metadata"] = {"temperature": 21.6, "experiment_id": "EXP001"}
 
-    curator = ln.Curator.from_tiledbsoma(  # type: ignore
-        "./small_dataset1.tiledbsoma",
-        var_index={"RNA": ("var_id", bt.Gene.ensembl_gene_id)},
-        **curator_params,
-    )
-    artifact = curator.save_artifact(key="examples/dataset1.tiledbsoma")
+    uns_schema = ln.Schema(
```
falexwolf: This is the "study_metadata" schema.

tests/curators/test_curators_examples.py
```diff
@@ -563,33 +563,62 @@ def test_anndata_curator_varT_curation_legacy(ccaplog):
...
+
+    assert artifact.schema == anndata_schema
+    assert "uns:study_metadata" in artifact.features.slots
+    assert artifact.features.slots[
+        "uns:study_metadata"
+    ].features.first() == ln.Feature.get(name="temperature")
+
+    with pytest.raises(InvalidArgument) as e:
+        adata = datasets.small_dataset1(otype="AnnData")
+        bad_schema = ln.Schema(
```
falexwolf: Please move this outside of the `pytest.raises`

tests/curators/test_curators_examples.py
```diff
@@ -563,33 +563,62 @@ def test_anndata_curator_varT_curation_legacy(ccaplog):
...
+            otype="AnnData",
+            slots={"uns:nonexistent": uns_schema},
+        ).save()
+        ln.curators.AnnDataCurator(adata, bad_schema)
+    assert (
+        "Schema slot 'uns:study_metadata' specifies path uns['study_metadata'] but key 'study_metadata' not found."
```
falexwolf: I think we shouldn't call the `:` structure "path". It's too confusing. I know it's hard to come up with something else but I'd rather refer to "keys" or "hierarchical keys".

falexwolf: I think "requires" would be clearer than "specifies"

falexwolf (approved):
I'll continue the review tonight or tomorrow morning! Haven't made it entirely through yet but am worried that comments get lost.

lamindb/curators/core.py
Zethson: I didn't have the impression that these functions were designed to even be reused among the dev API and therefore I added underscores. I am happy to revert that.

falexwolf: I think what we're building towards is a pretty type inference API that has these components and the ones used in `FeatureManager.add_values()` -- sooner or later it will be useful to move `is_list_of_type()` et al. (including this https://github.com/laminlabs/lamindb/blob/a935ab8a383f9b2981f6837bb9ce9aa3bc5524a6/lamindb/models/_feature_manager.py#L511 & [this](https://github.com/laminlabs/lamindb/blob/a935ab8a383f9b2981f6837bb9ce9aa3bc5524a6/lamindb/models/_feature_manager.py#L499) and likely others) into a dedicated module in `.base` and re-export them where they're currently being used.

`lamindb.base.dtype` could be a good module. But not urgent.

tests/curators/test_curators_examples.py
```diff
@@ -563,33 +563,62 @@ def test_anndata_curator_varT_curation_legacy(ccaplog):
...
+        in str(e.value)
+    )
+
+    with pytest.raises(InvalidArgument) as e:
+        bad_schema = ln.Schema(
+            otype="AnnData",
+            slots={"uns:temperature:nonexistent_nested": uns_schema},
+        ).save()
+        ln.curators.AnnDataCurator(adata, bad_schema)
+    assert "key 'study_metadata' not found" in str(e.value)
```
falexwolf, line 620: We have to pay attention to the word choice here and double-check that this is consistent with what pandera and lamindb issue otherwise.

"key `study_metadata` not found" sounds to me like there is a `study_metadata` key in the dictionary but it's not found in the schema (I think that intuition comes from FileNotFound error, KeyError)

I think pandera says "column xyz is missing in dataframe" -- I think a similar word choice would be clearer

tests/curators/test_curators_examples.py
```diff
@@ -642,54 +671,94 @@ def test_mudata_curator(
...
     artifact.delete(permanent=True)
+    SchemaComponent.filter().delete()
+    ln.Schema.filter().delete()
 
 
-@pytest.fixture(scope="module")
-def spatialdata_blobs_schema():
-    import sys
-    from pathlib import Path
+def test_mudata_curator_nested_uns():
```
falexwolf: it'd be really nice if all of these functions were based on an `unstructured_metadata` fixture (or `dictionary_metadata`)

This would reduce the number of lines and make it really easy to see how the mechanics is parallel across all curators and data structures

tests/curators/test_curators_examples.py
falexwolf: This is the test that I'm talking about: https://github.com/laminlabs/lamindb/blob/92a9eef3a09db1c89864693d69fa325a528093f2/tests/curators/test_curators_examples.py#L411

Essentially adding more parametrization to the `uns` case.

I admit that parametrizing tests in order to prevent a flat list of 100 disconnected redundant tests is hard and there might be reasons not to do this, so please object if you think so!

falexwolf:
# df.attrs

Given 

- parquet has the unstructured metadata in the spec (https://github.com/laminlabs/pfizer-lamin-usage/issues/256#issuecomment-3160678731) 
- Julianus thinks it'd be nice (https://github.com/laminlabs/pfizer-lamin-usage/issues/256#issuecomment-3160302200)
- it create some nice analogies across curators and users won't go like "why can't I validate my metadata in `df.attrs`"
- it'd be a severe breaking change that pandas would make that seems unlikely, discussions are already years old and there would be substantial opposition to remove `.attrs`
- also the polars API has support and we'll soon add more support for it

I'd suggest you also add `df.attrs` validation support here? WDYT?

PS: Here is a polars example

```
import polars as pl

# Create DataFrame
df = pl.DataFrame({"a": [1, 2, 3], "b": [4, 5, 6]})

# Write with custom metadata
metadata = {"source": "experiment_1", "created": "2024-01-01", "version": "1.0"}
df.write_parquet("data.parquet", metadata=metadata)
```

falexwolf:
# Documenting examples and re-using as fixtures

Another important thing that's missing is to document the logic here as part of the `Curator` documentation, and then re-export it in the `curate` guide.

I also think that it'd be good to extend the mini immuno schema: https://github.com/laminlabs/lamindb/blob/main/docs/scripts/define_mini_immuno_schema_flexible.py

We already have some "canonical metadata" for the dataset level metadata that would go into `uns`: https://github.com/laminlabs/lamindb/blob/a935ab8a383f9b2981f6837bb9ce9aa3bc5524a6/lamindb/examples/datasets/mini_immuno.py#L125-L130

One could just pick 2 of the dictionary fields in there.

---

Probably even better: one could make a `define_unstructured_schema.py` example that's then used as a fixture in the tests and in the `curate` guide and in the `Curators` docstring. This would maximally deduplicate the code and create clarity about that it's one and only one mechanism for whenever unstructured metadata is validated.

falexwolf:
# Defining what a slot is

We need to define what a `slot` is somewhere in the `curate` guide and re-export that in the `SlotCurator` children; because nobody will look into the base class definition and nobody will know what either a `slot` and a `SlotCurator` is.

This is particularly important now as we're departing from `slot = attribute of a python data structure` (*).

I'd say the old defintion (*) was indeed not a good definition given we deal with the problem of describing data in storage and not with describing a Python API. So, it's good that we depart from that.

Conversely, if you look at the `zarr` anndata schema on disk it's clear that `.obs` is in fact part of the zarr schema specification and thus not related attached to the Python API.

Hence: I think there is an easy intuitive explanation along the lines of "how do I enforce structure on unstructured metadata"?

So, you can say that depositing data in parquet, anndata, spatialdata etc. imposes conventions on how to structure arrays (and pandera does this in a more fine-grained way). Pydantic imposes conventions on how to structure dictionaries.

A slot is a "key" into a composite data structure that describes the where a set of features and their values a stored. In most cases feature measurements come as a table, but in some cases they come in unstructured or tensor form.

```python
dictionary  # set of key-value pairs where values are unstructured
df  # set of key-value pairs where values are columnar
df.attrs  # set of key-value pairs where values are unstructured
adata.var  # set of key-value pairs where values are columnar
adata.obs # set of key-value pairs where values are columnar
adata.X  # set of key-value pairs where values are sparse vectors
adata.uns  # set of key-value pairs where values are unstructured
...
```

---

The tricky part (as outlined on the Notion design doc) is explaining the ambiguity when attempting to validate

```python
{ 
    "temperature": 21.1,
    "study_params": {"a": 1, "b": "2"}
}
```

Is `"study_params"` a `feature` name or a `slot` key?

The default simple way is to consider it a feature:
```
ln.Schema(
   [
        ln.Feature(name="temperature", dtype="dict").save(),
        ln.Feature(name="study_params", dtype="dict").save()
   ],
   otype="dict",
).save()
```

But if you want to validate "a" and "b" then you need a slot:
```
ln.Schema(
   [
        ln.Feature(name="temperature", dtype="dict").save(),
   ],
   components=[
        "study_params": ln.Schema(
            ln.Feature(name="a", dtype=int).save()
            ln.Feature(name="b", dtype=int).save()
        ).save()
   ],
   otype="dict",
).save()
```

So far so good, but what about feature names "a" and "b" that appear under a different name space? We could indicate it this way rather pain-free:
```
study_params = ln.Feature(name="study_params", is_type=True).save()
ln.Feature(name="a", dtype=int, type=study_params).save()
ln.Feature(name="b", dtype=int, type=study_params).save()
ln.Schema(
   [
        ln.Feature(name="temperature", dtype="dict").save(),
   ],
   components=[
        study_params.name: study_params.to_schema()
   ],
   otype="dict",
).save()
```
where (new small method):
```
class Feature:
     def to_schema(self):
          assert self.is_type, "can only convert feature types to schemas"
```

Making the curator aware of the fact that it has to choose feature "a" with `type=study_params` in case it's nested under "study_params" is where things get tricky and what would then constitute a full `DictCurator`, I imagine. This is where we need a warning box that we don't yet support it.

lamindb/curators/core.py
Zethson: I started with such a module but haven't started refactoring/extracting methods from `FeatureManager.add_values()`. I fear that if I also start doing that now, the PR becomes massive and messy again. I am happy to immediately follow this PR up with that if you think I should complete that refactor.

tests/curators/test_curators_examples.py
Zethson: I myself (feel free to object!) prefer small atomic unit tests (that can be pytest parametrized) where it's clear what they're testing and what their side effects (that we don't want) could be. This is something that Phil Angerer kept stressing recently as well.

> If this was integrated into "one comprehensive anndata test" it'd be easier to ascertain correctness.

I don't think that we'd be winning a lot (if anything) if I integrated it so unless you feel strongly about this, I'd like to keep the isolated test.

tests/curators/test_curators_examples.py
Zethson: It is now actually more parallel to what we have for MuData and SpatialData is the outlier right? I'd rather love to break it out for SpatialData for an isolated test (see https://github.com/laminlabs/lamindb/pull/3029#discussion_r2288765983).

Like I said in my other comment, please let me know if you feel strongly about this and I'll integrate it into one long test (that I would prefer to avoid).

tests/curators/test_curators_examples.py
Zethson: Hmm but isn't the case here that we're only converting the dict for internal processing to a DataFrame? So from a user perspective, they are actually specifying a key to a dictionary (here `uns`) and telling them that a column is missing in a DataFrame is confusing. The error that is omitted here is not about the validation but solely about whether the key (formerly path) exists in the object.

lamindb/curators/core.py
Zethson: This is https://github.com/laminlabs/lamindb/blob/main/lamindb/curators/core.py#L932 case. Sorry, my fingers must've somehow slipped and I wrote `attrs` here but it should have just be `key`.

Zethson:
> I'd suggest you also add df.attrs validation support here? WDYT?

Yup!

> Documenting examples and re-using as fixtures

Yup! 

> Defining what a slot is

Yup!

Now without answering anything in detail, do you want me to tackle all 3 things in this PR or make a follow up PR? This one already has ~650 changed lines and these changes might get us closer to 1k and then we have another fat PR by me that I had wanted to avoid. PR feature creep :smile: 

I personally would prefer two more follow up PRs that can be more easily reviewed.

lamindb/curators/core.py
falexwolf: Yeah, definitely not as part of this PR. I think after this PR this should come next: https://laminlabs.slack.com/archives/C04FPE8V01W/p1755702104978259

After that we're in a position where things are comprehensive and we can see everything that's needed; so making that internal API pretty will make sense. I'd say not urgent but can be done whenever it feels like it's the right moment.

PR #3030, opened 2025-08-20
✨ Add `Feature.from_dict()` 

Zethson:
Adds a `Feature.from_dict` constructor for features.

```python
import lamindb as ln

dict = {
        "feat1": 42,
        "feat2": 3.14,
        "feat3": "test_string",  # string (ambiguous cat ? str)
        "feat4": True,
        "feat5": [1, 2, 3],
        "feat6": ["a", "b", "c"],  # list[str] (ambiguous list[cat ? str])
        "feat7": {"key": "value"},
    }
features = ln.Feature.from_dict(
        dict, field=ln.Feature.name, str_as_cat=True
    )
```

## technical details

- When (lists of) strings are passed, it's not always clear whether these are supposed to be `str` or `cat` type. To handle this case, I've introduced a `str_as_cat` parameter. If this is set to `None` and `str` or `list[str]` are detected, an error is raised to either set this value to `True` or create the features individually.
- For consistency, it might be better to rename `str_as_cat` to `coerce_dtype`

### DIFF
--- a/docs/arrays.ipynb
+++ b/docs/arrays.ipynb
@@ -77,6 +77,7 @@
    "outputs": [],
    "source": [
     "import lamindb as ln\n",
+    "import numpy as np\n",
     "\n",
     "ln.track(\"hsRyWJggf2Ca\")"
    ]
@@ -144,7 +145,7 @@
    },
    "outputs": [],
    "source": [
-    "adata = artifact.open()"
+    "access = artifact.open()"
    ]
   },
   {
@@ -164,7 +165,7 @@
    },
    "outputs": [],
    "source": [
-    "adata"
+    "access"
    ]
   },
   {
@@ -184,7 +185,7 @@
    },
    "outputs": [],
    "source": [
-    "adata.X"
+    "access.X"
    ]
   },
   {
@@ -204,11 +205,11 @@
    },
    "outputs": [],
    "source": [
-    "obs_idx = adata.obs.cell_type.isin([\"Dendritic cells\", \"CD14+ Monocytes\"]) & (\n",
-    "    adata.obs.percent_mito <= 0.05\n",
+    "obs_idx = access.obs.cell_type.isin([\"Dendritic cells\", \"CD14+ Monocytes\"]) & (\n",
+    "    access.obs.percent_mito <= 0.05\n",
     ")\n",
-    "adata_subset = adata[obs_idx]\n",
-    "adata_subset"
+    "access_subset = access[obs_idx]\n",
+    "access_subset"
    ]
   },
   {
@@ -228,7 +229,7 @@
    },
    "outputs": [],
    "source": [
-    "adata_subset.X"
+    "access_subset.X"
    ]
   },
   {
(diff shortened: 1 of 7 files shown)


falexwolf:
Looking! Great! 😄 

First comment: Rather than putting this example here in the GitHub PR comment where it will be lost for users, can you please put it in the `Feature` doc string or in the `Feature.from_dict()` doc string and then screenshot it in the PR description?

image

lamindb/models/feature.py
```diff
@@ -777,28 +777,117 @@ def __init__(
...
         for name, col in df.items():
             if name in categoricals:
                 dtypes[name] = "cat"
             else:
                 dtypes[name] = serialize_pandas_dtype(col.dtype)
-        with logger.mute():  # silence the warning "loaded record with exact same name "
-            features = [
-                Feature(name=name, dtype=dtype) for name, dtype in dtypes.items()
-            ]  # type: ignore
+        if mute:
```
falexwolf, line 803: I totally don't have a strong opinion but am wondering why you saw the need to un-mute? 🤔

lamindb/models/feature.py
```diff
@@ -777,28 +777,117 @@ def __init__(
...
+            with logger.mute():
+                features = [
+                    Feature(name=key, dtype=dtype) for key, dtype in dtypes.items()
+                ]  # type: ignore
+        else:
+            features = [Feature(name=key, dtype=dtype) for key, dtype in dtypes.items()]  # type: ignore
```
falexwolf: The fact that the key logic gets duplicated because of `with logger.mute()` isn't super nice

I suggest rather than duplicating core logic to toggle logger verbosity depending on the state of `mute`

https://github.com/laminlabs/lamin-utils/blob/2a2c54522565f8d043686c247868718dfb6fb0da/lamin_utils/_logger.py#L262

lamindb/models/feature.py
Zethson: I thought that giving users the option doesn't hurt and can help in edge cases. It's also a bit more consistent with constructors like `from_values`.

lamindb/models/feature.py
```diff
@@ -777,28 +777,117 @@ def __init__(
...
+                    )
+                    continue
+                if str_as_cat:
+                    dtype = "list[cat]"
+                else:
+                    dtype = "list[str]"
+
+            dtypes[key] = dtype
+
+        if ambiguous_keys:
```
falexwolf, line 873: Nicely done! I overlooked this complexity coming from the instances in the mlops docs where we just looped over the param dictionary and, I guess, defaulted to `str` rather than ever considering categoricals.

tests/core/test_feature.py
```diff
@@ -160,6 +173,44 @@ def test_feature_from_df(df):
...
+
+def test_feature_from_dict(dict_data):
+    # ambiguous str types
+    with pytest.raises(ValueError) as e:
+        features = ln.Feature.from_dict(dict_data, str_as_cat=None)
+    error_msg = str(e.value)
+    assert "Ambiguous dtypes detected" in error_msg
+    assert "'dict_feat3': str or cat" in error_msg
+    assert "'dict_feat6': list[str] or list[cat]" in error_msg
+    assert "Please specify `str_as_cat` parameter" in error_msg
```
falexwolf: We typically say "pass" rather than "specify"

"passing an argument to a function"

tests/core/test_feature.py
```diff
@@ -160,6 +173,44 @@ def test_feature_from_df(df):
...
-    ln.ULabel.filter().all().delete()
-    ln.Feature.filter().all().delete()
+    ln.Schema.filter().delete()
+    ln.ULabel.filter().delete()
+    ln.Feature.filter().delete()
+
+
+def test_feature_from_dict(dict_data):
+    # ambiguous str types
+    with pytest.raises(ValueError) as e:
```
falexwolf, line 183: Great!

falexwolf (approved):
Only mini requested changes. That's a really nice test and really nice UX I think!

lamindb/models/feature.py
falexwolf: > It's also a bit more consistent with constructors like from_values.

The reason for why we have it in `from_values()` is because `from_values()` is both heavily used internally _and_ through users.

In internal usage (inside curators and validation logic) all logging is handled outside `from_values()` -- you'd see double information on whether certain values are valid or not etc.

In usage through users they, of course, want to see which values are validated and which aren't: the logging is highly desired.

But writing all this I realize and now agree that users _might_ want to see the "feature already exists" warning. In fact I'd default `mute=False` consistent with `from_values()` and then set it to `True` where `from_df()` is used internally in the curators (if we even still do it; I think this was heavily used in the old curators when we were still double parsing feature identifiers both during validation and saving).

Zethson: Yup! We're now defaulting to `mute=False` and I adapted the single usage in the curators.

falexwolf:
I tried to fix the failing test by deleting this artifact:
image

falexwolf:
There is something off with idempotency of this access mgmt test in case it errors, I will fix this.

Zethson:
> There is something off with idempotency of this access mgmt test in case it errors, I will fix this.

@falexwolf the permission test is very flaky and/or keeps failing.

PR #3023, opened 2025-08-16
♻️ Simplify init_storage()

falexwolf:

PR #3032, opened 2025-08-21
♻️ Amend storage access tests

falexwolf:
Restores the test that was accidentally deleted in the second-to-last commit in the previous PR:

- https://github.com/laminlabs/lamindb/pull/3023

Makes the space-based access management test that hits `laminlabs/lamin-dev` idempotent.


falexwolf:
The FieldTracker gives problems at instance setup because it patches `__init__`. So, we're back to manually tracking these changes.

PR #3034, opened 2025-08-22
✨ Allow getting artifacts by path via `Artifact.get(path="...")`

Koncopd:
Example:
```python
artifact = ln.Artifact.get(path="s3://bucket/folder/file.h5ad")
# or in another instance
artifact = ln.Artifact.using("laminlabs/cellxgene").get(path="s3://bucket/folder/file.h5ad")
```

Internal [Slack discussion](https://laminlabs.slack.com/archives/C04FPE8V01W/p1746009345011899).

### DIFF
--- a/lamindb/models/artifact.py
+++ b/lamindb/models/artifact.py
@@ -1691,7 +1691,7 @@ class Artifact(SQLRecord, IsVersioned, TracksRun, TracksUpdates):
         )
 
     @classmethod
-    def get_from_path(cls, path: UPathStr) -> Artifact | None:
+    def get_by_path(cls, path: UPathStr) -> Artifact | None:
         """Get the latest artifact registered for the path if exists, ``None`` otherwise.
 
         Args:
@@ -1699,7 +1699,7 @@ class Artifact(SQLRecord, IsVersioned, TracksRun, TracksUpdates):
         """
         from .query_set import QuerySet
 
-        return QuerySet(model=cls).get_from_path(path)
+        return QuerySet(model=cls).get_by_path(path)
 
     @classmethod
     def get(
--- a/lamindb/models/artifact.py
+++ b/lamindb/models/artifact.py
@@ -1704,6 +1704,7 @@ class Artifact(SQLRecord, IsVersioned, TracksRun, TracksUpdates):
             idlike: Either a uid stub, uid or an integer id.
             is_run_input: Whether to track this artifact as run input.
             expressions: Fields and values passed as Django query expressions.
+                Use `path=...` to get an artifact for a local or remote filepath if exists.
 
         Raises:
             :exc:`docs:lamindb.errors.DoesNotExist`: In case no matching record is found.
@@ -1718,6 +1719,7 @@ class Artifact(SQLRecord, IsVersioned, TracksRun, TracksUpdates):
 
                 artifact = ln.Artifact.get("tCUkRcaEjTjhtozp0000")
                 artifact = ln.Arfifact.get(key="examples/my_file.parquet")
+                artifact = ln.Artifact.get(path="s3://bucket/folder/adata.h5ad")
         """
         from .query_set import QuerySet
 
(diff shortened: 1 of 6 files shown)


lamindb/models/artifact.py
```diff
@@ -1690,6 +1690,17 @@ def _cache_path(self) -> UPath:
             filepath, cache_key=cache_key
         )
 
+    @classmethod
+    def get_by_path(cls, path: UPathStr) -> Artifact | None:
```
falexwolf: This should never return `None` but hard error instead, shouldn't it?

falexwolf:
This is great! I like it much better than the first version with `get_by_path()` ☺️ 

image

In particular the first example needs to be part of the docstring of `Artifact.get()`. It's always better to just share screenshots of the docs because then we don't forget about adding logic to docs.

lamindb/models/artifact_set.py
```diff
@@ -120,3 +123,33 @@ def mapped(
...
         _track_run_input(artifacts, is_run_input)
         return ds
+
+
+def artifacts_from_path(artifacts: ArtifactSet, path: UPathStr) -> ArtifactSet:
+    upath = UPath(path)
+
+    path_str = upath.as_posix()
+
+    stem = upath.stem
```
falexwolf, line 134: What about regular file names that have length 16? Can you please add a comment for why this won't cause us issues?

falexwolf: I see, because `_key_is_virtual` guarantees it.

lamindb/models/artifact_set.py
```diff
@@ -120,3 +123,33 @@ def mapped(
         # track only if successful
         _track_run_input(artifacts, is_run_input)
         return ds
+
+
+def artifacts_from_path(artifacts: ArtifactSet, path: UPathStr) -> ArtifactSet:
```
falexwolf, line 128: Do you want to add a mini docstring that describes that an `output_field` annotation is added to a filtered copy of the artifactset that's passed?

tests/core/test_artifact.py
```diff
@@ -1011,3 +1011,19 @@ def test_artifact_get_tracking(df):
 
     artifact.delete(permanent=True)
     transform.delete()
+
+
+def test_get_by_path(df):
```
falexwolf, line 1017: Rather than adding a new test, can you use the existing test cases where we distinguish `_is_virtual_key`  True/False through parametrizing with the file path fixture?

Simply by adding `.get()` at the end of this test https://github.com/laminlabs/lamindb/blob/03a73bd8c25bb269bdfdaee4f2300d126c4454aa/tests/core/test_artifact.py#L526 you'd cover all cases.

falexwolf:
I do think people will eventually be confused when they start to view the virtual path in lamindb as a path, too. But of course that path won't yield the expected result.

So, in your test case, the following would error:

```python
ln.Artifact.get(path="default_storage_unit_core/df.parquet")
```

git can do more magic here because it never expose the real storage path at all; it always keeps the last layer of the version history as the materialized path and then through checking out earlier versions of the git repo one materializes earlier versions of files in the semantic path (based on the real path in blob store in `.git`).

lamindb, unfortunately, doesn't have this, and so we have two user facing concepts: `path` and `virtual path`.

I think it may be the least confusing if users who want to query by virtual path, use `storage` & `key` similar to how they'd do it with the S3 API:

```python
ln.Artifact.get(key="df.parquet", storage__root="default_storage_unit_core")
```

Of course they can omit the second argument in most cases because they'll not have key duplications across multiple storage locations in the same instance.

lamindb/models/artifact_set.py
Koncopd: Yes, if it fails to find in the first part, it just checks in the second part of the function.

tests/core/test_artifact.py
Koncopd: done

Koncopd:
>In particular the first example needs to be part of the docstring of Artifact.get(). 

Done, added to the docstring.

lamindb/models/artifact_set.py
Koncopd: I changed that to `.alias` to avoid this spillover.

PR #3036, opened 2025-08-25
✨ Support validating `attrs` of Pandas DataFrames

Zethson:
In reference to https://github.com/laminlabs/lamindb/pull/3029#issuecomment-3206598980.

- Enables support for the validation of the `attrs` slot of DataFrames using the DataFrameCurator

```python
import lamindb as ln
df = ln.core.datasets.mini_immuno.get_dataset1(otype="DataFrame")
df.attrs = {"temperature": 21.6, "experiment_id": "EXP001"}

study_metadata_schema = ln.Schema(
    features=[
        ln.Feature(name="temperature", dtype=float).save(),
        ln.Feature(name="experiment_id", dtype=str).save(),
    ],
).save()

main_schema = ln.Schema(
    features=[ln.Feature(name="perturbation", dtype="str").save()],
).save()

schema = ln.Schema(
    slots={
        "df": main_schema,
        "attrs": study_metadata_schema
    },
    otype="DataFrame",
).save()

curator = ln.curators.DataFrameCurator(
    df, schema=schema
)
curator.validate()
artifact = curator.save_artifact(key="examples/df_with_attrs.parquet")

artifact.describe()
```

<pre style="white-space:pre;overflow-x:auto;line-height:normal;font-family:Menlo,'DejaVu Sans Mono',consolas,'Courier New',monospace"><span style="font-weight: bold">Artifact</span><span style="color: #7f7f7f; text-decoration-color: #7f7f7f; font-weight: bold"> .parquet · DataFrame · dataset</span>
<span style="color: #7f7f7f; text-decoration-color: #7f7f7f">├── </span><span style="color: #00ffff; text-decoration-color: #00ffff; font-weight: bold">General</span>
<span style="color: #7f7f7f; text-decoration-color: #7f7f7f">│   ├── </span><span style="color: #7f7f7f; text-decoration-color: #7f7f7f">key: </span><span style="color: #00d7af; text-decoration-color: #00d7af">examples/df_with_attrs.parquet</span>
<span style="color: #7f7f7f; text-decoration-color: #7f7f7f">│   ├── </span><span style="color: #7f7f7f; text-decoration-color: #7f7f7f">uid: </span>40WdBYZLNbPXW8re0000          <span style="color: #7f7f7f; text-decoration-color: #7f7f7f">hash: </span>lxqqxmMIUsUAVLR9yJdaLA
<span style="color: #7f7f7f; text-decoration-color: #7f7f7f">│   ├── </span><span style="color: #7f7f7f; text-decoration-color: #7f7f7f">size: </span>9.3 KB                       <span style="color: #7f7f7f; text-decoration-color: #7f7f7f">transform: </span><span style="color: #00d7af; text-decoration-color: #00d7af">None</span>
<span style="color: #7f7f7f; text-decoration-color: #7f7f7f">│   ├── </span><span style="color: #7f7f7f; text-decoration-color: #7f7f7f">space: </span>all                         <span style="color: #7f7f7f; text-decoration-color: #7f7f7f">branch: </span>all
<span style="color: #7f7f7f; text-decoration-color: #7f7f7f">│   ├── </span><span style="color: #7f7f7f; text-decoration-color: #7f7f7f">created_by: </span>zethson                <span style="color: #7f7f7f; text-decoration-color: #7f7f7f">created_at: </span>2025-08-25 18:43:47
<span style="color: #7f7f7f; text-decoration-color: #7f7f7f">│   ├── </span><span style="color: #7f7f7f; text-decoration-color: #7f7f7f">n_observations: </span>3
<span style="color: #7f7f7f; text-decoration-color: #7f7f7f">│   └── </span><span style="color: #7f7f7f; text-decoration-color: #7f7f7f">storage path: </span><span style="color: #00d7af; text-decoration-color: #00d7af">/home/lukas/code/lamindb/run-tests</span>/examples/df_with_attrs.parquet
<span style="color: #7f7f7f; text-decoration-color: #7f7f7f">└── </span><span style="color: #ff00ff; text-decoration-color: #ff00ff; font-weight: bold">Dataset features</span>
<span style="color: #7f7f7f; text-decoration-color: #7f7f7f">    ├── </span><span style="color: #d787ff; text-decoration-color: #d787ff; font-weight: bold">df</span><span style="color: #7f7f7f; text-decoration-color: #7f7f7f; font-weight: bold"> • </span><span style="color: #ffafd7; text-decoration-color: #ffafd7; font-weight: bold">1</span><span style="font-weight: bold">                          </span><span style="color: #ffafd7; text-decoration-color: #ffafd7; font-weight: bold">[Feature]</span><span style="font-weight: bold">                                                                  </span>
<span style="color: #7f7f7f; text-decoration-color: #7f7f7f">    │   </span>perturbation                   <span style="color: #7f7f7f; text-decoration-color: #7f7f7f"> str                               </span>                                         
<span style="color: #7f7f7f; text-decoration-color: #7f7f7f">    └── </span><span style="color: #d787ff; text-decoration-color: #d787ff; font-weight: bold">attrs</span><span style="color: #7f7f7f; text-decoration-color: #7f7f7f; font-weight: bold"> • </span><span style="color: #ffafd7; text-decoration-color: #ffafd7; font-weight: bold">2</span><span style="font-weight: bold">                       </span><span style="color: #ffafd7; text-decoration-color: #ffafd7; font-weight: bold">[Feature]</span><span style="font-weight: bold">                                                                  </span>
<span style="color: #7f7f7f; text-decoration-color: #7f7f7f">        </span>temperature                    <span style="color: #7f7f7f; text-decoration-color: #7f7f7f"> float                             </span>                                         
<span style="color: #7f7f7f; text-decoration-color: #7f7f7f">        </span>experiment_id                  <span style="color: #7f7f7f; text-decoration-color: #7f7f7f"> str                               </span>                                         
</pre>

## Technical details

- Makes the `DataFrameCurator` a slot curator but we still allow the old behavior of not specifying slots.
- We have to indicate which Schema is the Schema for the main DataFrame. Currently, we are using `df` as a fixed slot. Alternatively, we find a way to automatically use all passed Features of the main Schemas as the main Schema. Now, it's just very explicit.

### DIFF
--- a/lamindb/curators/core.py
+++ b/lamindb/curators/core.py
@@ -352,6 +352,9 @@ class SlotsCurator(Curator):
 
         if self._artifact is None:
             type_mapping = [
+                (   lambda dataset: isinstance(dataset, pd.DataFrame),
+                    Artifact.from_df
+                ),
                 (
                     lambda dataset: data_is_scversedatastructure(dataset, "AnnData"),
                     Artifact.from_anndata,
@@ -398,7 +401,7 @@ class SlotsCurator(Curator):
 # Such an approach was never intended and there is room for a DictCurator in the future.
 # For more context, read https://laminlabs.slack.com/archives/C07DB677JF6/p1753994077716099 and
 # https://www.notion.so/laminlabs/Add-a-DictCurator-2422aeaa55e180b9a513f91d13970836
-class DataFrameCurator(Curator):
+class DataFrameCurator(SlotsCurator):
     # the example in the docstring is tested in test_curators_quickstart_example
     """Curator for `DataFrame`.
 
@@ -434,6 +437,27 @@ class DataFrameCurator(Curator):
         slot: str | None = None,
     ) -> None:
         super().__init__(dataset=dataset, schema=schema)
+
+        # Handle composite schemas including attrs slots
+        if slot is None and hasattr(schema, 'slots') and schema.slots:
+            for slot_name, slot_schema in schema.slots.items():
+                if slot_name.startswith("attrs"):
+                    path_parts = slot_name.split(":")
+                    if len(path_parts) >= 1 and path_parts[0] == "attrs":
+                        attrs_dict = getattr(self._dataset, 'attrs', None)
+                        if attrs_dict is not None:
+                            if len(path_parts) == 1:
+                                df = pd.DataFrame([attrs_dict])
+                            else:
+                                deeper_keys = path_parts[1:]
+                                data = _resolve_schema_slot_path(
+                                    attrs_dict, deeper_keys, slot_name, "attrs"
+                                )
+                                df = pd.DataFrame([data])
+                            self._slots[slot_name] = DataFrameCurator(df, slot_schema, slot=slot_name)
+                # The convention for composite Schemas is that "df" specifies the main Schema
+                elif slot_name == "df":
+                    self._slots[slot_name] = DataFrameCurator(self._dataset, slot_schema, slot=slot_name)
 
         categoricals = []
         features = []
@@ -617,6 +641,10 @@ class DataFrameCurator(Curator):
     @doc_args(VALIDATE_DOCSTRING)
     def validate(self) -> None:
         """{}"""  # noqa: D415
+        if self._schema.itype == "Composite":
+            super().validate()
+            return
+
         if self._schema.n > 0:
             try:
                 # first validate through pandera
@@ -645,6 +673,10 @@ class DataFrameCurator(Curator):
         """{}"""  # noqa: D415
         if not self._is_validated:
             self.validate()  # raises ValidationError if doesn't validate
+
+        if self._schema.itype == "Composite":
+            super().save_artifact(key=key, description=description, revises=revises, run=run)
+        
         if self._artifact is None:
             self._artifact = Artifact.from_df(
                 self._dataset,
@@ -1692,7 +1724,7 @@ def get_organism_kwargs(
 def annotate_artifact(
     artifact: Artifact,
     *,
-    curator: AnnDataCurator | SlotsCurator | None = None,
+    curator: SlotsCurator | None = None,
     cat_vectors: dict[str, CatVector] | None = None,
 ) -> Artifact:
     from .. import settings
@@ -1725,13 +1757,15 @@ def annotate_artifact(
         )
 
     # annotate with inferred schemas aka feature sets
-    if artifact.otype == "DataFrame":
+    if artifact.otype == "DataFrame" and curator is not None and curator._schema.itype != "Composite":
         features = cat_vectors["columns"].records
         if features is not None:
             index_feature = artifact.schema.index
+            # DataFrameCurators with Composite Schemas must use "Feature" here
+            itype = artifact.schema.itype if not (isinstance(curator, DataFrameCurator) and curator._schema.itype == "Composite") else "Feature"
             feature_set = Schema(
                 features=[f for f in features if f != index_feature],
-                itype=artifact.schema.itype,
+                itype=itype,
                 index=index_feature,
                 minimal_set=artifact.schema.minimal_set,
                 maximal_set=artifact.schema.maximal_set,
@@ -1751,42 +1785,43 @@ def annotate_artifact(
                 feature_set.save(), through_defaults={"slot": "columns"}
             )
     else:
-        for slot, slot_curator in curator._slots.items():
-            # var_index is backward compat (2025-05-01)
-            name = (
-                "var_index"
-                if (slot == "var" and "var_index" in slot_curator.cat._cat_vectors)
-                else "columns"
-            )
-            features = slot_curator.cat._cat_vectors[name].records
-            if features is None:
-                logger.warning(f"no features found for slot {slot}")
-                continue
-            validating_schema = slot_curator._schema
-            index_feature = validating_schema.index
-            feature_set = Schema(
-                features=[f for f in features if f != index_feature],
-                itype=validating_schema.itype,
-                index=index_feature,
-                minimal_set=validating_schema.minimal_set,
-                maximal_set=validating_schema.maximal_set,
-                coerce_dtype=validating_schema.coerce_dtype,
-                ordered_set=validating_schema.ordered_set,
-            )
-            if (
-                feature_set._state.adding
-                and len(features) > settings.annotation.n_max_records
-            ):
-                logger.important(
-                    f"not annotating with {len(features)} features for slot {slot} as it exceeds {settings.annotation.n_max_records} (ln.settings.annotation.n_max_records)"
+        if curator is not None:
+            for slot, slot_curator in curator._slots.items():
+                # var_index is backward compat (2025-05-01)
+                name = (
+                    "var_index"
+                    if (slot == "var" and "var_index" in slot_curator.cat._cat_vectors)
+                    else "columns"
                 )
-                itype = parse_cat_dtype(
-                    artifact.schema.slots[slot].itype, is_itype=True
-                )["field"]
-                feature_set = Schema(itype=itype, n=len(features))
-            artifact.feature_sets.add(
-                feature_set.save(), through_defaults={"slot": slot}
-            )
+                features = slot_curator.cat._cat_vectors[name].records
+                if features is None:
+                    logger.warning(f"no features found for slot {slot}")
+                    continue
+                validating_schema = slot_curator._schema
+                index_feature = validating_schema.index
+                feature_set = Schema(
+                    features=[f for f in features if f != index_feature],
+                    itype=validating_schema.itype,
+                    index=index_feature,
+                    minimal_set=validating_schema.minimal_set,
+                    maximal_set=validating_schema.maximal_set,
+                    coerce_dtype=validating_schema.coerce_dtype,
+                    ordered_set=validating_schema.ordered_set,
+                )
+                if (
+                    feature_set._state.adding
+                    and len(features) > settings.annotation.n_max_records
+                ):
+                    logger.important(
+                        f"not annotating with {len(features)} features for slot {slot} as it exceeds {settings.annotation.n_max_records} (ln.settings.annotation.n_max_records)"
+                    )
+                    itype = parse_cat_dtype(
+                        artifact.schema.slots[slot].itype, is_itype=True
+                    )["field"]
+                    feature_set = Schema(itype=itype, n=len(features))
+                artifact.feature_sets.add(
+                    feature_set.save(), through_defaults={"slot": slot}
+                )
 
     slug = ln_setup.settings.instance.slug
     if ln_setup.settings.instance.is_remote:  # pdagma: no cover
--- a/lamindb/curators/core.py
+++ b/lamindb/curators/core.py
@@ -352,9 +352,7 @@ class SlotsCurator(Curator):
 
         if self._artifact is None:
             type_mapping = [
-                (   lambda dataset: isinstance(dataset, pd.DataFrame),
-                    Artifact.from_df
-                ),
+                (lambda dataset: isinstance(dataset, pd.DataFrame), Artifact.from_df),
                 (
                     lambda dataset: data_is_scversedatastructure(dataset, "AnnData"),
                     Artifact.from_anndata,
@@ -440,12 +438,12 @@ class DataFrameCurator(SlotsCurator):
         self._slot = slot
 
         # Handle composite schemas including attrs slots
-        if slot is None and hasattr(schema, 'slots') and schema.slots:
+        if slot is None and hasattr(schema, "slots") and schema.slots:
             for slot_name, slot_schema in schema.slots.items():
                 if slot_name.startswith("attrs"):
                     path_parts = slot_name.split(":")
                     if len(path_parts) >= 1 and path_parts[0] == "attrs":
-                        attrs_dict = getattr(self._dataset, 'attrs', None)
+                        attrs_dict = getattr(self._dataset, "attrs", None)
                         if attrs_dict is not None:
                             if len(path_parts) == 1:
                                 df = pd.DataFrame([attrs_dict])
@@ -455,10 +453,14 @@ class DataFrameCurator(SlotsCurator):
                                     attrs_dict, deeper_keys, slot_name, "attrs"
                                 )
                                 df = pd.DataFrame([data])
-                            self._slots[slot_name] = DataFrameCurator(df, slot_schema, slot=slot_name)
+                            self._slots[slot_name] = DataFrameCurator(
+                                df, slot_schema, slot=slot_name
+                            )
                 # The convention for composite Schemas is that "df" specifies the main Schema
                 elif slot_name == "df":
-                    self._slots[slot_name] = DataFrameCurator(self._dataset, slot_schema, slot=slot_name)
+                    self._slots[slot_name] = DataFrameCurator(
+                        self._dataset, slot_schema, slot=slot_name
+                    )
 
         categoricals = []
         features = []
@@ -676,8 +678,10 @@ class DataFrameCurator(SlotsCurator):
             self.validate()  # raises ValidationError if doesn't validate
 
         if self._schema.itype == "Composite":
-            super().save_artifact(key=key, description=description, revises=revises, run=run)
-        
+            super().save_artifact(
+                key=key, description=description, revises=revises, run=run
+            )
+
         if self._artifact is None:
             self._artifact = Artifact.from_df(
                 self._dataset,
@@ -1758,12 +1762,23 @@ def annotate_artifact(
         )
 
     # annotate with inferred schemas aka feature sets
-    if artifact.otype == "DataFrame" and curator is not None and curator._schema.itype != "Composite":
+    if (
+        artifact.otype == "DataFrame"
+        and getattr(curator, "_schema", None) is None
+        or getattr(curator._schema, "itype", None) != "Composite"
+    ):
         features = cat_vectors["columns"].records
         if features is not None:
             index_feature = artifact.schema.index
             # DataFrameCurators with Composite Schemas must use "Feature" here
-            itype = artifact.schema.itype if not (isinstance(curator, DataFrameCurator) and curator._schema.itype == "Composite") else "Feature"
+            itype = (
+                artifact.schema.itype
+                if not (
+                    isinstance(curator, DataFrameCurator)
+                    and curator._schema.itype == "Composite"
+                )
+                else "Feature"
+            )
             feature_set = Schema(
                 features=[f for f in features if f != index_feature],
                 itype=itype,
--- a/lamindb/curators/core.py
+++ b/lamindb/curators/core.py
@@ -437,6 +437,7 @@ class DataFrameCurator(SlotsCurator):
         slot: str | None = None,
     ) -> None:
         super().__init__(dataset=dataset, schema=schema)
+        self._slot = slot
 
         # Handle composite schemas including attrs slots
         if slot is None and hasattr(schema, 'slots') and schema.slots:
--- a/tests/curators/test_curators_examples.py
+++ b/tests/curators/test_curators_examples.py
@@ -142,22 +142,6 @@ def mudata_papalexi21_subset_schema():
 def uns_study_metadata():
     return {"temperature": 21.6, "experiment_id": "EXP001"}
 
-@pytest.fixture(scope="module")
-def study_metadata_schema():
-    study_metadata_schema = ln.Schema(
-        features=[
-            ln.Feature(name="temperature", dtype=float).save(),
-            ln.Feature(name="experiment_id", dtype=str).save(),
-        ],
-    ).save()
-
-    yield study_metadata_schema
-
-    study_metadata_schema.delete(permanent=True)
-    ln.Schema.filter().delete()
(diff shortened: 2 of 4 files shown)


falexwolf:
Thank you, Lukas!

## Design

First design.

**Design A.**

```python
study_metadata_schema = ln.Schema(
    features=[
        ln.Feature(name="temperature", dtype=float).save(),
        ln.Feature(name="experiment_id", dtype=str).save(),
    ],
).save()
df_columns_schema = ln.Schema(
    features=[ln.Feature(name="perturbation", dtype="str").save()],
).save()
schema = ln.Schema(
    slots={"df": df_columns_schema, "attrs": study_metadata_schema},
    otype="DataFrame",
).save()
```

**Design B.**

```python
study_metadata_schema = ln.Schema(
    features=[
        ln.Feature(name="temperature", dtype=float).save(),
        ln.Feature(name="experiment_id", dtype=str).save(),
    ],
).save()
schema = ln.Schema(
    features=[ln.Feature(name="perturbation", dtype="str").save()],
    slots={"attrs": study_metadata_schema},
)
```

Why did you choose Design A?

Design A has two big disadvantages:

1. it's more complex: you need a 3rd schema record, hampering construction & deletes
2. it introduces a "slot" name convention that doesn't exist: there is no such thing as `df`
3. it introduces an artificial hierarchy that doesn't exist in the schema (related to 2)

To guide a decision of why A might be preferable over B, one could investigate how `pyarrow.lib.Schema` handles it or the parquet specification.

There is vague argument around "extendibility": it's harder to re-use existing code once the user makes the decision to leverage `.attrs` -- if one could add the `.attrs` slot to an existing schema, I think that would be intuitive

There is parallelism to nested dictionaries where we need the ability to have features and slots in the same schema object. Design B would be parallel to that.

falexwolf:
## Architecture

I haven't thought this through but my intuition is that rather than eliminating the "atomicity" of the existing `DataFrameCurator` class, I'd have renamed that existing class to `PureDataFrameCurator` (or similar) and then created a new `DataFrameCurator` class that inherits from `SlotsCurator`.

Because we have one class that composes (`SlotsCurator`) we need to maintain an atomic class that is composed.

I see that you're refactoring towards recursing through the constructor but that's a lot harder to maintain than if there is one atomic class and one composed class. Given the current `DataFrameCurator` is very complicated already, we should try to avoid any unnecessary in the class.

This might also explain why you're introducing the artificial hierarchy with the ad-hoc `df` convention. 

Unless, I'm missing something, please refactor towards maintaining a pure DataFrameCurator class in some form.

falexwolf (approved):
Docs and tests look great! But design and architecture should be addressed either in this PR or right away after. We shouldn't release the `df` convention for a slot.

Zethson:
Yeah I had also thought about both designs.

> Why did you choose Design A?

It's the most explicit. I don't like `df` either.

> there is no such thing as `df`

True! There is such a thing as `values` though... Maybe that would be a better name for such a slot.

> Architecture
> 
> I haven't thought this through but my intuition is that rather than eliminating the "atomicity" of the existing DataFrameCurator class, I'd have renamed that existing class to PureDataFrameCurator (or similar) and then created a new DataFrameCurator class that inherits from SlotsCurator.
> 
> Because we have one class that composes (SlotsCurator) we need to maintain an atomic class that is composed.
> 
> I see that you're refactoring towards recursing through the constructor but that's a lot harder to maintain than if there is one atomic class and one composed class. Given the current DataFrameCurator is very complicated already, we should try to avoid any unnecessary in the class.
> 
> This might also explain why you're introducing the artificial hierarchy with the ad-hoc df convention.
> 
> Unless, I'm missing something, please refactor towards maintaining a pure DataFrameCurator class in some form.

Hmm, I don't know whether this is better. Users will see two DataFrameCurators and have to understand when to use which. We wouldn't want to hide the `PureDataFrameCurator` from the user right?

This might also blow up the number of tests and add a lot of complexity for not a lot of gain. I am totally fine with refactoring it (before merging) in a way to avoid Design A but I don't know whether this is it. :thinking:

falexwolf:
> It's the most explicit.

Why do you think it is more explicit?

> There is such a thing as values though...

That's one reason why we should use Design B. We'd create something artificial.

> We wouldn't want to hide the PureDataFrameCurator from the user right?

We _do_ want to hide it. It's a technical detail of the implementation similar to how `SlotsCurator` is.

> This might also blow up the number of tests and add a lot of complexity for not a lot of gain. 

I don't think this adds more complexity. You're right that the combination of `features is not None` and `components is not None` is novel in the `Schema` constructor, and this _is_ new complexity. However, I'm pretty sure we need this anyway to handle dictionaries properly (because a key in a dict can act as a slot and as a feature).

docs/scripts/curate_dataframe_attrs.py
```diff
@@ -0,0 +1,21 @@
...
+    features=[
+        ln.Feature(name="temperature", dtype=float).save(),
+        ln.Feature(name="experiment_id", dtype=str).save(),
+    ],
+).save()
+schema = ln.Schema(
+    features=[ln.Feature(name="perturbation", dtype="str").save()],
+    slots={"attrs": study_metadata_schema},
+    otype="DataFrame",
+).save()
```
falexwolf, line 16: Pretty! This is much nicer UX than in the previous design. ☺️

lamindb/curators/core.py
```diff
@@ -663,6 +660,144 @@ def save_artifact(
...
+        dataset: pd.DataFrame | Artifact,
+        schema: Schema,
+        slot: str | None = None,
+    ) -> None:
+        super().__init__(dataset=dataset, schema=schema)
+
+        main_schema = schema
+        skip_slot = None
+        if not schema.features.exists() and schema.slots:
+            # Find the auto-generated feature slot
```
falexwolf: Why do you handle this on the `DataFrameCurator`-level and not on the `SlotsCurator` level? Any data structure can be annotated with external features! If this is meant as a temporary solution on the way to moving this into `SlotsCurator` that's OK and could be dealt with in another PR. ☺️ 

PS: In the comment, can you refer to "external features slots" rather than just "feature slot"?

lamindb/curators/core.py
```diff
@@ -663,6 +660,144 @@ def save_artifact(
...
 
 
+class DataFrameCurator(SlotsCurator):
+    # the example in the docstring is tested in test_curators_quickstart_example
+    """Curator for `DataFrame`.
+
+    Args:
+        dataset: The DataFrame-like object to validate & annotate.
+        schema: A :class:`~lamindb.Schema` object that defines the validation constraints.
+        slot: Indicate the slot in a composite curator for a composite data structure.
```
falexwolf, line 670: What does this mean? I don't think we should have this argument. 🤔

lamindb/curators/core.py
```diff
@@ -663,6 +660,144 @@ def save_artifact(
...
+            schema=main_schema,
+            slot=slot,
+        )
+
+        # Handle additional slots if present
+        if slot is None and schema.slots:
+            for slot_name, slot_schema in schema.slots.items():
+                if slot_name == skip_slot:
+                    continue
+                if slot_name.startswith("attrs"):
```
falexwolf, line 716: Can you explain again why we need to account for composite slot keys here? A DataFrame only comes with exactly one valid slot key `"attrs"`. If the dictionary within `"attrs"` has nested structure that should be validated by mapping keys of that dictionary both on features and slots of `df_schema.slots["attrs"]` then that complexity should be handled within that sub-schema, shouldn't it?

And this is where I realize that we have another instance of the "slot-key-ambiguity" by allowing the spatialdata schema for `spatialdata-db` in the following form:
```python
ln.Schema(
    slots={
        "attrs:bio": ...
        "attrs:tech": ...
    }
)
```

Implementation-wise, this would be nicer:
```
ln.Schema(
    slots={
        "bio": ...
        "tech": ...
    }
)
ln.Schema(
    slots={
        "attrs": dict_schema
    }
)
```

But this is arguably worse UX because it creates another schema.

So, I see why you're doing it this way here, and we just need to document & test this properly.

lamindb/curators/core.py
```diff
@@ -663,6 +660,144 @@ def save_artifact(
...
+        skip_slot = None
+        if not schema.features.exists() and schema.slots:
+            # Find the auto-generated feature slot
+            for slot_name, slot_schema in schema.slots.items():
+                if slot_name == "columns" or slot_name.startswith("__external"):
+                    main_schema = slot_schema
+                    skip_slot = slot_name
+                    break
+
+        # Always create atomic curator for the main DataFrame
```
falexwolf: Nice! I thought this could be done via inheritance but I see how composition is needed and better. ☺️

lamindb/curators/core.py
```diff
@@ -264,6 +264,12 @@ def __repr__(self) -> str:
 
         cls_name = colors.green(self.__class__.__name__)
 
+        # Hide AtomicDataFrameCurator implementation detail
```
falexwolf: Why do we want to hide this important information?

lamindb/curators/core.py
```diff
@@ -1725,7 +1884,11 @@ def annotate_artifact(
         )
 
     # annotate with inferred schemas aka feature sets
-    if artifact.otype == "DataFrame":
+    if (
+        artifact.otype == "DataFrame"
+        and getattr(curator, "_schema", None) is None
+        or getattr(curator._schema, "itype", None) != "Composite"
```
falexwolf: Are you allowing `"Composite"` as a valid `itype` for a dataframe schema? I think we shouldn't. DataFrame schemas should always have the itype map onto the feature type that maps on dataframe columns. It'd be confusing to allow two ways of making dataframe schemas, don't you think?

lamindb/curators/core.py
Zethson: Users don't know what an `AtomicDataFrameCurator` is. We're hiding it completely and it's undocumented. I thought that it's more confusing than helpful.

lamindb/curators/core.py
```diff
@@ -1745,48 +1908,57 @@ def annotate_artifact(
                 logger.important(
                     f"not annotating with {len(features)} features as it exceeds {settings.annotation.n_max_records} (ln.settings.annotation.n_max_records)"
                 )
-                itype = parse_cat_dtype(artifact.schema.itype, is_itype=True)["field"]
+                itype = (
+                    Feature.name
+                    if artifact.schema.itype == "Composite"
```
falexwolf, line 1908: This if condition is the opposite of what I'd have expected, but I didn't fully parse what comes before.

lamindb/models/schema.py
```diff
@@ -578,11 +578,24 @@ def __init__(
                 self.optionals.set(optional_features)
                 return None
         self._slots: dict[str, Schema] = {}
-        # if both features and a schema are provided, we use a slot for a new schema of the features
+        # if both features and a schema are provided, we store a new schema of the features in a slot
```
falexwolf: See comments in the other PR: https://github.com/laminlabs/lamindb/pull/3039

tests/curators/test_curators_examples.py
```diff
@@ -309,6 +309,65 @@ def test_dataframe_curator_validate_all_annotate_cat2(small_dataset1_schema):
...
+    ln.ULabel(name="IFNG", type=perturbation).save()
+
+    temperature_feature = ln.Feature(name="temperature", dtype=float).save()
+    experiment_id_feature = ln.Feature(name="experiment_id", dtype=str).save()
+    study_metadata_schema = ln.Schema(
+        features=[
+            temperature_feature,
+            experiment_id_feature,
+        ],
+    ).save()
```
falexwolf: It'd be interesting if one could use some of the infra in `ln.examples` for this, in particular the `mini_immuno_schema`.

tests/curators/test_curators_examples.py
```diff
@@ -309,6 +309,65 @@ def test_dataframe_curator_validate_all_annotate_cat2(small_dataset1_schema):
     schema.delete(permanent=True)
 
 
+def test_dataframe_attrs_validation(uns_study_metadata):
```
falexwolf: Similar to how we test iteratively composing an AnnData by adding slots `.obs`, `.var`, etc. we could create a parallel test that does the same for `DataFrame`, right?

The parametrization is whether `attrs` is populated or not; and this maps well on user decision of just keeping annotations in the database or also persisiting them inside the parquet file (via `.attrs`). This is parallel to what users do with `.uns`.

lamindb/curators/core.py
falexwolf: By that logic, shouldn't you hide any internal/technical class? I don't think this is advisable. I think we should not mess with this so that people have an easy time debugging and guessing the structure of the code. I think `AtomicDataFrameCurator` should be as visible as `SlotsCurator`.

If we want a simpler name, one could also call it `SimpleDataFrameCurator`. ☺️

falexwolf: >  it's undocumented

It should feature under `lamindb.curators.core`, I think, but not under `lamindb.curators`

tests/curators/test_curators_examples.py
Zethson: I assume that you are referring to `test_anndata_curator_different_components`. I am still not a fan of these long and complex tests that iteratively compose objects. They run longer, are more difficult to debug, cannot be parallelized, and have more potential side effects. `test_anndata_curator_different_components` has 80 lines and loops through 3 slots.

I also don't even like the current `test_dataframe_attrs_validation` because it also does too many things.

Therefore, unless you strongly object, I'll keep the current, isolated test design where individual tests test specific behavior. I don't think that we need to test for user decisions as long as we cover every possible code aka `if` path.

falexwolf: > Therefore, unless you strongly object, I'll keep the current, isolated test design where individual tests test specific behavior.

Ok, then keep it! It's not a big deal given the current state of tests. 6 months from now I hope to be in a better spot and I would want to discuss this in more depth.

Reason I will want to discuss this in more depth down the road is that I disagree with your points:

> They run longer

They run shorter because they offer more potential for optimization.

> are more difficult to debug

While harder to debug for less inexperienced devs, I'd say they're easier to debug for more experienced devs because they convey the big picture rather than featuring an isolated ad hoc piece of logic.

> cannot be parallelized

This is a whole other story but AFAIK parametrized tests run in parallel. `test_anndata_curator_different_components` needs to be refactored towards proper parametrization rather than having the parametrization logic inside the test.

> has 80 lines and loops through 3 slots.

Just imagine these here were individual tests: https://github.com/laminlabs/lamindb/blob/6209a0b0e599d182fd4dabb0ef7f64c5d7ea38a4/tests/core/test_artifact.py#L589-L594

The matrix is 6 x 2 x 3 x 2 = 72 (`get_test_filepaths` returns 6 file paths that represent different configs in perfect analogy to populating a data structure with different components). Writing isolated tests for these will go much beyond ~100 lines of code for a parametrized test. Same story for `test_anndata_curator_different_components`, just that it has less degrees of freedom.

Zethson: **Yes, I am all for parametrize!** It also results in isolated tests. I am against for loops in tests.

lamindb/curators/core.py
Zethson: Yes, I am currently allowing that because a Schema that's used for `attrs` validation has an `attrs` slot. Every Schema that has a slot, becomes a `Composite` Schema. Wouldn't it be confusing if a Schema that had a slot wouldn't be `Composite` anymore?

I understand that the `itype` could have the `feature type` but I wonder whether we'd violate a different consistency here (the slot consistency).

tests/curators/test_curators_examples.py
falexwolf: Ok, glad that we understand each other! My ask was to parametrize validating a dataframe with and without the `attrs` slot. ☺️ 

> The parametrization is whether attrs is populated or not

Apologies for not being clear on that the for-loops in that one test represent half-baked parametrization that needs to be ported to `pytest.parmetrize()`. 😅

lamindb/curators/core.py
Zethson: This parameter has always been there for DataFrameCurators. I only copy and pasted this: https://github.com/laminlabs/lamindb/blob/main/lamindb/curators/core.py#L412

lamindb/curators/core.py
Zethson: **This code has been replaced.**

Nevertheless, this PR is also specifically about `DataFrame` `attrs` validation and not `external features`. An AnnData object does not have `attrs`.

There is a chance that we can harmonize more code of:
1. DF: attrs
2. Anndata: uns
3. MuData: uns & mod:uns
4. Spatialdata: attrs

by moving some code into SlotsCurator but I actually think that for now it's fine. We are explicitly validating the right slots at the places (the specific Curators) where we expect them to be. The actual ugly part, which is the nested path handling & flatting of dicts etc, is already shared.

lamindb/curators/core.py
falexwolf: Ok! Then this is for another PR. I vaguely remember why it's there.

Zethson:
@falexwolf refactored a bit and addressed your comments. Feel free to have another look if you want. Otherwise, I'd merge it now.

falexwolf:
Briefly looking.

lamindb/curators/core.py
```diff
@@ -5,6 +5,7 @@
 
    Curator
    SlotsCurator
+   AtomicDataFrameCurator
```
falexwolf: Why do you prefer the name AtomicDataFrameCurator over SimpleDataFrameCurator? Do you have a precedent that merits the more complicated language? 🤔 

While atomicity is a design principle, it's rarely a naming principle. Or is it?

lamindb/curators/core.py
```diff
@@ -1741,7 +1882,7 @@ def annotate_artifact(
         )
 
     # annotate with inferred schemas aka feature sets
-    if artifact.otype == "DataFrame":
+    if artifact.otype == "DataFrame" and getattr(curator, "_schema", None) is None:
```
falexwolf: What does this second part in the if condition mean? Can you add a comment?

lamindb/curators/core.py
Zethson: I interpret Atomic as "does one thing only" which is kind of the opposite of the Slot based Curator.

"Simple" is a term that doesn't really convey anything. I don't think that `SimpleDataFrameCurator` has more meaning.

lamindb/curators/core.py
```diff
@@ -1761,48 +1902,57 @@ def annotate_artifact(
...
-                coerce_dtype=validating_schema.coerce_dtype,
-                ordered_set=validating_schema.ordered_set,
-            )
-            if (
-                feature_set._state.adding
-                and len(features) > settings.annotation.n_max_records
-            ):
-                logger.important(
-                    f"not annotating with {len(features)} features for slot {slot} as it exceeds {settings.annotation.n_max_records} (ln.settings.annotation.n_max_records)"
+        if curator is not None:
```
falexwolf: Why is this needed now but wasn't before? We already had generic logic that annotates slot curators. Why not just remove the special casing for the data frame curator that was necessary prior to it becoming a slot curator?

I don't like the high level of indendation this causes. Maybe I'm misunderstanding but I think curator should always be not None given we're in the curator module.

tests/curators/test_curators_examples.py
```diff
@@ -305,7 +324,76 @@ def test_dataframe_curator_validate_all_annotate_cat2(small_dataset1_schema):
         "T cell",
         "B cell",
     }
+
+    artifact.delete(permanent=True)
+    schema.delete(permanent=True)
+
+
+@pytest.mark.parametrize("include_attrs_slot", [True, False])
```
falexwolf, line 332: Wow, awesome! 😄

falexwolf (approved):
Great! Mini cosmetic/philosophical remarks.

lamindb/curators/core.py
falexwolf: > I interpret Atomic as "does one thing only" which is kind of the opposite of the Slot based Curator.

Yeah, I think that's the common meaning and the design principle we're pursuing here.

> "Simple" is a term that doesn't really convey anything.

I don't agree: I think "simple" conveys that it's less complex (has fewer elements) than the user-facing DataFrameCurator.

> I don't think that SimpleDataFrameCurator has more meaning.

I think it has neither less nor more meaning. It uses simpler language to convey the same idea. Some people might say "Atomic" conveys the idea more explicitly. But I find use of simple language more important in case there is no ambiguity. And I don't see the danger for ambiguity.

---

Overall I wish we had a shorter simpler accurate name.

How about `DataFrameComponentCurator`? This would be closer to the language used in the code base. And it'd indicate that this is what the `SlotCurator` uses. We could even abbreviate to `ComponentCurator` because we only have one component curator. I fail to see how exactly things will be when/if we have a `DictCurator`.

Given all this I'd call it `ComponentCurator` in analogy with `SlotsCurator` and it will be a good choice for the upcoming months or years. If we ever have `DictCurator` in some form we could opt for more complex naming.

What do you think?

Zethson: I don't have a strong opinion and am therefore happy to name it `ComponentCurator`

lamindb/curators/core.py
Zethson: This is actually a relict from before the refactoring. I missed this. Thanks! It's gone now.

falexwolf: Great!

PR #3037, opened 2025-08-26
✨ Add `executable` type to transform

sunnyosun:
See context: https://github.com/laminlabs/laminhub/pull/3269

### DIFF
--- a/lamindb/base/types.py
+++ b/lamindb/base/types.py
@@ -35,7 +35,7 @@ ListLike = Union[list[str], pd.Series, np.array]
 StrField = Union[str, FieldAttr]  # typing.TypeAlias
 
 TransformType = Literal[
-    "pipeline", "notebook", "upload", "script", "function", "linker"
+    "pipeline", "notebook", "upload", "script", "function", "linker", "executable"
 ]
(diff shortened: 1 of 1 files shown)


lamindb/base/types.py
```diff
@@ -35,7 +35,7 @@
 StrField = Union[str, FieldAttr]  # typing.TypeAlias
 
 TransformType = Literal[
-    "pipeline", "notebook", "upload", "script", "function", "linker"
+    "pipeline", "notebook", "upload", "script", "function", "linker", "executable"
```
falexwolf, line 38: Why do we need a separate type of transform for this? Any transform should be executable.

I think we should have internal logic that determines whether something "runs natively on the hub" or not rather than coding this via a transform type.

sunnyosun: Can you make a suggestion? We thought this would be a migration-free way to both let user to filter the executable pipelines, and also let the hub to easily display those as options.

falexwolf: https://laminlabs.slack.com/archives/C07D9G0BN1J/p1756457587955029?thread_ts=1756196562.764339&cid=C07D9G0BN1J

PR #3039, opened 2025-08-26
✨ Support validating linked features: adding a `schema` argument to `.features.add_values()`

Zethson:
## `schema` arg for `FeatureManager.add_values(dict, schema)`

To validate datasets with arbitrary features and values that are not necessarily a part of the dataset, we now allow for `Feature.add_values(dict, schema)`:

```python
import lamindb as ln
df = ln.core.datasets.small_dataset1(otype="DataFrame")
artifact = ln.Artifact.from_dataframe(df, description="test dataset").save()

species = ln.Feature(name="species", dtype="str").save()
split = ln.Feature(name="split", dtype="str").save()
schema = ln.Schema([species, split]).save()
artifact.features.add_values({"species": "bird", "split": "train"}, schema=schema)
artifact.save()
artifact.describe()
```

<pre style="white-space:pre;overflow-x:auto;line-height:normal;font-family:Menlo,'DejaVu Sans Mono',consolas,'Courier New',monospace"><span style="font-weight: bold">Artifact</span><span style="color: #7f7f7f; text-decoration-color: #7f7f7f; font-weight: bold"> .parquet · DataFrame · dataset</span>
<span style="color: #7f7f7f; text-decoration-color: #7f7f7f">├── </span><span style="color: #00ffff; text-decoration-color: #00ffff; font-weight: bold">General</span>
<span style="color: #7f7f7f; text-decoration-color: #7f7f7f">│   ├── </span><span style="color: #7f7f7f; text-decoration-color: #7f7f7f">description: </span>test dataset
<span style="color: #7f7f7f; text-decoration-color: #7f7f7f">│   ├── </span><span style="color: #7f7f7f; text-decoration-color: #7f7f7f">uid: </span>Dof52Gk43vXUendP0000          <span style="color: #7f7f7f; text-decoration-color: #7f7f7f">hash: </span>aUPgpQYtBZ_zvZvl637r9g
<span style="color: #7f7f7f; text-decoration-color: #7f7f7f">│   ├── </span><span style="color: #7f7f7f; text-decoration-color: #7f7f7f">size: </span>9.6 KB                       <span style="color: #7f7f7f; text-decoration-color: #7f7f7f">transform: </span><span style="color: #00d7af; text-decoration-color: #00d7af">None</span>
<span style="color: #7f7f7f; text-decoration-color: #7f7f7f">│   ├── </span><span style="color: #7f7f7f; text-decoration-color: #7f7f7f">space: </span>all                         <span style="color: #7f7f7f; text-decoration-color: #7f7f7f">branch: </span>all
<span style="color: #7f7f7f; text-decoration-color: #7f7f7f">│   ├── </span><span style="color: #7f7f7f; text-decoration-color: #7f7f7f">created_by: </span>zethson                <span style="color: #7f7f7f; text-decoration-color: #7f7f7f">created_at: </span>2025-08-26 15:30:51
<span style="color: #7f7f7f; text-decoration-color: #7f7f7f">│   ├── </span><span style="color: #7f7f7f; text-decoration-color: #7f7f7f">n_observations: </span>3
<span style="color: #7f7f7f; text-decoration-color: #7f7f7f">│   └── </span><span style="color: #7f7f7f; text-decoration-color: #7f7f7f">storage path: </span><span style="color: #00d7af; text-decoration-color: #00d7af">/home/lukas/code/lamindb/run-tests</span>/Dof52Gk43vXUendP0000
<span style="color: #7f7f7f; text-decoration-color: #7f7f7f">└── </span><span style="color: #ff8700; text-decoration-color: #ff8700; font-weight: bold">Linked features</span>
<span style="color: #7f7f7f; text-decoration-color: #7f7f7f">    └── </span>species                        <span style="color: #7f7f7f; text-decoration-color: #7f7f7f"> str                               </span> bird                                    
<span style="color: #7f7f7f; text-decoration-color: #7f7f7f">        </span>split                          <span style="color: #7f7f7f; text-decoration-color: #7f7f7f"> str                               </span> train                                   
</pre>

## `Artifact(data, schema)`

we support a similar workflow for `Artifact`:

```python
species = ln.Feature(name="species", dtype="str").save()
split = ln.Feature(name="split", dtype="str").save()
schema = ln.Schema([species, split]).save()
artifact = ln.Artifact(
        "test_file",
        features={"species": "bird", "split": "train"},
        schema=schema,
        description="test",
    ).save()
```

## `Artifact.from_df(df, schema_with_external)`

analogously for `Artifact.from_dataframe`:

```python
species = ln.Feature(name="species", dtype="str").save()
split = ln.Feature(name="split", dtype="str").save()
external_schema = ln.Schema(features=[species, split]).save()

feat1 = ln.Feature(name="feat1", dtype="int").save()
feat2 = ln.Feature(name="feat2", dtype="int").save()
schema = ln.Schema(
    features=[feat1, feat2], slots={"__external__": external_schema}, otype="DataFrame"
).save()

artifact = ln.Artifact.from_dataframe(
    df,
    features={"species": "bird", "split": "train"},
    schema=schema,
    description="test dataframe with external features",
).save()
```

### DIFF
--- a/lamindb/models/artifact.py
+++ b/lamindb/models/artifact.py
@@ -1469,7 +1469,14 @@ class Artifact(SQLRecord, IsVersioned, TracksRun, TracksUpdates):
                         f"Composite schema has {len(schema.slots)} slots. "
                         "External feature validation only supports schemas with a single slot."
                     )
-                validation_schema = next(iter(schema.slots.values()))
+                try:
+                    validation_schema = next(
+                        k for k in schema.slots.keys() if k.startswith("__external")
+                    )
+                except StopIteration:
+                    raise ValueError(
+                        "External feature validation requires a slot that starts with __external."
+                    ) from None
 
             external_curator = DataFrameCurator(temp_df, validation_schema)
             external_curator.validate()
@@ -1890,16 +1897,26 @@ class Artifact(SQLRecord, IsVersioned, TracksRun, TracksUpdates):
                 )
                 return artifact
 
+            # Handle external features validation for Composite schemas
             if schema.itype == "Composite" and features is not None:
-                if schema is not None and schema.itype == "Composite":
-                    validation_schema = next(iter(schema.slots.values()))
-                    temp_df = pd.DataFrame([features])
-                    external_curator = DataFrameCurator(temp_df, validation_schema)
-                    external_curator.validate()
+                try:
+                    external_slot = next(
+                        k for k in schema.slots.keys() if "external" in k
+                    )
+                    validation_schema = schema.slots[external_slot]
+                except StopIteration:
+                    raise ValueError(
+                        "External feature validation requires a slot that starts with __external."
+                    ) from None
 
+                external_curator = DataFrameCurator(
+                    pd.DataFrame([features]), validation_schema
+                )
+                external_curator.validate()
                 artifact._external_features = features
-            else:
-                # Regular schema means DataFrame validation
+
+            # Validate main DataFrame if not Composite or if Composite has attrs
+            if schema.itype != "Composite" or "attrs" in schema.slots:
                 curator = DataFrameCurator(artifact, schema)
                 curator.validate()
                 artifact.schema = schema
--- a/lamindb/models/schema.py
+++ b/lamindb/models/schema.py
@@ -578,10 +578,11 @@ class Schema(SQLRecord, CanCurate, TracksRun):
                 self.optionals.set(optional_features)
                 return None
         self._slots: dict[str, Schema] = {}
-        # if both features and a schema are provided, we use an internal slot for a new schema of the features
+        # if both features and a schema are provided, we use a slot for a new schema of the features
         if features and slots:
             main_schema = Schema(features=features).save()
-            slots["__features__"] = main_schema
+            slot_name = f"__external_{name}__" if name else "__external__"
+            slots[slot_name] = main_schema
             features = []
 
         if features:


lamindb/models/_feature_manager.py
```diff
@@ -841,59 +841,70 @@ def add_values(
...
-        # figure out which of the values go where
+        if schema is not None:
+            records = schema.members.filter(name__in=keys)
+            if len(records) != len(keys):
+                missing_keys = [
+                    key
+                    for key in keys
+                    if key not in records.values_list("name", flat=True)
+                ]
+                raise ValueError(
```
falexwolf: I think this should probably be a `ValidationError`.

There is a bigger question though of whether you could leverage `DataFrameCurator` aka `DictCurator` to provide this logic rather than re-implementing it?

Why not do:
```
if schema is not None:
    curator = DataFrameCurator(values, schema)
    curator.validate()
```

This is what I had in mind when saying "there is a lot of overlap between building DictCurator and validating add_values()". 🤔 

Because currently we don't have the full-fledged DictCurator but the way in which dict-validation is flawed would at least be consistently improved by working on the upstream logic blocks.

If we re-implement here from scratch we'll have two construction sites, right? ☺️

lamindb/models/artifact.py
```diff
@@ -2811,11 +2820,19 @@ def save(
                 local_path_cache,
             )
             logger.important(f"moved local artifact to cache: {local_path_cache}")
+
+        # annotate Artifact
         if hasattr(self, "_curator"):
             curator = self._curator
             delattr(self, "_curator")
-            # just annotates this artifact
```
falexwolf, line 2817: I think this comment was helpful ☺️

lamindb/models/artifact.py
```diff
@@ -1445,15 +1444,24 @@ def __init__(
...
         description: str | None = kwargs.pop("description", None)
         revises: Artifact | None = kwargs.pop("revises", None)
         overwrite_versions: bool | None = kwargs.pop("overwrite_versions", None)
         version: str | None = kwargs.pop("version", None)
+
+        features: dict[str, Any] = kwargs.pop("features", None)
+        schema: Schema | None = kwargs.pop("schema", None)
+        if features is not None:
+            self._staged_features = features
+            self._staged_schema = schema
```
falexwolf: The way this is done for the `from_...()` constructors is this:

https://github.com/laminlabs/lamindb/blob/e84e1d9b5cdbc78ad6772f1889d2ec11fe04717f/lamindb/models/artifact.py#L1856

I'd suggest we do it the same way for blob-like (path-derived) artifacts.

You create a `_curator = Curator(features, schema)` object, validate() and validate() it upon construction in complete analogy to `from_...()` et al

This way will also allow to completely abstract out the validation logic at some point (out-of-scope for this PR).

lamindb/models/_feature_manager.py
```diff
@@ -861,39 +861,48 @@ def add_values(
...
-            )
-            raise ValidationError(msg)
 
-        # figure out which of the values go where
+        if schema is not None:
+            from lamindb.curators import DataFrameCurator
+
+            temp_df = pd.DataFrame([values])
+            curator = DataFrameCurator(temp_df, schema)
+            curator.validate()
```
falexwolf, line 874: Great!

lamindb/models/artifact.py
```diff
@@ -1445,15 +1444,39 @@ def __init__(
...
+
+            temp_df = pd.DataFrame([features])
+            validation_schema = schema
+            if schema.itype == "Composite" and schema.slots:
+                if len(schema.slots) > 1:
+                    raise ValueError(
+                        f"Composite schema has {len(schema.slots)} slots. "
+                        "External feature validation only supports schemas with a single slot."
+                    )
+                validation_schema = next(iter(schema.slots.values()))
```
falexwolf: Here we should use a naming convention along the lines of: https://laminlabs.slack.com/archives/C04FPE8V01W/p1756709498076439?thread_ts=1755702104.978259&cid=C04FPE8V01W

Instead of `next(iter())`

lamindb/models/artifact.py
```diff
@@ -1863,10 +1890,21 @@ def from_dataframe(
...
+            if schema.itype == "Composite" and features is not None:
+                if schema is not None and schema.itype == "Composite":
+                    validation_schema = next(iter(schema.slots.values()))
+                    temp_df = pd.DataFrame([features])
+                    external_curator = DataFrameCurator(temp_df, validation_schema)
+                    external_curator.validate()
+
+                artifact._external_features = features
+            else:
+                # Regular schema means DataFrame validation
```
falexwolf: Need to pay attention to the edge case where `.attrs` needs to be validated in addition to an external slot.

lamindb/models/schema.py
```diff
@@ -578,6 +578,12 @@ def __init__(
                 self.optionals.set(optional_features)
                 return None
         self._slots: dict[str, Schema] = {}
+        # if both features and a schema are provided, we use an internal slot for a new schema of the features
+        if features and slots:
+            main_schema = Schema(features=features).save()
+            slots["__features__"] = main_schema
```
falexwolf: https://laminlabs.slack.com/archives/C04FPE8V01W/p1756709498076439?thread_ts=1755702104.978259&cid=C04FPE8V01W

tests/core/test_artifact.py
```diff
@@ -337,6 +337,58 @@ def test_revise_artifact(df):
...
 
+def test_create_external_schema(tsv_file):
+    species = ln.Feature(name="species", dtype="str").save()
+    split = ln.Feature(name="split", dtype="str").save()
+    schema = ln.Schema(features=[species, split]).save()
+
+    artifact = ln.Artifact(
+        tsv_file,
+        features={"species": "bird", "split": "train"},
+        schema=schema,
```
falexwolf, line 348: What happens here if I don't pass a schema? How about parametrizing this test with `schema in [ln.Schema(features=[species, split]).save(), None]`?

tests/core/test_artifact.py
```diff
@@ -337,6 +337,58 @@ def test_revise_artifact(df):
     artifact.delete(permanent=True)  # permanent deletion
 
 
+def test_create_external_schema(tsv_file):
+    species = ln.Feature(name="species", dtype="str").save()
```
falexwolf, line 341: FYI: idiomatic would be to _not_ quote `"str"`, but just use `str` -- no need to change of course

tests/core/test_artifact.py
```diff
@@ -337,6 +337,58 @@ def test_revise_artifact(df):
...
+
+def test_from_dataframe_external_schema(df):
+    species = ln.Feature(name="species", dtype="str").save()
+    split = ln.Feature(name="split", dtype="str").save()
+    external_schema = ln.Schema(features=[species, split]).save()
+
+    feat1 = ln.Feature(name="feat1", dtype="int").save()
+    feat2 = ln.Feature(name="feat2", dtype="int").save()
+    schema = ln.Schema(
+        features=[feat1, feat2], slots={"external": external_schema}, otype="DataFrame"
```
falexwolf, line 367: Why is this `"external"` and not `"__external__"` as in the spec discussed yesterday?

https://laminlabs.slack.com/archives/C04FPE8V01W/p1756709498076439?thread_ts=1755702104.978259&cid=C04FPE8V01W

lamindb/models/schema.py
```diff
@@ -578,6 +578,13 @@ def __init__(
                 self.optionals.set(optional_features)
                 return None
         self._slots: dict[str, Schema] = {}
+        # if both features and a schema are provided, we use a slot for a new schema of the features
+        if features and slots:
+            main_schema = Schema(features=features).save()
+            slot_name = f"__external_{name}__" if name else "__external__"
```
falexwolf, line 584: Why does this slot inherit the `name` of the schema? Why not just `__external__`?

lamindb/models/schema.py
```diff
@@ -578,6 +578,13 @@ def __init__(
                 self.optionals.set(optional_features)
                 return None
         self._slots: dict[str, Schema] = {}
+        # if both features and a schema are provided, we use a slot for a new schema of the features
+        if features and slots:
+            main_schema = Schema(features=features).save()
```
falexwolf, line 583: This is not OK because you're saving an object to the database in the constructor. We're avoiding this everywhere else in the codebase. When the user calls `Schema()`, they expect this to be idempotent aka not mutate the database. This has to be changed.

However, I don't understand the rationale for the design choice of implicitly creating a schema component in the first place.

I think I just don't get the design. 🤔 I'm tempted to argue along the parallel argument that a dataframe schema should, IMO, respect the hierarchy of the data structure and not introduce artificial hierarchy. But I'm not sure whether this applies.

I'll try to understand this better by looking at examples.

lamindb/models/schema.py
```diff
@@ -578,6 +578,13 @@ def __init__(
                 self.optionals.set(optional_features)
                 return None
         self._slots: dict[str, Schema] = {}
+        # if both features and a schema are provided, we use a slot for a new schema of the features
```
falexwolf, line 581: This wants to say "and a slot", right?

tests/core/test_artifact.py
```diff
@@ -337,6 +337,58 @@ def test_revise_artifact(df):
...
+        description="test dataframe with external features",
+    ).save()
+
+    assert artifact.features.get_values() == {"species": "bird", "split": "train"}
+
+    # Cleanup
+    artifact.delete(permanent=True)
+    schema.delete(permanent=True)
+    external_schema.delete(permanent=True)
+    for feature in [species, split, feat1, feat2]:
```
falexwolf, line 383: This line should not be needed as deleting a schema cascades to deleting the schema-feature links. Are you sure it's needed?

lamindb/models/schema.py
falexwolf: So, it fires in this example:

image

And I'm not pretty sure that the feedback I gave for https://github.com/laminlabs/lamindb/pull/3036 applies here, too: we should keep the number of schema records minimal to keep the database complexity low. So, I'm pretty sure that we should _not_ create an additional schema (and definitely not implicitly save it to the database).

Why did you choose the design you chose?

tests/core/test_artifact.py
falexwolf: What lacks here are tests for the appropriate construction of the `schema` object. In your case that would be the implicit "auxiliary schema". But I think that design should be changed to a schema that has `schema.features` and `schema.components` equal what's passed to constructor.

falexwolf:
Thanks for the improvements, Lukas!

I think there is one rather serious design flaw in the comments, let's please discuss in the comments. I think the design needs to be refactored on that point.

The PR lacks documentation: the examples in the PR description should show in the docs and could then be screenshotted to document the PR.

There are also slight inaccuracies in the documentation: for instance, the word "annotation" should be replaced by "validatation"; I fixed some of these already. Just give it another read before you add it to the docs. ☺️ 

image

tests/core/test_artifact.py
Zethson: > Why is this "external" and not "__external__" as in the spec discussed yesterday?

That's legacy because the initial example had that. Can be adapted.

>  But I think that design should be changed to a schema that has schema.features and schema.components equal what's passed to constructor.

That's what I asked about in https://laminlabs.slack.com/archives/C04FPE8V01W/p1756394131780189?thread_ts=1755702104.978259&cid=C04FPE8V01W and I understood your answer as: "let's create an `__external__` Schema.

lamindb/models/schema.py
Zethson: That's what I asked about it in https://laminlabs.slack.com/archives/C04FPE8V01W/p1756394131780189?thread_ts=1755702104.978259&cid=C04FPE8V01W. For the reasons that you mentioned, I also didn't like that approach but then I took your answers in the thread as "go for it and give it the `__external__` name. I think there's a different approach though.

Zethson:
> The PR lacks documentation: the examples in the PR description should show in the docs and could then be screenshotted to document the PR.

This is coming with a new PR that addresses fixtures and documentation of all recent changes.

tests/core/test_artifact.py
Zethson: Otherwise, there's

```
E           django.db.models.deletion.ProtectedError: ("Cannot delete some instances of model 'Feature' because they are referenced through protected foreign keys: 'SchemaFeature.feature'.", {<SchemaFeature: SchemaFeature object (5)>})
```
errors.

tests/core/test_artifact.py
falexwolf: Oh, I see how you thought that! Seeing your snippet on Slack I didn't realize that you wanted to create an auxiliary schema implicitly in the schema constructor.

I was just talking about an explicit naming convention for explicit schema creation as in the design spec:

image

I'm sorry, I didn't parse your question correctly. 😔

tests/core/test_artifact.py
falexwolf: This error indicates that you didn't delete the schemas that should be deleted at the end of this test. Is it possible that the implicitly created schema is the culprit?

Can you please investigate so that we can remove these ad hoc lines?

Zethson: We can indeed get rid of it with the alternative design: https://github.com/laminlabs/lamindb/pull/3052

PR #3048, opened 2025-09-01
📝 Improve documentation for "slots"

Zethson:
Addresses https://github.com/laminlabs/lamindb/pull/3029#issuecomment-3206757421

Adds more details about what slots are and adds current limitations of nested slot validation to curation guide.


falexwolf:
> scverse datastructures

I'd not advise to spell as one word. If you want to, it has to be capitalized or something like this.

I'd say "like the scverse data structures"

> compartment

Why a new word? We have "component" for this in the API and I think it's more established in this context.

> contains each feature set

Why not say "schema"? At this point people know what a schema is. The term "feature set" is legacy / only used to explain schemas.

...

I have several more questions. I'll rewrite and let you review. ☺️

Zethson:
Thanks! I'm fine with the changes.

PR #3051, opened 2025-09-02
✨ Implement lazy artifact saving for streaming data

Koncopd:
For https://github.com/laminlabs/pfizer-lamin-usage/issues/335

so this should work like this:
```python
lazy_artifact = ln.Artifact.from_lazy(suffix=".zarr", overwrite_versions=True, key="myzarr.zarr")
lazy_artifact.path # has the path in root/.lamindb/
... # stream into the path
artifact= lazy_artifact.save() # get the actual artifact
```

### DIFF
--- a/lamindb/models/__init__.py
+++ b/lamindb/models/__init__.py
@@ -9,6 +9,7 @@
    BasicQuerySet
    QuerySet
    ArtifactSet
+   LazyArtifact
    QueryManager
    SQLRecordList
    FeatureManager
@@ -49,7 +50,7 @@ from .schema import Schema
 from .ulabel import ULabel
 
 # should come last as it needs everything else
-from .artifact import Artifact
+from .artifact import Artifact, LazyArtifact
 from ._feature_manager import FeatureManager
 from ._label_manager import LabelManager
 from .collection import Collection, CollectionArtifact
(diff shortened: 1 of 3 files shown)


falexwolf:
This is cool! 

The one thing I'd change is rename `.lazy_artifact()` to `.from_lazy()` or something similar so that it's clear that this is a constructor similar (but different) to the other constructors.

Koncopd:
But it is different from other `.from_*` functions, it doesn't require any input of data, it just returns a lazy object.

falexwolf:
> But it is different from other .from_* functions, it doesn't require any input of data, it just returns a lazy object.

I agree. I still think `from_lazy()` is the better naming convention.

Koncopd:
But from here is super confusing, no? Maybe .as_lazy or .as_lazy_artifact?

falexwolf:
> Maybe .as_lazy or .as_lazy_artifact?

This would indicate an instance method to me.

> But from here is super confusing, no? 

I think it's intuitive because you're creating an artifact-like object and it's 100% a parallel alternative to making an Artifact like the `Artifact()` constructor and the other `from_...()` class methods. So, I think it fits in neatly.

Koncopd:
Ok, i have renamed the function.

Koncopd:
Ok, i assume it is fine and merge.

PR #3052, opened 2025-09-02
📝 Add validating linked features guide & examples

Zethson:
Addresses feedback raised in https://github.com/laminlabs/lamindb/pull/3029

- No longer saving an additional Schema under the hood
- Adds examples to docstrings and the curate guide

image


lamindb/models/_feature_manager.py
```diff
@@ -849,6 +849,13 @@ def add_values(
             values: A dictionary of keys (features) & values (labels, numbers, booleans).
             feature_field: The field of a reference registry to map keys of the dictionary.
             schema: Schema to validate against.
+
+        Examples:
+
+            To validate external features::
```
falexwolf: This example is great but should likely be moved here:

image

Or, alternatively, the stuff that's currently shown here https://docs.lamin.ai/lamindb.artifact#lamindb.Artifact.features should be moved here.

lamindb/models/artifact.py
```diff
@@ -1092,6 +1092,11 @@ class Artifact(SQLRecord, IsVersioned, TracksRun, TracksUpdates):
             schema = ln.Schema(itype=ln.Feature)  # a schema that merely enforces that feature names exist in the Feature registry
             artifact = ln.Artifact.from_dataframe("./my_file.parquet", key="my_dataset.parquet", schema=schema).save()  # validated and annotated
 
+        To validate & annotate by **external features**, pass a `schema`::
+
+            schema = ln.Schema(itype=ln.Feature).save()
```
falexwolf: Can you use this instead? https://docs.lamin.ai/lamindb.examples.schemas.valid_features

lamindb/models/artifact.py
```diff
@@ -1092,6 +1092,11 @@ class Artifact(SQLRecord, IsVersioned, TracksRun, TracksUpdates):
             schema = ln.Schema(itype=ln.Feature)  # a schema that merely enforces that feature names exist in the Feature registry
             artifact = ln.Artifact.from_dataframe("./my_file.parquet", key="my_dataset.parquet", schema=schema).save()  # validated and annotated
 
+        To validate & annotate by **external features**, pass a `schema`::
```
falexwolf: What if I only want to annotate but don't care about validation?

lamindb/models/artifact.py
falexwolf: Wait: This isn't even needed because `artifact.features.add_values()` always expects this. Hence, `valid_features` is a requirement anyway and passing this schema is pointless given the way lamindb works. 🤔 

So, it comes back to: https://github.com/laminlabs/lamindb/pull/3052/files#r2318909767

tests/core/test_artifact.py
```diff
@@ -337,34 +336,50 @@ def test_revise_artifact(df):
     artifact.delete(permanent=True)  # permanent deletion
 
 
-def test_create_external_schema(tsv_file):
-    species = ln.Feature(name="species", dtype="str").save()
-    split = ln.Feature(name="split", dtype="str").save()
-    schema = ln.Schema(features=[species, split]).save()
-
+@pytest.mark.parametrize(
```
falexwolf, line 339: Great!

lamindb/models/artifact.py
Zethson: Hmm, currently, nothing happens to the passed features if a schema has not been passed. I think this could be done but is this really a pattern that we want to encourage?

falexwolf: Yes, just like `.features.add_values()` should it be encouraged and there is no problem with it (see https://github.com/laminlabs/lamindb/pull/3052#discussion_r2318912683).

> nothing happens to the passed features if a schema has not been passed

I think this isn't intuitive. I think it should behave like `.features.add_values()` without a schema being passed. We're talking here about a one-linification of this
```python
artifact = Artifact(...).save()
artifact.features.add_values()
```

A customer requested it and I also thought it was strange that I often found myself doing:
```python
artifact = Artifact.from_anndata(adata, schema=...).save()
artifact.features.add_values(adata.uns)
```

Zethson: This is now supported and tested.

PR #3060, opened 2025-09-06
Featurequeryusing

falexwolf:


lamindb/models/query_set.py
```diff
@@ -921,6 +922,40 @@ def get(self, idlike: int | str | None = None, **expressions) -> SQLRecord:
...
         """Query a set of records."""
+        if expressions and self.model.__name__ == "Artifact":
+            from ._feature_manager import filter_base
+            from .artifact import Artifact
+            from .feature import Feature
+
+            keys_normalized = [key.split("__")[0] for key in expressions]
+            field_or_feature = keys_normalized[0].split("__")[0]
+            if field_or_feature in Artifact.__get_available_fields__():
+                qs = super().filter(*queries, **expressions)
```
Koncopd, line 938: i suspect we want to call `QuerySet.filter` here, but `BaseQuerySet.filter`, so moving the code block is maybe undesirable.

Koncopd:
We should be really careful with any `.filter` changes.

lamindb/models/query_set.py
Koncopd: Interactions between different filter types (`QuerySet.filter`, `BaseQuerySet.filter`, `.objects.filter`) are pretty complicated.

falexwolf:
This attempts to fix this issue, and I think this is needed: https://github.com/laminlabs/pfizer-lamin-usage/issues/339

Do you see what the remaining problem is?

Koncopd:
I have an idea, but can't really work on this right now.

falexwolf:
Oh, I think it's just that we exclude private fields 🤔 

image

There doesn't seem to be any basic problem here?

Koncopd:
I don't really thing this should go into `QuerySet.filter`.

Koncopd:
I am already leaning to using `.object.filter` when i need to query because i feel that `QuerySet.filter` became pretty unpredictable.

lamindb/models/query_set.py
```diff
@@ -921,6 +922,44 @@ def get(self, idlike: int | str | None = None, **expressions) -> SQLRecord:
...
     def filter(self, *queries, **expressions) -> QuerySet:
         """Query a set of records."""
+        if expressions and self.model.__name__ == "Artifact":
+            from ._feature_manager import filter_base
+            from .artifact import Artifact
+            from .feature import Feature
+
+            keys_normalized = [key.split("__")[0] for key in expressions]
+            field_or_feature = keys_normalized[0].split("__")[0]
+            available_fields = {f.name for f in Artifact._meta.get_fields()}
```
Koncopd, line 932: No, this doesn't help, the problem is from BaseQuerySet.filter called via super, not from QuerySet.filter, so this doesn't have any effect at all. I don't think this is needed at all.

falexwolf: This is needed to decide whether we have a feature-based or a Django-based query. And yes, it fixed the initial issues in this PR.

falexwolf:
> I am already leaning to using .object.filter when i need to query because i feel that QuerySet.filter became pretty unpredictable.

I'm pretty sure you're talking about low-level queries. And I think it's the right way to use `.objects.filter()` for those. `Registry.filter()` is a high-level LaminDB construct that will soon be mostly used to query by features rather than by fields.

falexwolf:
> I don't really thing this should go into QuerySet.filter.

Where do you think it should go to enable this:

```python
ln.Artifact.using("laminlabs/cellxgene").filter(n_of_donors__gte=100).to_dataframe()
```

Koncopd:
I think to make this change cleanly the artifact part / features part of the filter should be split from non-artifact part and should call this non-artifact path. Something like this.

Koncopd:
I can try doing this somewhat later.

Koncopd:
Ok, trying to figure out now.

Koncopd:
Hm, i think using `filter_base` inside `QuerySet` also doesn't make much sense because `QuerySet` can be filtered already, and `filter_base` just discards that.

Koncopd:
This adds a ton of ad hoc things to available fields, i strongly believe this is not how it should be done.

Koncopd:
Just to be more specific, this variant basically makes it impossible to use .alias or .annotate and then .filter on the annotation for query sets of artifacts.

falexwolf:
Yes, agreed on all fronts.

I'll study your PR in more detail.

PR #3061, opened 2025-09-06
✨ Query by feature across instances

Koncopd:
This PR makes it possible to query by features across instances:
```python
ln.Artifact.using("laminlabs/cellxgene").filter(n_of_donors__gte=100)
```

Supersedes:

- https://github.com/laminlabs/lamindb/pull/3060

Fixes https://github.com/laminlabs/pfizer-lamin-usage/issues/339

### DIFF
--- a/lamindb/models/artifact_set.py
+++ b/lamindb/models/artifact_set.py
@@ -13,7 +13,6 @@ from ..core._mapped_collection import MappedCollection
 from ..core.storage._backed_access import _open_dataframe
 from .artifact import Artifact, _track_run_input
 from .collection import Collection, _load_concat_artifacts
-from .query_set import BasicQuerySet
 
 if TYPE_CHECKING:
     from anndata import AnnData
@@ -129,7 +128,8 @@ class ArtifactSet(Iterable):
 
 def artifacts_from_path(artifacts: ArtifactSet, path: UPathStr) -> ArtifactSet:
     """Returns artifacts in the query set that are registered for the provided path."""
-    assert type(artifacts) is BasicQuerySet  # not QuerySet # noqa: S101
+    # not QuerySet but BasicQuerySet
+    assert type(artifacts).__name__ in {"BasicQuerySet", "ArtifactBasicQuerySet"}  # noqa: S101
 
     upath = UPath(path)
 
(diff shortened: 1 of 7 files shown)


falexwolf:
I think this _might_ be cleaner, yes. Bug looking at test breakages more things seem to be shady, WDYT? I'd have a pretty good idea how to fix tests in my PR, but don't really know what's going on here.

The `hasattr()` check is not pretty, but it'd be simple to replace with `isinstance(..., ArtifactSet)` I imagine.

Koncopd:
I understand pretty well why the tests are failing here, the problem is the same in your PR actually i believe. 
There are multiple problems actually.

Koncopd:
Explaining all the problems requires a call i think, this got pretty complicated.

Koncopd:
Both variants are quite bad actually.

Koncopd:
@falexwolf i think we should not merge any of this this weekend. This requires some further thinking, this is more serious than you think, at least i believe so.

Koncopd:
What is left here is adapting `filter_base` to properly accept `QuerySet` as an input, not only `Registry`.

But i really don't like both variants.

falexwolf:
> @falexwolf i think we should not merge any of this this weekend. This requires some further thinking, this is more serious than you think, at least i believe so.

I believe you. Let's call Monday morning!

falexwolf:
At least your solution is a lot nicer already than what I naively thought was going to work.

lamindb/models/artifact_set.py
```diff
@@ -40,6 +46,47 @@ class ArtifactSet(Iterable):
         >>> artifacts # an instance of ArtifactQuerySet inheriting from ArtifactSet
     """
 
+    def _filter_with_features(self, *queries, **expressions) -> ArtifactSet:
```
falexwolf: One issue with this is that `Record` and `Run` relates to features in the same way as `Artifact`.

All of `Artifact`, `Record`, and `Run` should be queryable features with the same mechanism. In fact, `ln.Record.to_dataframe(features=True)` already works in the same way as `ln.Artifact.to_dataframe(features=True)`, creating a table in which features map into columns of that table.

But this just for context for where the implementation should go in the future.

Koncopd: I have added this.

falexwolf (approved):
Cool! You managed to get this to be really pretty!

PR #3073, opened 2025-09-09
♻️ Extend soft deletes to querysets

Zethson:
Implement unified soft delete behavior for querysets

- Querysets now support soft deletion matching individual record behavior  
- Added `permanent` parameter to querysets for explicit hard deletion
- Provides consistent deletion API across all data access patterns

### DIFF
--- a/tests/curators/test_curators_examples.py
+++ b/tests/curators/test_curators_examples.py
@@ -139,6 +139,7 @@ def mudata_papalexi21_subset_schema():
         af.delete(permanent=True)
     ln.Schema.filter().delete(permanent=True)
     ln.Feature.filter().delete(permanent=True)
+    bt.models.SchemaGene.filter().delete()
     bt.Gene.filter().delete(permanent=True)
     ln.ULabel.filter(type__isnull=False).delete(permanent=True)
     ln.ULabel.filter().delete(permanent=True)
@@ -179,6 +180,7 @@ def spatialdata_blobs_schema():
     SchemaComponent.filter().delete(permanent=True)
 
     ln.Schema.filter().delete(permanent=True)
+    bt.models.SchemaGene.filter().delete()
     bt.Gene.filter().delete(permanent=True)
     ln.ULabel.filter(type__isnull=False).delete(permanent=True)
     ln.ULabel.filter().delete(permanent=True)


lamindb/models/query_set.py
```diff
@@ -778,17 +778,39 @@ def df(
...
+
+            For any `QuerySet` object `qs`, call:
+
+            >>> qs.delete()
+        """
         from lamindb.models import Artifact, Collection, Run, Storage, Transform
 
-        # both Transform & Run might reference artifacts
-        if self.model in {Artifact, Collection, Transform, Run, Storage}:
+        if len(self) < 10000:
```
falexwolf: This number is too small, I'd make it 1. Is there any reason not to make it 1?

lamindb/models/query_set.py
```diff
@@ -778,17 +778,39 @@ def df(
...
+                # both Transform & Run might reference artifacts
+                if self.model in {Artifact, Collection, Transform, Run, Storage}:
+                    logger.important(f"deleting {record}")
+                if isinstance(record, SQLRecord):
+                    record.delete(*args, permanent=permanent, **kwargs)  # type: ignore
+                else:
+                    record.delete(*args, **kwargs)
         else:
-            super().delete(*args, **kwargs)
+            if not permanent:
```
falexwolf: For individual record deletion we're logging that it moves the record to the trash. I'd an analogous logging statement here.

https://github.com/laminlabs/lamindb/blob/1d840ce5359eeba8cae49962229edb7000656ee3/lamindb/models/sqlrecord.py#L1202

lamindb/models/save.py
```diff
@@ -47,11 +47,10 @@ def save(
 
     Args:
         records: Multiple :class:`~lamindb.models.SQLRecord` objects.
-        ignore_conflicts: If ``True``, do not error if some records violate a
-            unique or another constraint. However, it won't inplace update the id
-            fields of records. If you need records with ids, you need to query
-            them from the database.
-        batch_size: Number of records to process in each batch. Defaults to 10000.
+        ignore_conflicts: If ``True``, do not error if some records violate a unique or another constraint.
```
falexwolf: Given you're editing, can remove double-backticks ☺️

lamindb/models/query_set.py
Zethson: 1? Do you mean 100000? 1 million? If we increase that number, it becomes harder to test because we'd create a ton of records which takes time.

lamindb/models/query_set.py
```diff
@@ -778,17 +778,46 @@ def df(
...
         from lamindb.models import Artifact, Collection, Run, Storage, Transform
 
-        # both Transform & Run might reference artifacts
-        if self.model in {Artifact, Collection, Transform, Run, Storage}:
+        if len(self) < 10000:
             for record in self:
-                logger.important(f"deleting {record}")
-                record.delete(*args, **kwargs)
+                # both Transform & Run might reference artifacts
+                if self.model in {Artifact, Collection, Transform, Run, Storage}:
```
Zethson: Unclear to me whether we should always omit this message or only in this case. I'm fine with keeping it restricted to these major Records but please tell me if we should always have it.

lamindb/models/query_set.py
```diff
@@ -778,17 +778,46 @@ def df(
...
+                if isinstance(record, SQLRecord):
+                    record.delete(*args, permanent=permanent, **kwargs)  # type: ignore
+                    logger.warning(
+                        f"moved {len(self)} records to trash (branch_id = -1)"
+                    )
+                else:
+                    record.delete(*args, **kwargs)
         else:
-            super().delete(*args, **kwargs)
+            if not permanent:
```
Zethson: Should we also omit a warning here that we're moving a queryset to the trash?

lamindb/models/query_set.py
falexwolf: I mean 1. For any querset that's more than a single record I'd _not_ use a Python loop that makes network requests because that is unbearably slow.

lamindb/models/query_set.py
falexwolf: I think the messages should be consistent across record types, but I might misunderstand what you mean.

lamindb/models/query_set.py
falexwolf: I would not omit the warning but print it analogous to https://github.com/laminlabs/lamindb/pull/3073/files#r2333709379

But maybe I misunderstand, happy to call!

lamindb/models/query_set.py
Zethson: No, this is fine. Then the number was "too large" and not too small. I tripped over that comment of yours.

Zethson:
Hmm, I cannot reproduce these 3 failures locally....

lamindb/models/query_set.py
```diff
@@ -778,17 +778,35 @@ def df(
...
+        """
         from lamindb.models import Artifact, Collection, Run, Storage, Transform
 
-        # both Transform & Run might reference artifacts
+        # all these models have non-trivial delete behavior, hence we need to handle in a loop
         if self.model in {Artifact, Collection, Transform, Run, Storage}:
             for record in self:
-                logger.important(f"deleting {record}")
-                record.delete(*args, **kwargs)
+                record.delete(*args, permanent=permanent, **kwargs)  # type: ignore
```
Koncopd, line 803: this is incorrect for `Storage`, it doesn't have `permanent`.

falexwolf: Great catch, Sergei!

PR #3077, opened 2025-09-10
🐛 Better warnings during queryset soft delete

falexwolf:


lamindb/models/query_set.py
```diff
@@ -821,14 +821,25 @@ def delete(self, *args, permanent: bool | None = None, **kwargs):
...
         # all these models have non-trivial delete behavior, hence we need to handle in a loop
-        if self.model in {Artifact, Collection, Transform, Run, Storage}:
+        if self.model in {Artifact, Collection, Transform, Run}:
             for record in self:
-                record.delete(*args, permanent=permanent, **kwargs)  # type: ignore
+                record.delete(*args, permanent=permanent, **kwargs)
+        elif self.model is Storage:  # storage does not have soft delete
+            if permanent is False:
+                logger.warning(
+                    "the Storage registry doesn't support soft delete, hard deleting"
```
Koncopd, line 830: I think it has `branch` though.

Koncopd (approved):

lamindb/models/query_set.py
falexwolf: Yes, but it still doesn't support soft delete due to the complex requirements for deleting a storage (in particular the artifacts within).

PR #3092, opened 2025-09-22
🐛 Throw a clearer error when creating a schema without features & components and `is_type=True`

sunnyosun:
When creating a schema record, one of the following must be provided:
- `features`
- `slot`
- `itype`
- `is_type=True`

### DIFF
--- a/tests/core/test_schema.py
+++ b/tests/core/test_schema.py
@@ -548,7 +548,7 @@ def test_schema_already_saved_aux():
 
 
 def test_schema_not_saved_describe():
-    schema = ln.Schema(name="not_saved_schema", itype=ln.Feature)
+    schema = ln.Schema(name="NotSavedSchema", is_type=True)
     with pytest.raises(ValueError) as e:
         schema.describe()
     assert "Schema must be saved before describing" in str(e.value)


falexwolf (changes requested):
Can you please let it error if neither of `type`, `features`, nor `components` is provided? It doesn't make sense to create a feature without any of these and should throw a clear, tested error.

Can you please also document that doing `ln.Schema(name="TechSample", type=Sample).save()` makes a "folder" along the lines of making a `ULabel` or `Record` type?

sunnyosun:
> Can you please let it error if neither of `type`, `features`, nor `components` is provided? It doesn't make sense to create a feature without any of these and should throw a clear, tested error.
> 
> Can you please also document that doing `ln.Schema(name="TechSample", type=Sample).save()` makes a "folder" along the lines of making a `ULabel` or `Record` type?

But then why do we even allow `ln.Schema(name="TechSample", type=Sample).save()`? This is not a type/folder. Making folders were already previously implemented and tested, but one needs to pass `is_type=True` to be a folder.

falexwolf:
> But then why do we even allow ln.Schema(name="TechSample", type=Sample).save()? This is not a type/folder. Making folders were already previously implemented and tested, but one needs to pass is_type=True to be a folder.

Right. We should not allow it. Unless I'm missing something only `ln.Schema(name="TechSample", type=Sample, is_type=True).save()` makes sense and should be allowed.

So, we should throw a clear error upon `ln.Schema(name="TechSample", type=Sample).save()` and document creating schema types similar to how we document creating ulabel types.

falexwolf:
Can you finalize updating docs & error behavior and then feed this back to the users who originally raised it?

lamindb/models/schema.py
```diff
@@ -1211,12 +1211,14 @@ def get_type_str(dtype: str | None) -> str | None:
...
 
 
-def _get_related_name(self: Schema) -> str:
+def _get_related_name(self: Schema) -> str | None:
     related_models = dict_related_model_to_related_name(self, instance=self._state.db)
-    related_name = related_models.get(
-        parse_cat_dtype(self.itype, is_itype=True)["registry_str"]
-    )
-    return related_name
+    if self.itype:
```
falexwolf, line 1226: I think we should keep this modification, because `itype=None` should be allowed for composite schemas.

tests/core/test_schema.py
```diff
@@ -561,7 +561,11 @@ def test_schema_is_type():
     assert BioSample.hash is None
     assert BioSample.type == Sample
     assert BioSample.is_type
+    # create a schema without any features but with a type
+    TechSample = ln.Schema(name="TechSample", type=Sample).save()
```
falexwolf: Can you test that an error is raised here?

And then add a test that calls the same with `is_type=True`, which shouldn't error.

And then document both.

PR #3094, opened 2025-09-23
🐛 Fix inconsistency between `.get()` and `.filter()` due to `branch_id`

Koncopd:
Fixes https://github.com/laminlabs/pfizer-lamin-usage/issues/407

Includes https://github.com/laminlabs/lamin-cli/pull/154

### DIFF
--- a/lamindb/models/query_set.py
+++ b/lamindb/models/query_set.py
@@ -249,7 +249,7 @@ def get(
                 result = qs.filter(**expressions).order_by("-created_at").first()
                 if result is not None:
                     return result
-            raise registry.DoesNotExist from e
+            raise e
 
 
 class SQLRecordList(UserList, Generic[T]):
--- a/tests/core/test_queryset.py
+++ b/tests/core/test_queryset.py
@@ -212,6 +212,40 @@ def test_get_doesnotexist_error():
     )
 
 
+def test_get():
+    branch = ln.Branch(name="test_branch").save()
+
+    artifact = ln.Artifact.from_dataframe(
+        ln.User.to_dataframe(), key="df_test_get.parquet"
+    )
+    artifact.branch = branch
+    artifact.save()
+    # errors if doesn't find or multiple records found
+    ln.Artifact.get(key="df_test_get.parquet", branch=branch)
+    ln.Artifact.get(key="df_test_get.parquet", branch_id=branch.id)
+    ln.Artifact.get(key="df_test_get.parquet", branch__in=[branch])
+    ln.Artifact.get(key="df_test_get.parquet", branch_id__in=[branch.id])
+    ln.Artifact.get(key="df_test_get.parquet", branch=None)
+    ln.Artifact.get(key="df_test_get.parquet", branch_id=None)
+
+    with pytest.raises(ln.Artifact.DoesNotExist):
+        ln.Artifact.get(key="df_test_get.parquet")
+
+    ln.Artifact.get(artifact.id)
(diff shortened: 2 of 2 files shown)


falexwolf:
I'm glad a test is failing because there should be tests for the behavior of `.get()` vs. `filter()` in the presence of records in trash and on other branches.

I need to look but it strikes me as odd that we ignore `branch_id` for `get()` for filters that do not query by id.

I don't understand anymore (despite the comment), why I added this:

```
        # don't want branch_id here in .get(), only in .filter()
        expressions.pop("branch_id", None)
```

There must have been a case that required it and I very much hope there is a test. The comment should be expanded to explain the design choice.

If we can't identify the reason for why we would `.pop("branch_id")`, I think we should not pop it.

Note that queries by `id` or `uid` should be excluded from this rationale both in `.get()` and `.filter()`; indeed the trash shouldn't take effect if id/uid is provided.

Koncopd:
I have commented in the slack thread.

falexwolf:
Either way we should expand the tests for get() and filter() so that cases involving soft delete are properly covered; and then add the corresponding documentation to both methods.

Koncopd:
Ok, playing with this a bit, i start thinking you are right. My main problem is that it looks inconsistent that `get` by `id` or `uid` gives objects out of the trash also but doesn't give when using other expressions if we follow you proposal.

Koncopd:
>Note that queries by id or uid should be excluded from this rationale both in .get() and .filter(); indeed the trash shouldn't take effect if id/uid is provided.

This is done already.

Koncopd:
Ok, doing as you suggested. If you rather agree with my comment, we can just revert.

Koncopd:
Ok, so we have https://github.com/laminlabs/lamin-cli/blob/120e83c482b91b920b2891b61380f4ca5e3e540b/tests/core/test_save_annotate_files.py#L44 failing because 
the artifact is not the main branch, but in a custom one.
`artifact = ln.Artifact.get(key="mytest") # in branch "contrib1"`

Koncopd:
The test is ofc easy to fix just like:
`artifact = ln.Artifact.get(key="mytest", branch=branch) # in branch "contrib1"`
But i am not sure if we want to require specifying non-standard branches explicitly.

@falexwolf what do you think?

falexwolf:
I think the custom branch should be supplied here.

We do not want the queries across branches to interfere with each other.

The only exception is id/uid IMO because these are cross-branch entities. Everything else is not and might be duplicated or have branch specific semantics (depending on constraints)

falexwolf:
The failing test is a good example: a user has a new artifact version with the same key on another branch. Users working on the default or other branches should not query this artifact unless they explicitly request it.

Koncopd:
Ok, got it.

falexwolf:
Now I start to remember one use case that could explain that the trash state is ignored through popping the branch_id.

Say a user trashes a transform or run report or run environment. If `get(hash=...)` doesn't return that trashed transform or artifact, upon hash equality, one will have a duplicated artifact and run into an error for transform.

Now, I need to ponder on this a little bit; I think this says only something about hash constraints and lookups, which are indeed a bit similar to id/UID lookups; it's not an argument for popping the branch for a regular query.

I think we might want to consider adding hash queries to the exception.

Koncopd:
Another thing is `branch` and `hash`, because `hash` is actually checked in all branches i think. https://github.com/laminlabs/lamindb/blob/9c53446253caae3142f33bdc093323637af92cb6/lamindb/models/artifact.py#L334

PR #3095, opened 2025-09-23
🚸 Add the current branch by default for filtering

Koncopd:
This PR enables to query via `.filter()` and `.get()` on the current branch by adding the branch to the branches list on which the querying is done.

This means that instead of only the main branch queries are performed on the main and current branches if no branch is specified in the query expressions.


falexwolf:
This looks great!

But I'd say because of the way lamindb branching is built we should also include the default branch when filtering. So "current branch + default branch" instead of just "current branch".

Koncopd:
Hm, this logic is a bit different, but ok, I can change to that.

Koncopd:
Done.

Koncopd:
Ok, merging then.

PR #3102, opened 2025-09-26
✨ Allow virtual keys for artifacts registering paths in existing storages

Koncopd:


lamindb/models/artifact.py
```diff
@@ -2815,6 +2816,22 @@ def delete(
         """
         super().delete(permanent=permanent, storage=storage, using_key=using_key)
 
+    @property
+    def _real_key(self) -> str | None:
+        """A real key for artifacts that represent paths in existing storages and also have a virtual key."""
```
falexwolf: I'd make `_real_key` a public attribute called: `real_key`. Below is a tentative docstring. By having it public I think much transparency is created about the inner workings of lamindb for those interested.

An optional real storage key. 

It defaults to `f".lamindb/{uid}{suffix}"` if `_key_is_virtual = True` (`f".lamindb/{stem_uid}{suffix}"` if `overwrite_versions=True`). Otherwise it equals `key`.

If the user wants to deviate from this pattern, `_real_key` can be used to store the actual storage key.

Koncopd: But still as an aux field, right?

falexwolf: Right, I was going back and forth about this. Given we're doing the migrations anyway, we could add it as a an actual field in SQL.

I guess we could make the actual field private and nullable and then add the property on top. I think this is cleaner than the aux field, isn't it?

Koncopd: >I think this is cleaner than the aux field, isn't it?

I don't really see a point in a real field, if a property is on top anyways.

falexwolf: It's just much nicer to interact with, also slightly more efficient to store. The `_aux` field isn't particularly readable code, is it? You also have to think of the hub team who needs to work with this. They don't have the Python code but just the database.

Koncopd: Do you think it should be `real_key` or `_real_key`? I don't really see a point in making it private and adding a property, it is also good to have it public for queries.

falexwolf: How do you achieve this then?

> It defaults to f".lamindb/{uid}{suffix}" if _key_is_virtual = True (f".lamindb/{stem_uid}{suffix}" if overwrite_versions=True). Otherwise it equals key.

I think the only good solution is to have it be private and then the property public.

falexwolf: But maybe I'm missing something.

Koncopd: Now it is always written with the actual path.

Koncopd: i thought it would be very confusing otherwise.

falexwolf: I don't think we should duplicate the data. I think `_real_key` should be `None` in 95% of cases and only in those customized cases it's not `None`. I agree that the `uid`-based key is essentially immutable but it's still a lot of data duplication. 🤔

falexwolf: I mean we haven't had a `_real_key` for 3 years now and it wasn't confusing and `None`. So adding now the optional custom key seems to be like a natural pattern for having something that's not `None` in rare cases.

Koncopd: I am really against yet another complicated rule of checking whether something is `None`, then checking something else is `None`, then checking something else. This is why i chose to always populate `_real_key`. It is pretty simple to understand at least and it is `None` only for old artifacts.

In any case, your approach means closing this PR and starting a new one, because this one is already too far in the current direction. I can do that you don't agree with my reasoning. Probably i should have discussed my plan with you first before implementing.

falexwolf: I don't understand this entirely because you're saying that all existing artifacts would have `_real_key = None`. So to me it's more consistent and simpler if just all artifacts have `_real_key = None`.

If we really want to duplicate data then should write a data migration that populates this column for all legacy artifacts.

Koncopd:
@falexwolf ok, this is now implements `_real_key` the way you wanted. We need to decide on a better name and if we want to make it a real fileld.

Koncopd:
Maybe we should call it `_overwrite_key`?

falexwolf:
Why don't you like the name `_real_key`? Isn't this the most consistent with the name `_key_is_virtual` and the whole logic around it?

And yes, I think it should be an actual field in the database.

Koncopd:
>Why don't you like the name _real_key?

I think you didn't like it, no? I am actually fine with this name.

>And yes, I think it should be an actual field in the database.

Ok, will do.

Koncopd:
@falexwolf i have made it a field, but now i have to migrate `lamindata` and `lamin-dev` to run the tests. I think it is a backwards-compatible migration though.

falexwolf:
> but now i have to migrate lamindata and lamin-dev to run the tests.

Yes, please do so!

falexwolf:
Also `lamin-site-assets`.

falexwolf:
Can this be merged now?

Koncopd:
Not yet, I will finish some additional tests and merge.

falexwolf:
Great! Now I can continue with fixing all the Page-related migrations.

PR #3115, opened 2025-09-30
✨ Add PyTorch Lightning integration with a `Callback` class

Zethson:
Related to https://github.com/laminlabs/lamin-mlops/pull/30#issuecomment-3352181340

This PR bundles a few changes so that they can be more easily tested and used in https://github.com/laminlabs/lamin-mlops/pull/30.

## PyTorch Lightning callback

The callback automatically saves `Artifacts` for checkpoints and optionally annotates them with recorded `trainer` metrics.

It can be used like:

```python
lamindb_callback = ln.integrations.lightning.Callback(
    path=ckpt_dir / ckpt_filename, key=artifact_key, annotate_by=metrics_to_annotate
)

trainer = pl.Trainer(
    callbacks=[lamindb_callback]
)
```

Tested by proxy in those notebooks.

## ML tracking frameworks

Two new modules that provide utility functions to create all features that are useful when working with the respective ML tracking framework:

```python
ln.examples.mlflow.save_mlflow_features()
ln.examples.wandb.save_wandb_features()
```

## A `type` parameter for `ln.Feature.from_dict()`

This makes it easier to group several related features into a common Feature type:

```python
hyperparameter = ln.Feature(name="Autoencoder hyperparameter", is_type=True).save()
hyperparams = ln.Feature.from_dict({"batch_size": 32, "lr": 0.001}, type=hyperparameter)
```

### DIFF
--- a/lamindb/integrations/_lightning.py
+++ b/lamindb/integrations/_lightning.py
@@ -16,13 +16,13 @@ import lamindb as ln
 
 
 class Callback(pl.Callback):
-    """Saves PyTorch Lightning model checkpoints to LaminDB instance after each training epoch.
+    """Saves PyTorch Lightning model checkpoints to the LaminDB instance after each training epoch.
 
     Creates version families of artifacts for given `key` (relative file path).
 
     Args:
-        path: Path to the checkpoint.
-        key: Artifact key.
+        path: Path to the checkpoint
+        key: Artifact key
         features: Additional feature values that every checkpoint gets annotated by.
 
     Examples:
--- a/lamindb/models/feature.py
+++ b/lamindb/models/feature.py
@@ -847,11 +847,11 @@ class Feature(SQLRecord, CanCurate, TracksRun, TracksUpdates):
         """Create Feature records for dictionary keys.
 
         Args:
-            dictionary: Source dictionary to extract key information from.
-            field: FieldAttr for Feature model validation, defaults to `Feature.name`.
-            str_as_cat: Whether to interpret string values as categorical.
-            type: Feature type of all created features.
-            mute: Whether to mute dtype inference and feature creation warnings.
+            dictionary: Source dictionary to extract key information from
+            field: FieldAttr for Feature model validation, defaults to `Feature.name`
+            str_as_cat: Whether to interpret string values as categorical
+            type: Feature type of all created features
+            mute: Whether to mute dtype inference and feature creation warnings
         """
         from lamindb.models._feature_manager import infer_feature_type_convert_json
 


lamindb/integrations/__init__.py
```diff
@@ -6,9 +6,27 @@
...
+        return LightningCallback
+
+
 from lamindb.core.storage import save_tiledbsoma_experiment
 
 from ._croissant import curate_from_croissant
 from ._vitessce import save_vitessce_config
+
+__all__ = [
+    "LightningCallback",
```
falexwolf: Given this won't be imported by default and the `ln.integrations.do_something()` pattern won't work, can we structure this like a module import?

```
from lamindb.integrations import lightning as lnpl

lnpl.Callback
```

I appreciate that this is inconsistent to these function imports. But the inconsistency is anchored in the different user story: if an integration really consists in just one high-level convenience function (which is the best case I think), then it's nice to tab complete it via `ln.integrations.do_something()`.

But if the integration requires its own module and classes with potentially more complexity, then I think the module structure is more intuitive.

PS: This is the suggestion from here: https://github.com/laminlabs/lamin-mlops/pull/30#issuecomment-3352181340

Of course, happy to discuss other rationales that I might be missing!

Zethson: > then it's nice to tab complete it via ln.integrations.do_something().

There are ways to make the tab complete work (that I haven't necessarily added here yet).

It can be currently used like:

```python
lamindb_callback = ln.integrations.LightningCallback(mlflow_run_id, key_prefix="testmodels/mlflow")
```

which I find somewhat intuitive.

Nevertheless, you seem to have a strong opinion here and I don't so I'll change it.

falexwolf: > It can be currently used like:

Unless I'm missing something this won't work unless you have a dedicated import statement within `__init__.py`. 🤔 

If that doesn't exist one needs an explicit import statements:
```
from lamindb.integrations import LightningCallback
```

Or is `__all__` doing the trick? But I thought `__all__` doesn't affect the Python interpreter beyond the `*` import and is otherwise only guidance for editors.

Or is Python no longer behaving this way. I'm like 99% sure I ran into this often times, hence all the targeted re-exports for those part of the API that should be immediately available in a given namespace.

falexwolf: (Irrespective of whether that pattern works or not I find the grouping by a `lightning` module better for the reasons outlined above. ☺️)

Zethson: > Unless I'm missing something this won't work unless you have a dedicated import statement within __init__.py. 🤔

It does work. The `__getattr__` gave us what we needed. But anyways - already adapted. I'll finish this tomorrow.

falexwolf: I see! Well then this definitely changed over the years and we should likely be _much less aggressive_ about re-exports because that really affects import times. ☺️

falexwolf:
Given you say the `__getattr__` can resolve despite the lacking re-export, I even suggest this ☺️ 

```python
import lamindb as ln

lamindb_callback = ln.integrations.lightning.LightningCallback(
    path=ckpt_dir / ckpt_filename, key=artifact_key, annotate_by=metrics_to_annotate
)

trainer = pl.Trainer(
    callbacks=[lamindb_callback]
)
````

File: lamindb/integrations/_lightning.py [hunk 2]
```diff
@@ -0,0 +1,83 @@
...
+                if metric_name in trainer.callback_metrics:
+                    metric_value = trainer.callback_metrics[metric_name]
+                    feature_values[metric_name] = (
+                        metric_value.item()
+                        if hasattr(metric_value, "item")
+                        else float(metric_value)
+                    )
+
+        if feature_values:
+            af.features.add_values(feature_values)
```
falexwolf, line 82: Here (see https://github.com/laminlabs/lamin-mlops/pull/30 for context).

lamindb/core/_context.py
```diff
@@ -135,6 +135,9 @@ def flush(self):
         if not self.file.closed:
             self.file.flush()
 
+    def isatty(self) -> bool:
```
falexwolf, line 141: Can you add a docstring or comment for why this is needed, please?

lamindb/examples/__init__.py
```diff
@@ -7,8 +7,9 @@
    datasets
    cellxgene
    croissant
+   mlflow
 
 """
 
-from . import croissant, datasets, schemas
+from . import croissant, datasets, ml_tracking, schemas
```
falexwolf: For consistency with the `integrations` module and `croissant` and `cellxgene`, can we call this `mlflow` and `wandb` rather than introducing an ad-hoc grouping?

If you want to create a new namespace, then I'd say `mlops` for consistency with `lamin-mlops`. Danger is that this might change whereas `mlflow` and `wandb` won't change. ☺️

lamindb/examples/ml_tracking/__init__.py
```diff
@@ -0,0 +1,98 @@
...
+    - mlflow_experiment_id
+    - mlflow_experiment_name
+    - mlflow_user_id
+    - mlflow_status
+    - mlflow_lifecycle_stage
+    - mlflow_artifact_uri
+    - start_time
+    - end_time
+    """
+    mlflow_features = [
```
falexwolf: Can you group all 8 MLFlow features under an "MLFlow" type?

```
ln.Feature(name="MLFlow", is_type=True).save()
```

lamindb/examples/ml_tracking/__init__.py
```diff
@@ -0,0 +1,98 @@
...
+        ln.Feature(name="mlflow_run_id", dtype=str).save(),
+        ln.Feature(name="mlflow_run_name", dtype=str).save(),
+        ln.Feature(name="mlflow_experiment_id", dtype=str).save(),
+        ln.Feature(name="mlflow_experiment_name", dtype=str).save(),
+        ln.Feature(name="mlflow_user_id", dtype=str).save(),
+        ln.Feature(name="mlflow_status", dtype=str).save(),
+        ln.Feature(name="mlflow_lifecycle_stage", dtype=str).save(),
+        ln.Feature(name="mlflow_artifact_uri", dtype=str).save(),
+        ln.Feature(name="start_time", dtype=int).save(),
```
falexwolf: The `start_time` and `end_time` feature would benefit from a clearer name; or if one wants this general name that will likely interfere with user data, then I'd use `dtype` `datetime` rather than int.

lamindb/examples/ml_tracking/__init__.py
```diff
@@ -0,0 +1,98 @@
...
+        ln.Feature(name="end_time", dtype=int).save(),
+    ]
+
+    schema = ln.Schema(
```
falexwolf: Do we actually need a schema? Are we passing exactly those X features up there to a single call and want to validate them during that call? If not, let's not create a `Schema`.

lamindb/examples/ml_tracking/__init__.py
```diff
@@ -0,0 +1,98 @@
...
+        otype="DataFrame",
+        minimal_set=False,
+        maximal_set=True,
+        coerce_dtype=True,
+    ).save()
+
+    return schema
+
+
+def create_wandb_schema() -> Schema:
```
falexwolf: Same comments as for `MLFlow`.

lamindb/integrations/__init__.py
```diff
@@ -6,9 +6,33 @@
    save_vitessce_config
    save_tiledbsoma_experiment
    curate_from_croissant
+   lightning
 """
 
+from typing import Any
+
+
+def __getattr__(attr_name: str) -> Any:
```
falexwolf, line 15: I see now: we misunderstood each other. Python still behaves as it always did. Without a re-export of `lightning` a call  `ln.integrations.lightning` fails.

We can debate whether we want to introduce the "hack" of defining `__getattr__()`. 

My first reaction is skepticism: I've just spent months so that we'd be able to remove the `__getattr__()` hacks within the `__init__.py` of lamindb and schema module roots: https://github.com/laminlabs/lamindb/pull/2851 and all analogous PRs.

What you're doing here is building a lazy import mechanism only for the sake of enabling the pattern `ln.integrations.lightning.Callback()`. If the PEP that you linked this morning becomes real that we'd have a use case.

I see the benefit in UX and consistency. If there is no issue with scaling this pattern across 20 integrations and if there are no weird edge cases, I'm in favor of keeping it here. ☺️

falexwolf: A comment would be very helpful, though.

lamindb/integrations/__init__.py
```diff
@@ -6,9 +6,33 @@
...
+    raise AttributeError(f"module has no attribute {attr_name!r}")
+
+
 from lamindb.core.storage import save_tiledbsoma_experiment
 
 from ._croissant import curate_from_croissant
 from ._vitessce import save_vitessce_config
+
+
+def __dir__():
```
falexwolf, line 30: Why this?

falexwolf: Do you need this to get `__getattr__()` to work? It'd be nice to not have to do this.

lamindb/integrations/_lightning.py
```diff
@@ -0,0 +1,77 @@
...
+from typing import Any
+
+import lightning as pl
+from lightning.pytorch import LightningModule, Trainer
+
+import lamindb as ln
+
+
+class Callback(pl.Callback):
+    """Saves PyTorch Lightning model checkpoints to LaminDB after each training epoch.
```
falexwolf: "Or when aborting."

lamindb/integrations/_lightning.py
```diff
@@ -0,0 +1,77 @@
...
+
+    Creates version families of artifacts for given `key` (relative file path).
+
+    Args:
+        path: Path to the checkpoint
+        key: Artifact key in LaminDB storage
```
falexwolf: Because this can be confusing in several ways, I'd remove "in LaminDB storage".

There should be periods at each of the sentence.

lamindb/integrations/_lightning.py
```diff
@@ -0,0 +1,77 @@
...
+        features: Additional feature values that every checkpoint gets annotated by.
+            Examples are { "mlflow_run_id": mlflow_run.info.run_id }.
```
falexwolf: Better to have an example under `Examples:`; then code styling is also easier.

lamindb/integrations/_lightning.py
```diff
@@ -0,0 +1,77 @@
...
+        path: str | Path,
+        key: str,
+        features: dict[str, Any] | None = None,
+    ):
+        self.path = Path(path)
+        self.key = key
+        self.features = features or {}
+
+    def on_train_start(self, trainer: Trainer, pl_module: LightningModule) -> None:
+        """Validates that LaminDB Features exist for all specified params."""
```
falexwolf: "Features" should be lower-cased for consistency with rest of docs. ☺️

lamindb/models/feature.py
```diff
@@ -838,6 +838,7 @@ def from_dict(
         field: FieldAttr | None = None,
         *,
         str_as_cat: bool | None = None,
+        feature_type: str | SQLRecord | None = None,
```
falexwolf: Please name it `type` rather than `feature_type` for consistency with `from_values()` and the feature constructor itself.

Zethson: I explicitly avoided that because it's a keyword in Python but yeah consistency is key.

lamindb/models/feature.py
```diff
@@ -904,6 +906,11 @@ def from_dict(
         try:
             features = [Feature(name=key, dtype=dtype) for key, dtype in dtypes.items()]  # type: ignore
             assert len(features) == len(dictionary)  # noqa: S101
+            if feature_type:
+                if isinstance(feature_type, str):
+                    feature_type = Feature(name=feature_type, is_type=True).save()  # type: ignore
```
falexwolf: The decision to create the type within the `from_dict()` constructor deserves attention.

For the standard constructor we don't create the type. If you do `Feature(name="...", dtype="...", type=type).save()` and `type` doesn't exist, then you'll get an error.

Similarly if you do `Gene.from_values(["..."], organism=organism).save()`; the `organism` has to exist, it won't be created for you via `from_values()`.

I'd say it's more consistent & hence predictable if the `type` creation is not dealt with by `from_dict()`. It's additional user input.

lamindb/integrations/_lightning.py
Zethson: Well that's only mildly correct. If you don't let the first epoch finish and abort this early, nothing will be saved (there is nothing to save).

I understand why you're suggesting that but I think that we we have is clear enough.

lamindb/models/feature.py
falexwolf: Your code would also become simpler and you could leverage the consistent signature for the `type` argument across `from_dict()` and `Feature()` by simply passing `type` into `Feature()` without additional validation: `Feature()` takes care of it.

```python
type: SQLRecord | None = None
features = [Feature(name=key, dtype=dtype, type=type) for key, dtype in dtypes.items()]
```

tests/core/test_feature.py
```diff
@@ -214,3 +214,26 @@ def test_feature_from_dict(dict_data):
...
     )
     assert len(features_with_field) == len(dict_data)
+
+
+def test_feature_from_dict_type(dict_data):
+    # feature_type str
+    features = ln.Feature.from_dict(
+        dict_data, str_as_cat=True, feature_type="Testdata_str"
+    )
+    ln.save(features)
```
falexwolf: This can be simplified: Because you `return SQLRecordList(features)` you can do `features.save()`.

Hence, this becomes one line:
```python
features = ln.Feature.from_dict(
    dict_data, str_as_cat=True, feature_type="Testdata_str"
).save()
```

lamindb/models/feature.py
falexwolf: That decision was made 9 months ago: https://laminlabs.slack.com/archives/C046F63LJJ2/p1736764042941039?thread_ts=1736406964.410579&cid=C046F63LJJ2

I get it & agree but alternatives `folder` or `concept` are both more problematic.

The wandb `Artifact` also uses `type`.

So, I still wouldn't know a better word.

lamindb/examples/ml_tracking/__init__.py
Zethson: No we don't need a Schema. This is purely for consistency with CxG and because you had requested a Schema. I'll rewrite this.

lamindb/integrations/__init__.py
Zethson: Overwriting `get_attr` is widely used at scverse where Phil introduced us to it as a simple lazy import solution. I find it somewhat intuitive now.

I don't see any drawbacks so I'll document and keep it.

lamindb/integrations/__init__.py
Zethson: The `__dir__` override is needed because we're using `__getattr__` for lazy imports that aren't defined at the module level. Without `__dir__`, autocomplete sees only what's in the module's namespace at import time.

I can add another comment explaining this.

lamindb/examples/ml_tracking/__init__.py
falexwolf: Ok! Right, I see how this is was confusing:

> What gets really scary now is the "define_features_and_labels()` part.
> To this end we could add ln.integrations.create_mlflow_schema() similar to what we have here: https://docs.lamin.ai/lamindb.examples.schemas
> And similar to import_source(). Just something that makes all these MLFlow-related types and features with one line.

But yeah, I meant "just something that makes these MLFlow-related types and features"; not a literal `Schema` object.

The language comes from drawing the analogy with databases and warehouses where "schema" is "all the stuff you need to define before you can actually add data to the system". In that analogy, defining features & labels is "part of defining the schema".

But of course it's confusing because we have a literal `Schema` object that's inspired by pydantic/pyarrow/arrays.

lamindb/examples/mlflow/__init__.py
```diff
@@ -0,0 +1,38 @@
...
+.. autosummary::
+   :toctree: .
+
+   save_mlflow_features
+"""
+
+import lamindb as ln
+
+
+def save_mlflow_features():
```
Zethson, line 12: I know this this one function and the wandb one are already in modules that carry the framework name but I still think that being explicit here makes it more readable.

falexwolf (approved):
Looks great! Thank you!

tests/core/test_feature.py
```diff
@@ -214,3 +214,15 @@ def test_feature_from_dict(dict_data):
         dict_data, field=ln.Feature.name, str_as_cat=False
     )
     assert len(features_with_field) == len(dict_data)
+
+
+def test_feature_from_dict_type(dict_data):
+    feature_type = ln.Feature(name="Testdata_feature_type", is_type=True).save()
+    features = ln.Feature.from_dict(
+        dict_data, str_as_cat=True, type=feature_type
```
falexwolf, line 222: Do we have `str_as_cat` anywhere else in the API?

falexwolf: This is the one thing that still bothers me. Also because it makes `ULabel` which we'll deprecate at some point.

falexwolf: Is it maybe the safer path to just not add `str_as_cat` and default to `str` for `str`-typed values? 🤔

Zethson: https://github.com/laminlabs/lamindb/issues/3132 follow up here.

PR #3117, opened 2025-09-30
🔊 Raise an error in `AnnDataCurator` when `schema.otype != 'AnnData'`, no exceptions for `otype is None`

sunnyosun:


sunnyosun:
@falexwolf, shall we make the `otype` a required field in the PR?

falexwolf:
I suggest something like this in the constructor:

```
if self.otype is None:
     assert not components, "you have to pass `otype` if you pass `components`"
```

For schema types and schemas that can be used for dictionaries and dataframes, it's kinda better to still allow `None`.

But for everything else we should hard error. I think the condition up there captures it, does it?

sunnyosun:
> I suggest something like this in the constructor:
> 
> ```
> if self.otype is None:
>      assert not components, "you have to pass `otype` if you pass `components`"
> ```
> 
> For schema types and schemas that can be used for dictionaries and dataframes, it's kinda better to still allow `None`.
> 
> But for everything else we should hard error. I think the condition up there captures it, does it?

This is already the case: https://github.com/laminlabs/lamindb/blob/7b54cced395883a09c38a80587d9aa1f1ecb68f8/lamindb/models/schema.py#L648-L651

falexwolf:
> This is already the case:

But how was the user then able to run into the bug? Why didn't this prevent the issue we saw yesterday? `otype` should have been set for an `AnnData` schema.

sunnyosun:
> > This is already the case:
> 
> But how was the user then able to run into the bug? Why didn't this prevent the issue we saw yesterday? `otype` should have been set for an `AnnData` schema.

The user passed a DataFrame schema without setting `otype`, the change in this PR should fix the issue.

falexwolf:
Ok! Do we need the same for spatialdata et al? Can you briefly check?

sunnyosun:
Already checked, all other curators are proper!

PR #3110, opened 2025-09-28
🗃️ Add a `page` and an `is_locked` field to `SQLRecord`, add a `Page` registry 

falexwolf:
Allows adding versioned markdown pages to every `SQLRecord` and locking.

- https://github.com/laminlabs/bionty/pull/322
- https://github.com/laminlabs/wetlab/pull/121

PR #3120, opened 2025-10-01
🏗️ Re-design page schema to block schema

falexwolf:
Re-architect the schema around the `Page` entity that was introduced a few days ago:

- https://github.com/laminlabs/lamindb/pull/3110

Now, we're leveraging a `Block` model that, in the first use case, will serve markdown pages.

Comes with:

- https://github.com/laminlabs/bionty/pull/323
- https://github.com/laminlabs/wetlab/pull/123


falexwolf:
Are these permission failures because we don't cover these new tables appropriately with RLS or except them?

image

Or is some other update necessary?

lamindb/models/block.py
```diff
@@ -0,0 +1,174 @@
...
+    content: str = TextField()
+    """Content of the block."""
+    hash: str = CharField(max_length=22, db_index=True, null=True)
+    """Content hash of the block."""
+    type: str = CharField(
+        max_length=22, db_index=True, default="mdpage", db_default="mdpage"
+    )
+    """Type of the block.
+
+    Only current option is: "mdpage" (markdown page).
```
falexwolf, line 39: In the future, we likely also want "comment" or "post".

falexwolf: Or could actually be a "block" that's part of a page rather than the full page.

falexwolf:
Oh sorry, @Koncopd, forgot to ping you.

lamindb/models/block.py
```diff
@@ -0,0 +1,174 @@
...
+    """
+    vertical_pos: int = models.SmallIntegerField(default=0, db_default=0, db_index=True)
+    """The vertical position of the block in the GUI as as an integer that can be converted into a percentage.
+
+    Say, one could map 10k to 100% and 0 to 0%. Using an integer allows exact positioning (e.g. to group comments on the same spot) vs. a float would have ambiguities.
```
falexwolf: I was worried about a float and so I made this an integer. But after talking to AI I think a 0.25 is very likely not ambiguous, so I'd rather go back to float and then also add `vertical_pos` to `Record`.

Koncopd:
>Are these permission failures because we don't cover these new tables appropriately with RLS or except them?

Yes, it looks so, there is an error during the setup:
```
psycopg2.errors.UndefinedColumn: column r.space_id does not exist
LINE 2429:                         AND r.space_id IN (SELECT id FROM sp...
```
I think the new tables cause this.

lamindb/models/block.py
```diff
@@ -0,0 +1,174 @@
...
+        editable=False, db_default=models.functions.Now(), db_index=True
+    )
+    """Time of creation of record."""
+    created_by: User = ForeignKey(
+        "User", CASCADE, default=None, related_name="+", null=True
+    )
+    """Creator of block."""
+
+
+class RootBlock(BlockMixin, BaseSQLRecord):
```
falexwolf, line 56: This content can appear at the root of a registry.

lamindb/models/block.py
chaichontat: I think we already discussed this and we came to a conclusion that the float would be better mainly for performance reasons.

PR #3124, opened 2025-10-02
🐛 Fix re-creation of soft-deleted records

sunnyosun:
- CanValidate functions didn't take into account branches; this PR fixes it by only validating records that are not in hidden or trashed branches.
- When user tries to add a new record whose uid or ontology_id matches an existing record in the trash, we now rescue this record from trash to avoid unique constraint errors in the database.

Before | After
--- | ---
sqlite Screenshot 2025-10-02 at 15 59 23 | sqlite Screenshot 2025-10-02 at 15 58 38
postgres Screenshot 2025-10-06 at 12 14 04 | postgres Screenshot 2025-10-06 at 12 14 20

### DIFF
--- a/lamindb/core/_mapped_collection.py
+++ b/lamindb/core/_mapped_collection.py
@@ -634,8 +634,8 @@ class MappedCollection:
     def __exit__(self, exc_type, exc_val, exc_tb):
         self.close()
 
-    @staticmethod
-    def torch_worker_init_fn(worker_id):
+    @classmethod
+    def torch_worker_init_fn(cls, worker_id):
         """`worker_init_fn` for `torch.utils.data.DataLoader`.
 
         Improves performance for `num_workers > 1`.
--- a/lamindb/models/has_parents.py
+++ b/lamindb/models/has_parents.py
@@ -42,13 +42,22 @@ def _query_relatives(
     attr: str,
     cls: type[HasParents],
 ) -> QuerySet:
-    relatives = cls.objects.none()  # type: ignore
-    if len(records) == 0:
+    from .query_set import get_default_branch_ids
+
+    branch_ids = get_default_branch_ids()
+
+    def query_relatives_on_branches(records, attr, cls) -> QuerySet:
+        relatives = cls.objects.none()  # type: ignore
+        if len(records) == 0:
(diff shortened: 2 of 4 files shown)


lamindb/models/sqlrecord.py
```diff
@@ -913,13 +912,54 @@ def save(self: T, *args, **kwargs) -> T:
...
                     pre_existing_record = self.__class__.get(root=self.root)
                     init_self_from_db(self, pre_existing_record)
+                elif (
+                    isinstance(e, IntegrityError)
+                    and ("ontology_id" in error_msg or "uid" in error_msg)
+                    and (
+                        "UNIQUE constraint failed" in error_msg
+                        or "duplicate key value violates unique constraint" in error_msg
+                    )
+                ):
```
falexwolf, line 924: Can you move all of this into its own function with a docstring explaining what it does?

Otherwise this is too much cryptic logic.

Does the test cover all this logic?
```python
def test_recreate_soft_deleted_record():
    record = bt.Ethnicity.from_source(ontology_id="HANCESTRO:0006").save()
    assert record.branch_id == 1
    record.delete()
    assert record.branch_id == -1
    record = bt.Ethnicity.from_source(ontology_id="HANCESTRO:0006")
    record.description = "new description"
    record.save()
    # now this record is recovered from the trash with the new description
    assert record.branch_id == 1
    assert record.description == "new description"
    bt.Ethnicity.objects.filter().delete()def test_recreate_soft_deleted_record():
    record = bt.Ethnicity.from_source(ontology_id="HANCESTRO:0006").save()
    assert record.branch_id == 1
    record.delete()
    assert record.branch_id == -1
    record = bt.Ethnicity.from_source(ontology_id="HANCESTRO:0006")
    record.description = "new description"
    record.save()
    # now this record is recovered from the trash with the new description
    assert record.branch_id == 1
    assert record.description == "new description"
    bt.Ethnicity.objects.filter().delete()
```

I'm not saying we need a unit test for this function but we definitely need a comment in the test and the function linking them together. Otherwise it'll be hopeless to maintain all of this.

Already now I half-forgot the rationales behind "what is recovered" and "what is not recovered".

PR #3127, opened 2025-10-04
♻️ Expand `Record` so that it can act as an ontology

falexwolf:
`Record` gains `ULabel`'s additional fields: `parents`, `children`, `reference`, `reference_type`. This is so that `Record` can also model full hierarchical ontologies.

This also means that, in the future, `Record` can replace `ULabel` and we don't need to bother users with learning both concepts (most users will likely never go as far as to engage with the ontology features of Record; if they need to learn the ontology concept they'll have a reason).

Many of the code examples that centered on `ULabel` now center on `Record` even though `ULabel` just continues to work as before.

# Docs

Before | After
--- | ---
image | image
image | image
/ | image
/ | image

# Other

**Squashed migrations**. This PR also squashes migrations to accelerate the instance creation time.


tests/curators/test_cxg_curator.py
```diff
@@ -31,6 +31,7 @@ def create_schema(version: CELLxGENESchemaVersions, **kwargs):
         entity.filter().delete(permanent=True)
 
 
+@pytest.mark.filterwarnings("ignore:SyntaxWarning")
```
falexwolf, line 34: Trying this.

PR #3129, opened 2025-10-06
🐛 Extend `query_relatives()` to respect branches

falexwolf:
`query_relatives()` did neither respect the trash & archive states nor a branch that a user checked out.

This led to this bug:

- https://github.com/laminlabs/pfizer-lamin-usage/issues/390

Is a follow-up to:

- https://github.com/laminlabs/lamindb/pull/3095/

### DIFF
--- a/lamindb/models/query_set.py
+++ b/lamindb/models/query_set.py
@@ -61,11 +61,22 @@ def get_keys_from_df(data: list, registry: SQLRecord) -> list[str]:
 
 
 def get_default_branch_ids() -> list[int]:
+    """Return branch IDs to include in default queries.
+
+    By default, queries include records on the main branch (branch_id=1) but exclude trashed (branch_id=-1)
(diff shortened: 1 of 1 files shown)


lamindb/models/query_set.py
```diff
@@ -60,6 +60,14 @@ def get_keys_from_df(data: list, registry: SQLRecord) -> list[str]:
     return keys
 
 
+def get_default_branch_ids() -> list[int]:
+    branch_id = setup_settings.branch.id
+    branches = [0, 1]
```
Koncopd: why 0?

falexwolf: 0 is archive.

The question is whether we want the archive to be included in default queries.

Pondering on how other tools do it (e.g. Slack, GitHub), I think they indeed exclude archived records.

So, we should probably do the same and not include `0`. 🤔

falexwolf: Fixed and commented on this.

Issue #3132, opened 2025-10-06
Consider deprecating `str_as_cat` for `Feature.from_dict()`


Zethson:
### Add a description

https://github.com/laminlabs/lamindb/pull/3115#discussion_r2407153715

it should go through a few deprecating cycles in any case.

PR #3134, opened 2025-10-07
🎨 Deprecate `str_as_cat` in `Feature.from_dict()`

Zethson:
Fixes https://github.com/laminlabs/lamindb/issues/3132

There's different options to handle this deprecation. A few of these:

1. Option 1 (error on ambiguity): Keep raising ValueError for ambiguous types but remove `str_as_cat` from error message, forcing users to create features explicitly before calling from_dict().
2. Option 2 (default with warning): Default ambiguous strings to str and warn every time, guiding users toward explicit feature creation.
3.  Option 3 (silent default, chosen): Default ambiguous strings to str silently, only warning when deprecated `str_as_cat` parameter is used, minimizing noise while maintaining backward compatibility during deprecation period.

I went for option 3.


falexwolf:
Looks great! I'd merge and release, OK?

PR #3136, opened 2025-10-07
🐛 Fix import of `lightning` integration

Zethson:
Otherwise we run into RecursionErrors

https://laminlabs.slack.com/archives/C05FDBBFJ1F/p1759837682744229


falexwolf:
Ah, this really is pretty hacky if one can't even make a public module public. There might be more side effects with `__getattr__()` etc.

I mean I guess we keep these hacks for now. But it's not pretty, is it?

Zethson:
Let's get this in and then I'll see whether there's a nicer way. I don't consider this a hack btw.

I'll follow up

falexwolf:
Oh man, this was even the reason for why the docs were killed.

lamindb/models/feature.py
```diff
@@ -850,7 +850,7 @@ def from_dict(
         Args:
             dictionary: Source dictionary to extract key information from
             field: FieldAttr for Feature model validation, defaults to `Feature.name`
-            str_as_cat: Deprecated. Will be removed in LaminDB 1.11.6.
+            str_as_cat: Deprecated. Will be removed in LaminDB 1.12.3.
```
falexwolf: We'll remove it in LaminDB 2.

falexwolf: We don't remove stuff on patch releases. ☺️

falexwolf: (Well of course unless it's an accident or we can't avoid it)

falexwolf:
But yeah, we need a working release.

falexwolf:
Merge?

PR #3139, opened 2025-10-08
🐛 Require an existing restricted storage location when writing to a r…

falexwolf:
…estricted space

### DIFF
--- a/lamindb/errors.py
+++ b/lamindb/errors.py
@@ -52,6 +52,12 @@ class UnknownStorageLocation(Exception):
     pass
 
 
+class NoStorageLocationForSpace(Exception):
+    """No storage location found for space."""
+
+    pass
+
+
 # equivalent to Django's DoesNotExist
 # and SQLAlchemy's NoResultFound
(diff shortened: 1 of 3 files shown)


Koncopd:
Hm, i realize this is problematic:
```python
artifact = ln.Artifact(...)
artifact.space = space
artifact.save()
```
right?

falexwolf:
Let's call. We have to ensure that `storage.space` and `artifact.space` stay in sync. 🤔 -- I believe we had a thread on this somewhere already.

Koncopd:
There should be a check on `save` then probably.

falexwolf:
This is the common integrity problem when fields are mutated that shouldn't be. We're already monitoring `key` and `name`. We can do that for space, too.

But it really is a bit more complicated because moving an artifact from one space into another now requires moving it in storage and that's something we still need to work on.

Can I merge this PR either way?

Koncopd (approved):

Koncopd:
Yes, looks good to me.

PR #3144, opened 2025-10-09
⚡️ Speed up `Curator.add_new_from()`

sunnyosun:


sunnyosun:
@falexwolf it should be safe here, right?

falexwolf:
Yes, should be fine!

PR #3153, opened 2025-10-12
🐛 Sanitize dataframe proxy for `dict` validation to enable schema-based validation of categoricals in dicts

falexwolf:
Because dictionaries don't have dtype `Categorical`, the schema-based validation through a proxy `DataFrame` will fail unless the `dtype` in the columns is adjusted to be categorical.

This PR fixes the associated bug and introduces a test for it.

- https://github.com/laminlabs/pfizer-lamin-usage/issues/441

Also: Rename 'linked features' to 'external features' for consistency with the special keyword `__external__`, consistent with hub:

- https://github.com/laminlabs/laminhub/pull/3533

The curator test suite is very long running; mostly because of the CxG tests. This PR removes the legacy curator tests and speeds up instance creation by removing the `wetlab` schema module. It also suppresses 100s/1000s of pronto warnings.

### DIFF
--- a/lamindb/models/_feature_manager.py
+++ b/lamindb/models/_feature_manager.py
@@ -470,7 +470,7 @@ def describe_features(
         for child in int_features_tree_children:
             dataset_tree.add(child)
 
-    # Linked features
+    # External features
     ext_features_tree_children = []
     if external_data:
         ext_features_tree_children.append(
@@ -481,7 +481,7 @@ def describe_features(
             )
         )
     # ext_features_tree = None
-    ext_features_header = Text("Linked features", style="bold dark_orange")
+    ext_features_header = Text("External features", style="bold dark_orange")
     if ext_features_tree_children:
         ext_features_tree = tree.add(ext_features_header)
         for child in ext_features_tree_children:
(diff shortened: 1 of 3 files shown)


falexwolf:
This first commit accounts for the fact that dictionaries do not have a `Categorical` dtype. Hence, we need to sanitize the proxy dataframe.

I notice that this pattern of creating a `temp_df` is sprinkled into several places of the code base. And in this first commit I might have forgotten to update some of them.

This danger means we have to de-duplicate the "proxy dataframe" logic. I'll do this in the next commit.

falexwolf:
Also found the pattern here:

image

Some of these other places are also dangerous:

image

falexwolf:
It seems like this will be a bigger refactor. So, let's first delete the test suite of the legacy curators; similar to how we deleted the legacy TiledbsomaCurator tests 2 weeks ago:

- https://github.com/laminlabs/lamindb/pull/3135

We can omit `curators/_legacy.py` in the coverage compute.

falexwolf:
We'll also suppress the 100s/1000s of warnings from the cxg curator tests.

falexwolf:
Now we have a `DictCurator` and I added it in all places where previously there was the pattern 

```python
tmp_df = pd.DataFrame([dictionary])
DataFrameCurator(tmp_df, schema)
```

Now this reads:

```python
DictCurator(dictionary, schema)
```

falexwolf:
FYI @Zethson, this is a partial refactor of the schema-based validation of external features. Please take a look when you have time!

falexwolf:
I found two more rather significant bugs and fixed them here:

- https://github.com/laminlabs/lamindb/pull/3155

PR #3165, opened 2025-10-13
🐛 Set db tokens for connections in all contexts (threads, tasks), disable task-specific database wrappers in jupyter for `ipykernel>=7.0.0`

Koncopd:
Related https://github.com/laminlabs/lamindb-setup/pull/1166

`ipykernel==7.0.0` makes `jupyter` execute every cell with a new task. And in `django` database wrappers (`django.db.connections`) are context specific, meaning they are recreated for every thread or task. This means that with the new `ipykernel`  each cell has its own `django` database wrapper. This is problematic because it is just slow, and also breaks setting db tokens for fine grained permissions.

The PR in `lamindb-setup`:

- It disables task-specific dtabase wrapper switching in `jupyter` with `ipykernel>=7.0.0`, meaning they are preserved between tasks (especially between cells). Thread-specificity is preserved.
- On establishing a new database connection, it is now checked that the db token is set and then adds it if it isn't. This is important for multi-threading where we didn't have this check.


Koncopd:
todo: make the execute wrapper for db token through signals on connection created

PR #3174, opened 2025-10-15
✨ Enable curating features with nested subtypes

sunnyosun:
This is a follow-up PR to:

- https://github.com/laminlabs/lamindb/pull/3169

This PR implements the proper curation for features with nested subtypes. It ensures the type record is fetched based on the specified parent types in `feature.dtype`, and raises an error when there's ambiguity.

```python
# first we create LabA->Sample
LabA = ln.Record(name="LabA", is_type=True).save()
Sample = ln.Record(name="Sample", is_type=True, type=LabA).save()
for name in ["sample1", "sample2", "sample3", "sample4"]:
    ln.Record(name=name, type=Sample).save()

# then we create LabB->Sample
LabB = ln.Record(name="LabB", is_type=True).save()
Sample = ln.Record(name="Sample", is_type=True, type=LabB).save()
for name in ["sample5", "sample6"]:
    ln.Record(name=name, type=Sample).save()

# "sample5" is not part of LabA, so it should not be validated
with pytest.raises(ValidationError):
    curator = ln.curators.DataFrameCurator(nested_cat_df, nested_cat_schema)
    curator.validate()
```

Also refactor `parse_dtype` to remove the redundancy between `subtype_str` and `nested_subtypes`:
- `subtype_str` → `filter_str`
- `nested_subtypes` → `subtypes_list`

### DIFF
--- a/tests/curators/test_curators_examples.py
+++ b/tests/curators/test_curators_examples.py
@@ -247,8 +247,6 @@ def test_dataframe_curator(mini_immuno_schema: ln.Schema):
 
     assert artifact.schema == mini_immuno_schema
     assert artifact.features.slots["columns"].n == 5
-    print("describe output:")
-    print(artifact.features.describe(return_str=True))
     assert (
         artifact.features.describe(return_str=True)
         == """\


tests/curators/test_curators_examples.py
```diff
@@ -247,6 +247,8 @@ def test_dataframe_curator(mini_immuno_schema: ln.Schema):
 
     assert artifact.schema == mini_immuno_schema
     assert artifact.features.slots["columns"].n == 5
+    print("describe output:")
```
falexwolf: debug stmt

falexwolf (approved):
This looks great except for the PR description.

We should drop this non-sensical USCostumer example. Let's use Lab and Sample and carefully think through these examples.

sunnyosun:
Updated PR description!

PR #3177, opened 2025-10-16
✨ Add `Transform.from_git()`

falexwolf:
See new docs:

left | right
--- | ---
image | image

Addresses the `lamindb` part of:

- https://github.com/laminlabs/nf-lamin/issues/76

### DIFF
--- a/lamindb/models/transform.py
+++ b/lamindb/models/transform.py
@@ -68,10 +68,12 @@ class Transform(SQLRecord, IsVersioned):
 
         If you switch on
         :attr:`~lamindb.core.Settings.sync_git_repo` a script-like transform is
-        synched to its hashed state in a git repository upon calling `ln.track()`.
+        synched to its hashed state in a git repository upon calling `ln.track()`::
 
-        >>> ln.settings.sync_git_repo = "https://github.com/laminlabs/lamindb"
-        >>> ln.track()
+            ln.settings.sync_git_repo = "https://github.com/laminlabs/lamindb"
+            ln.track()
+
+        Alternatively, you create transforms that map pipelines via `Transform.from_git()`.
 
     The definition of transforms and runs is consistent the OpenLineage
     specification where a :class:`~lamindb.Transform` record would be called a
@@ -95,7 +97,6 @@ class Transform(SQLRecord, IsVersioned):
 
     Notes:
         - :doc:`docs:track`
-        - :doc:`docs:data-flow`
         - :doc:`docs:redun`
         - :doc:`docs:nextflow`
         - :doc:`docs:snakemake`
@@ -387,7 +388,7 @@ class Transform(SQLRecord, IsVersioned):
                 ).save()
                 assert transform.version == "v2.0.0"
 
-            Create a _sliding_ transform from a Nextflow repo's `dev` branch.
+            Create a *sliding transform* from a Nextflow repo's `dev` branch.
             Unlike a regular transform, a sliding transform doesn't pin a specific source code state,
             but adapts to whatever the referenced state on the branch is::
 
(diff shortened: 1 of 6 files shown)


lamindb/models/transform.py
```diff
@@ -328,13 +336,87 @@ def __init__(
...
+            source_code += f"\nentrypoint: {entrypoint}"
+        if branch is not None and version == branch:
+            # sliding transform, no defined source code state
+            source_code += f"\nbranch: {branch}"
+            reference = None
+            reference_type = None
+        else:
+            # regular transform, defined source code state
+            source_code += f"\ncommit: {commit_hash}"
+            reference = f"{url}/blob/{commit_hash}/{path}"
```
rcannood: This will only work for GitHub/Gitea/Gogs repositories, not for GitLab, BitBucket or Azure DevOps.

lamindb/models/transform.py
```diff
@@ -328,13 +336,87 @@ def __init__(
...
         return self.key.split("/")[-1]
 
+    @classmethod
+    def from_git(
+        cls,
+        url: str,
+        path: str,
+        key: str | None = None,
+        version: str | None = None,
+        entrypoint: str | None = None,
```
rcannood, line 360: How do the path and entrypoint get stored? Do they get stored in `_aux`?

Also, this is Nextflow specific but how do we want to handle the schema files? Every [url, path, entrypoint] should have a separate schema file which we can provide.

lamindb/models/transform.py
```diff
@@ -77,7 +78,6 @@ class Transform(SQLRecord, IsVersioned):
     "job" and a :class:`~lamindb.Run` record a "run".
 
     Args:
-        name: `str` A name or title.
         key: `str | None = None` A short name or path-like semantic key.
         type: `TransformType | None = "pipeline"` See :class:`~lamindb.base.types.TransformType`.
         revises: `Transform | None = None` An old version of the transform.
```
rcannood, line 90: Add docs for version, description, reference, reference_type, source_code?

lamindb/models/transform.py
```diff
@@ -196,9 +196,13 @@ class Meta(SQLRecord.Meta, IsVersioned.Meta):
     @overload
     def __init__(
```
rcannood, line 203: In `from_git` there are arguments for the Nextflow's `mainScript` setting (i.e. `path`) and the `entrypoint`. Should they also be added as arguments here?

lamindb/models/transform.py
falexwolf: Yes! Of course!

lamindb/models/transform.py
falexwolf: They get stored within the source code, but we could consider duplicating that information in `_aux`. I'd prefer not to duplicate, however.

lamindb/models/transform.py
falexwolf: Yes. These other ones are out of scope right now. I'm just not quite sure how to error because github can be on custom domains.

lamindb/models/transform.py
sunnyosun: What is the entrypoint here? Is it the same as `entryName`? From the nextflow pipelines config, there are `mainScript`, `entryName` and `schemaName`.

falexwolf:
Robrecht & I caught up on a call. Conclusion is that overall the design works. I'll fix all these evident issues and expand the tests to failure modes, then merge and release within about one hour.

lamindb/models/transform.py
falexwolf: I see. I didn't read the comment about `schemaName`, unfortunately. I'm not sure whether this should be handled on the transform level. It more seems like a required parameter. 🤔

Would Julianus need to pass this in some form? How do we need to expand examples to cover for this case?

PR #3180, opened 2025-10-16
✨ Allow specifying branch in `ln.view()`

Zethson:
Add (optional) branch filtering to `ln.view()`.

- Default to current branch (`settings.branch.id`) when branch not specified

```python
ln.view(branch=ln.Branch.get(name="main"))
```


falexwolf:
This should already work out of the box meanwhile via pre-configured branches.

PR #3187, opened 2025-10-17
🐛 Fix a bug that prevented re-curating existing artifacts in some cases

falexwolf:


falexwolf:
FYI: @Zethson @sunnyosun

I ran into this surprising failure working on a curation example this morning.

```python
    def test_curator_remote():
        lamindata_artifacts = ln.Artifact.using("laminlabs/lamindata")
>       curator = ln.curators.DataFrameCurator(
            lamindata_artifacts.get("F6PrJLEnWeHqUHrK0000"),
            schema=ln.examples.schemas.valid_features(),
        )

tests/curators/test_curators_remote.py:6: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
lamindb/curators/core.py:709: in __init__
    self._atomic_curator = ComponentCurator(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = ComponentCurator, artifact: F6PrJLEnWeHqUHrK0000(Schema: valid_features, unvalidated)
dataset = Artifact(uid='F6PrJLEnWeHqUHrK0000', is_latest=True, key='schmidt22/gws-crispr-ifng-hits.parquet', description='Hits f...e_id=1, storage_id=2, run_id=1099, schema_id=149, created_by_id=9, created_at=2025-10-17 03:07:46 UTC, is_locked=False)
schema = Schema(uid='0000000000000000', name='valid_features', is_type=False, itype='Feature', hash='kMi7B_N88uu-YnbTLDU-DA', m...alse, maximal_set=False, branch_id=1, space_id=1, created_by_id=1, created_at=2025-10-17 08:03:08 UTC, is_locked=False)
slot = None

    def __init__(
        self,
        dataset: pd.DataFrame | Artifact,
        schema: Schema,
        slot: str | None = None,
    ) -> None:
        super().__init__(dataset=dataset, schema=schema)
    
        categoricals = []
        features = []
        feature_ids: set[int] = set()
    
        if schema.flexible:
>           features += Feature.filter(name__in=self._dataset.keys()).to_list()
                                                ^^^^^^^^^^^^^^^^^^
E           AttributeError: 'Artifact' object has no attribute 'keys'. Did you mean: 'key'?

lamindb/curators/core.py:440: AttributeError
```

I believe that the inheritance of `SlotsCurator` from `Curator` needs to be re-thought, because we don't want to load artifacts into memory for each component or at least not both for the `SlotsCurator` and the components.

This re-architecting is beyond the scope of this PR; so I'm just adding a test for curating remote artifacts, which, it seems was missing and triggering this bug.

falexwolf:
Essentially, the `Curator` base class was defined for `ComponentCurator`, if you will. Given we're now in the "age of ComponentCurator" that original design choice doesn't make much sense anymore. 😅 

This being said: for as long as everything is tested a refactor isn't urgent.

sunnyosun:
Hmm weird we haven't ran into this, the `self._dataset` should be `self._artifact` here.

PR #3189, opened 2025-10-17
♻️ Enable bulk-re-creating records that would violate a unique constraint

sunnyosun:
This PR implements bulk saving of records that violate unique constraints due to other records in non-main branches.

This occurs mainly when records are imported from ontology sources as they have the same uids as records in trash.

```python

import bionty as bt

# first create records from public source
records = bt.Ethnicity.from_values(["asian", "white"]).save()
# one parent is also created
ethnicities = bt.Ethnicity.filter()
assert ethnicities.count() == 3
# soft delete the records including parent
ethnicities.delete()
# then create them again from public source
# the new records will now have the same uids as they are hashed from the ontology_ids
assert bt.Ethnicity.filter().count() == 0
new_records = bt.Ethnicity.from_values(["asian", "white", "african"])
assert new_records[0].branch_id == 1
assert new_records[0].uid == records[0].uid
# after saving, the trashed records should be restored
new_records.save()
assert new_records[0].branch_id == 1
ethnicities = bt.Ethnicity.filter()
# the parent should also be restored
assert ethnicities.count() == 4
```

### DIFF
--- a/lamindb/models/save.py
+++ b/lamindb/models/save.py
@@ -155,7 +155,6 @@ def bulk_create(
                 )
             try:
                 registry.objects.bulk_create(batch, ignore_conflicts=ignore_conflicts)
-                # records[:] = created  # In-place list update; does not seem to be necessary
             # handle unique constraint violations due to non-default branches
             except IntegrityError as e:
                 error_msg = str(e)
--- a/lamindb/models/sqlrecord.py
+++ b/lamindb/models/sqlrecord.py
@@ -931,6 +931,8 @@ class BaseSQLRecord(models.Model, metaclass=Registry):
                     init_self_from_db(self, pre_existing_record)
                 elif (
                     isinstance(e, IntegrityError)
+                    # for Storageeven if uid was in the error message, we can retrieve based on
+                    # the root because it's going to be the same root
                     and any(
                         field in error_msg for field in ("root", "ontology_id", "uid")
                     )


lamindb/models/save.py
```diff
@@ -153,8 +153,51 @@ def bulk_create(
                 logger.info(
                     f"Processing batch {batch_num}/{total_batches} for {model_name}: {len(batch)} records"
                 )
-            registry.objects.bulk_create(batch, ignore_conflicts=ignore_conflicts)
-            # records[:] = created  # In-place list update; does not seem to be necessary
+            try:
+                registry.objects.bulk_create(batch, ignore_conflicts=ignore_conflicts)
+                # records[:] = created  # In-place list update; does not seem to be necessary
```
falexwolf: Remove this comment?

lamindb/models/sqlrecord.py
```diff
@@ -912,50 +929,29 @@ def save(self: T, *args, **kwargs) -> T:
...
                     init_self_from_db(self, pre_existing_record)
-                elif (
-                    self.__class__.__name__ == "Storage"
-                    and isinstance(e, IntegrityError)
-                    and ("root" in error_msg or "uid" in error_msg)
-                    and unique_constraint_error_in_error_message(error_msg)
-                ):
-                    # even if uid was in the error message, we can retrieve based on
-                    # the root because it's going to be the same root
-                    pre_existing_record = self.__class__.get(root=self.root)
```
falexwolf, line 923: Could you add this as an example in the comments in the new abstracted implementation? Just so that it's easy for a developer to understand what's happening in terms of examples. The new code is good except for not providing these examples.

falexwolf (approved):
Great!

PR #3191, opened 2025-10-20
🐛 Exclude records in trash from unique constraint on `name` under a `type`, add `name` uniqueness under a `type` and `is_valid_type` constraints to all typed entities 

falexwolf:
While name uniqueness under a `type` was already enforced through validation on the Python level, it's now also enforced on the DB level.

### DIFF
--- a/lamindb/migrations/0136_remove_record_unique_name_type_space_and_more.py
+++ b/lamindb/migrations/0136_remove_record_unique_name_type_space_and_more.py
@@ -2,6 +2,136 @@
 
 from django.db import migrations, models
 
+CREATE_IS_VALID_FEATURE_TYPE_FUNCTION = """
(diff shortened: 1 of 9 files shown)


lamindb/migrations/0136_remove_record_unique_name_type_space_and_more.py
```diff
@@ -0,0 +1,64 @@
...
+
+    operations = [
+        migrations.RemoveConstraint(
+            model_name="record",
+            name="unique_name_type_space",
+        ),
+        migrations.AddConstraint(
+            model_name="feature",
+            constraint=models.UniqueConstraint(
+                condition=models.Q(("branch_id", -1), _negated=True),
```
sunnyosun, line 149: That's so cool!

PR #3196, opened 2025-10-20
🐛 Capture outputs of root logger in `LogStreamTracker`

falexwolf:


lamindb/core/_context.py
```diff
@@ -183,6 +183,8 @@ def start(self, run: Run):
             signal.signal(signal.SIGINT, self.cleanup)
         # handle exceptions
         sys.excepthook = self.handle_exception
+        # reset handler for lamin logger because sys.stdout has been replaced
+        set_handler(logger)
 
     def finish(self):
```
Koncopd, line 193: i think `set_handler(logger)` is also needed here.

falexwolf: Why?

Koncopd: so that it resets the reference to stdout when the redirection is cleared.

falexwolf: Right. Of course!

PR #3207, opened 2025-10-22
🚸 Ask for extra storage-level delete confirmation in case `_real_key` is not `None`

falexwolf:


falexwolf:
FYI: @Koncopd, do you agree this makes sense given that `_real_key is not None` essentially behaves like `not _key_is_virtual`, meaning the user has something in storage with a semantic key that they'd likely not want deleted?

I ran into this just now, accidentally deleting a test file with these specs:

image

falexwolf:
The new behavior is as for standard real keys (standard case of `_is_virtual_key` is `False` with `_real_key is None`):

```python
>>> import lamindb as ln
→ connected lamindb: laminlabs/lamindata
>>> ln.Artifact.get("eW679SbenFtwCQj60000").delete(permanent=True)
Artifact record deleted. Do you ALSO want to delete the data in storage at s3://lamindata/schmidt22_perturbseq/schmidt22_perturbseq.h5ad? (y/n) You can't undo this action.n
→ a file/folder remains here: s3://lamindata/schmidt22_perturbseq/schmidt22_perturbseq.h5ad
```

Koncopd:
"You can't undo this action.n"
What is this `.n` at the end?

Koncopd:
Yes, i agree. I just missed some parts where this check should have been also introduced.

falexwolf:
> What is this .n at the end?

It's just because I typed `n` as a response to the dialogue.

PR #3209, opened 2025-10-22
🚸 Allow features to have reserved Django field names

sunnyosun:
Resolves: https://github.com/laminlabs/pfizer-lamin-usage/issues/469

In cases when features names are the same with reserved field name of the registry, the django field name becomes `__lamindb_{model_name}_{field_name}__` in the dataframe.

### DIFF
--- a/tests/curators/test_curators_examples.py
+++ b/tests/curators/test_curators_examples.py
@@ -250,7 +250,7 @@ def test_dataframe_curator(mini_immuno_schema: ln.Schema):
     assert (
         artifact.features.describe(return_str=True)
         == """\
-Artifact .parquet · DataFrame · dataset
+Artifact: examples/dataset1.parquet
 └── Dataset features
     └── columns (5)
         cell_type_by_expe…  bionty.CellType         B cell, CD8-positive, alpha…
@@ -873,7 +873,7 @@ def test_spatialdata_curator(
     assert artifact.features.get_values()["disease"] == "Alzheimer disease"
     assert (
         artifact.features.describe(return_str=True)
-        == """Artifact .zarr · SpatialData · dataset
+        == """Artifact: examples/spatialdata1.zarr
 └── Dataset features
     ├── attrs:bio (2)
     │   developmental_sta…  bionty.DevelopmentalS…  adult stage
--- a/tests/curators/test_records.py
+++ b/tests/curators/test_records.py
@@ -132,7 +132,7 @@ id,uid,name,treatment,cell_line,preparation_date,project,__lamindb_record_uid__,
     assert (
         artifact.features.describe(return_str=True)
         == """\
-Artifact .csv · DataFrame · dataset
+Artifact: sheet_exports/My samples 2025-06.csv
 └── Dataset features
     └── columns (7)
         cell_line           bionty.CellLine         HEK293T cell
@@ -217,7 +217,7 @@ Sample_X,https://raw.githubusercontent.com/nf-core/test-datasets/scrnaseq/testda
     assert (
         artifact.features.describe(return_str=True)
         == """\
-Artifact .csv · DataFrame · dataset
+Artifact: sheet_exports/RNA-seq nextflow samplesheet 001.csv
 └── Dataset features
     └── columns (4)
         sample              Record[BioSample]       Sample_X, Sample_Y


lamindb/examples/fixtures/sheets.py
```diff
@@ -46,6 +46,9 @@ def populate_sheets_compound_treatment():
 
     # Samples ---------------------------
 
+    id_feature = ln.Feature(name="id", dtype=int).save()
```
falexwolf, line 50: Can you add a comment to document how these are special?

lamindb/models/query_set.py
```diff
@@ -550,13 +550,32 @@ def analyze_lookup_cardinality(
...
 ) -> pd.DataFrame:
+    """Reorder subset of columns in dataframe to specified position."""
     valid_columns = [col for col in column_order if col in df.columns]
     all_cols = df.columns.tolist()
     remaining_cols = [col for col in all_cols if col not in valid_columns]
     new_order = remaining_cols[:position] + valid_columns + remaining_cols[position:]
     return df[new_order]
 
 
+def _field_encoder(field_name: str) -> str:
```
falexwolf: Can we call this `encode_field_as_column_name()`?

lamindb/models/query_set.py
```diff
@@ -550,13 +550,32 @@ def analyze_lookup_cardinality(
...
+    """Encode laminDB specific fields in dataframe with __lamindb_record_{field_name}__."""
+    return f"__lamindb_record_{field_name}__"
+
+
+def _encode_lamindb_fields(
```
falexwolf: Can you use a more descriptive public function name?

lamindb/models/query_set.py
```diff
@@ -584,29 +603,32 @@ def reshape_annotate_result(
 
     # initialize result with basic fields, need a copy as we're modifying it
     # will give us warnings otherwise
-    result = df[field_names].copy()
+    fields_map = _encode_lamindb_fields(registry, df.columns)
```
falexwolf: Can you add a comment that at this point, the only columns that `df` contains are Django fields?

lamindb/models/query_set.py
```diff
@@ -584,29 +603,32 @@ def reshape_annotate_result(
...
+    result = df[field_names].copy().rename(columns=fields_map)
```
falexwolf: Why do we need this and the line below. Please add a comment.

lamindb/models/query_set.py
```diff
@@ -550,13 +550,32 @@ def analyze_lookup_cardinality(
...
+    """Reorder subset of columns in dataframe to specified position."""
     valid_columns = [col for col in column_order if col in df.columns]
     all_cols = df.columns.tolist()
     remaining_cols = [col for col in all_cols if col not in valid_columns]
     new_order = remaining_cols[:position] + valid_columns + remaining_cols[position:]
     return df[new_order]
 
 
+def _field_encoder(field_name: str) -> str:
+    """Encode laminDB specific fields in dataframe with __lamindb_record_{field_name}__."""
```
falexwolf: This is the name for `lamindb_record`. But what about `lamindb_artifact`, `lamindb_run`, `bionty_...` etc?

Shouldn't this encoding name reflect the name of the underlying SQL table?

I'd do `__lamindb_artifact_id__` for `ln.Artifact.id` in exact correspondence with the name in on the SQL level.

lamindb/models/query_set.py
```diff
@@ -668,9 +692,13 @@ def extract_single_element(s):
         result = reorder_subset_columns_in_df(result, feature_names)
 
     if cols_from_include:
-        result = process_cols_from_include(df, result, cols_from_include)
-
-    return result.drop_duplicates(subset=["id"])
+        result = process_cols_from_include(df_encoded, result, cols_from_include)
+    # here we revert the field encoding
+    # but in cases when the result already has a column with the field name (like a feature also called 'id'),
```
falexwolf: This unpredictable magic is a problem...

I'd say that if `features=True` we should always encode. But if `features=False` we should never encode.

Something like this. 🤔

sunnyosun:
The failed CI is due to an artifact was deleted from lamindata: F6PrJLEnWeHqUHrK0000

Merging this now and fix CI in the next PR.

PR #3211, opened 2025-10-22
♻️ Move `define_valid_features.py` and similar scripts into the package

falexwolf:
It was possible to create an environment in which the script in the `docs` folder is not discovered and hence we receive this error.

image

This refactor solves the problem at no downside. Code organization now appears more canonical.


falexwolf:
FYI @Zethson @sunnyosun

PR #3221, opened 2025-10-23
🐛 Fix exporting sheets with User feature

sunnyosun:
Define the User-Record M2M relationship in User, rather than Record, to be consistent with other models' definitions. This resolves the issue that feature of dtype User not being annotated, as the code that parses the linked model only deals with backward relationships.


sunnyosun:
@falexwolf this will cause a migration, but I guess it's fine?
```python


from django.db import migrations, models


class Migration(migrations.Migration):
    dependencies = [
        ("lamindb", "0137_squashed"),
    ]

    operations = [
        migrations.RemoveField(
            model_name="record",
            name="linked_users",
        ),
        migrations.AddField(
            model_name="user",
            name="linked_in_records",
            field=models.ManyToManyField(
                related_name="linked_users",
                through="lamindb.RecordUser",
                to="lamindb.record",
            ),
        ),
    ]


```

sunnyosun:
@falexwolf this test is failing because the artifact was removed by you I think: https://github.com/laminlabs/lamindb/actions/runs/18754533892/job/53502936747?pr=3221

Which new artifact should we use here instead? https://github.com/laminlabs/lamindb/blob/fe40e38036f8d01eb643cd04440a038e6aab928d/tests/curators/test_curators_remote.py#L7

falexwolf:
I fixed the test in my PR. Apologies.

lamindb/migrations/0138_remove_record_linked_users_user_linked_in_records.py
```diff
@@ -0,0 +1,25 @@
+# Generated by Django 5.1.8 on 2025-10-23 16:27
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("lamindb", "0137_squashed"),
```
falexwolf: This cannot depend on the squashed migration, it has to depend on the actual 0137 migration.

Please also rename 0137_squashed to 0138_squashed and append these two lines at the bottom.

falexwolf: Also, please add the 0138 migration here: https://github.com/laminlabs/lamindb/blob/b4e4f60481a5b618def4eab2e21ba71f30772ec5/lamindb/migrations/0137_squashed.py#L335

lamindb/models/query_set.py
```diff
@@ -446,14 +446,18 @@ def get_feature_annotate_kwargs(
...
             registry, obj.related_name
-        ).through.__get_name_with_module__(): obj.related_model.__get_name_with_module__()
+        ).through.__get_name_with_module__(): obj.related_model
         for obj in registry._meta.related_objects
         if obj.related_model.__get_name_with_module__() in cat_feature_types
     }
     if registry is Artifact:
-        link_models_on_models["ArtifactULabel"] = "ULabel"
+        from .ulabel import (
+            ULabel,  # needs to be imported here to avoid circular imports
```
falexwolf: At top of function, please.

tests/curators/test_records.py
```diff
@@ -245,3 +245,26 @@ def test_record_soft_deleted_recreate():
     assert record.branch_id == 1
     assert record.name == "test_record 2"
     ln.Record.objects.filter().delete()
+
+
+def test_annotate_with_user_feature():
```
falexwolf, line 250: Please delete everything you create here at the end of the test.

falexwolf (approved):
Tiny comments. Otherwise looks great!

Issue #3223, opened 2025-10-24
🐛 Curation of empty lists doesn't respect nullable


sunnyosun:
### Add a description

```python
feature_list = ln.Feature(name="list_tissue", dtype=list[bt.Tissue.ontology_id], nullable=True).save()
feature_int = ln.Feature(name="feature int", dtype=int, nullable=True).save()
schema = ln.Schema(name="test_list_feature_schema", features=[feature_list, feature_int], coerce_dtype=True).save()

df = pd.DataFrame(
    {
        "list_tissue": [],
        "feature int": []
    }
)
df
```

<img width="1224" height="619" alt="Image" src="https://github.com/user-attachments/assets/19f5df62-aa8b-40e2-85c3-f3600da784f9" />

PR #3224, opened 2025-10-24
🎨 Correctly handle empty lists in curation

sunnyosun:
Resolves: https://github.com/laminlabs/lamindb/issues/3223

The issue is that nullable in Pandera controls whether None/NaN values are allowed in the column, not whether the column can be empty (0 rows).

```python
# This is what nullable=True handles:
df_with_nulls = pd.DataFrame({
    "col": [1, 2, None, 4]  # None/NaN values in rows
})

# This is not handled:
df_empty = pd.DataFrame({
    "col": []  # 0 rows total
})
```

This PR implements the handling of empty columns in your `check_dtype` function to cover such cases.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,7 @@ dependencies = [
     "nbconvert>=7.2.1",  # avoid lxml[html_clean] dependency
     # Others
     "pyyaml",
-    "pyarrow",
+    "pyarrow<22.0.0",  # 22.0.0 breaks saving spatialdata objects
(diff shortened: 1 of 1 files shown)


sunnyosun:
The CI failed when saving `spatialdata`, due to the new release of `pyarrow 22.0.0` 3hrs ago.

I'll make a temporary upper bound for now so the CI passes.

falexwolf (approved):
Looks good!

PR #3231, opened 2025-10-25
🚸 Enable users to annotate & validate without having to learn about the `Schema` concept

falexwolf:
Enables short cuts for validating a `DataFrame` & `AnnData` so that a user doesn't have to learn about the `Schema` concept.

```python
ln.Artifact.from_dataframe(df, key="...", schema="valid_features").save()
ln.Artifact.from_anndata(adata, key="...", schema="ensembl_gene_ids_and_valid_features_in_obs").save()
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -176,6 +176,7 @@ jobs:
       - run: nox -s "install_ci(group='docs')"
       - uses: actions/download-artifact@v4
       - run: nox -s clidocs
+      - run: nox -s cp_scripts
       - run: nox -s docs
       - run: rm -r ./_build/html/.doctrees # do not want to deploy with cloudflare
       - uses: cloudflare/wrangler-action@v3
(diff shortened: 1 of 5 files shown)


falexwolf:
Also fixes the `curate_dataframe_external_features.py` example, which assumed `feat1` and `feat2`, which aren't part of the mini immuno dataset. This is broken on the main docs.

image

FYI @Zethson

falexwolf:
Something is off with the `date` validation in the context of the external schema validation. Need to investigate.

falexwolf:
We're accepting an ISO-format date in `features.add_values()` but `pandera` doesn't accept it without `coerce_dtype = True`.

In particular given that also parquet has a dedicated 32bit integer-based `DATE` dtype we should adopt the behavior from `pandera`. I'll do this in another PR.

Study: https://lamin.ai/laminlabs/lamindata/transform/U2T805mjF6TL0001

falexwolf:
There is another issue with the "external features" example pointing to a potential problem of idempotency when re-curating an aritfact with multiple slots. 🤔 

image

This has nothing to do with this PR, so I'll fix it in another one.

PR #3237, opened 2025-10-27
🐛 Allow `features.remove_values()` to work with features of `dtype=list[cat]`

sunnyosun:
`features.remove_values()` now works with features of dtype list[cat].

Screenshot 2025-10-28 at 11 15 12

### DIFF
--- a/lamindb/models/_feature_manager.py
+++ b/lamindb/models/_feature_manager.py
@@ -1189,20 +1189,28 @@ class FeatureManager:
                 for obj in Artifact._meta.related_objects
                 if obj.related_model.__get_name_with_module__() in link_models_on_models
             }.pop()
-            feature_values = getattr(self._host, link_attribute).filter(**filter_kwargs)
+
+            link_records = getattr(self._host, link_attribute).filter(**filter_kwargs)
+            if not link_records.exists():
+                logger.warning(
+                    f"no feature '{feature_record.name}' with value '{value}' found on artifact '{self._host.uid}'!"
+                )
+                return
+            link_records.delete()
         else:
             if value is not None:
                 filter_kwargs["value"] = value
+
             feature_values = self._host._feature_values.filter(**filter_kwargs)
-        if not feature_values.exists():
-            logger.warning(
-                f"no feature '{feature_record.name}' with value '{value}' found on artifact '{self._host.uid}'!"
-            )
-            return
-        feature_values.delete()
-        # this might leave a dangling feature_value record
-        # but we don't want to pay the price of making another query just to remove this annotation
-        # we can clean the FeatureValue registry periodically if we want to
+            if not feature_values.exists():
+                logger.warning(
+                    f"no feature '{feature_record.name}' with value '{value}' found on artifact '{self._host.uid}'!"
+                )
+                return
+            self._host._feature_values.remove(*feature_values)
+            # this might leave a dangling feature_value record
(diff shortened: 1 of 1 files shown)


falexwolf:
Yes, please add tests, and then let's merge.

falexwolf:
Here we should print a warning if no error was found:

image

Alternatively we should throw an error. Django throws an error upon `.remove()` if the thing that's to be removed doesn't exist.

lamindb/models/_feature_manager.py
```diff
@@ -1186,11 +1189,22 @@ def remove_values(
                 for obj in Artifact._meta.related_objects
                 if obj.related_model.__get_name_with_module__() in link_models_on_models
             }.pop()
-            getattr(self._host, link_attribute).filter(**filter_kwargs).delete()
+            feature_values = getattr(self._host, link_attribute).filter(**filter_kwargs)
+            if not feature_values.exists():
```
falexwolf: Why is this duplicated with the statement in L1203?

lamindb/models/_feature_manager.py
```diff
@@ -1186,11 +1189,22 @@ def remove_values(
...
+                logger.warning(
+                    f"no feature '{feature_record.name}' with value '{value}' found on artifact '{self._host.uid}'!"
+                )
+                return
+            feature_values.delete()
```
falexwolf: What if this feature value is also linked to something else. It shouldn't be deleted then, right? There has to be a conditional check similar to what we have below L1200.

tests/core/test_feature_label_manager.py
```diff
@@ -359,6 +359,26 @@ def test_features_add_remove(adata):
     ln.Feature(name="study_metadata", dtype=dict).save()
     artifact.features.add_values({"study_metadata": {"detail1": "123", "detail2": 1}})
 
+    # test add/remove a list feature
+    ln.Feature(name="list_of_numbers", dtype=list[float]).save()
+    artifact.features.add_values({"list_of_numbers": [1.0, 2.0, 3.0]})
+    assert artifact.features.get_values()["list_of_numbers"] == [1.0, 2.0, 3.0]
+    artifact.features.remove_values("list_of_numbers")
+    # remove a non-linked value, this should do nothing but print a warning
```
falexwolf: Can you test for the warning with ccaplog?

lamindb/models/_feature_manager.py
sunnyosun: I don't know what you are talking about here, L1192 already ensures they are linked to the self._host.

tests/core/test_feature_label_manager.py
falexwolf: It's also worthwhile considering writing a new test function. I wanted to pull this test function apart anyway; it grew much too big.

lamindb/models/_feature_manager.py
falexwolf: But they can also be linked to something else, right? This isn't a link record, is it?

falexwolf: I think you're right and this is indeed a `feature_value_link` record rather than a `feature_value` record. The later would be an object of the `FeatureValue` registry.

So, the logic here is correct.

falexwolf:
Once this is merged I'll work on the high-level `Record()` API ☺️

PR #3238, opened 2025-10-27
⚡️ More optimal recursive childrens query

falexwolf:


lamindb/models/has_parents.py
```diff
@@ -43,9 +43,13 @@ def query_relatives_on_branches(records, attr, cls) -> QuerySet:
...
         for record in records:
-            relatives = relatives.union(
-                getattr(record, attr).filter(branch_id__in=branch_ids)
-            )
+            more_relatives = getattr(record, attr).filter(branch_id__in=branch_ids)
+            # avoid unnecessary union calls, say something has 30 children these grow a lot
+            # this is a database request but still faster than union calls
+            # we never ran into an issue for Postgres, but for SQLite
+            # it might be that Postgres is smart enough to optimize this away
+            if more_relatives.exists():
```
Koncopd: i had problems with this exists check, it was slowing down the whole function significantly for some reason.

falexwolf: I see. Good that you tell me. Then I'll remove it; I also have another solution.

falexwolf: Removed it. I feared it would slow down because it's splitting the whole thing into many queries.

There is another simplification (see latest two commits), which mean that we're unlikely to ever run into this also for SQLite.

The previous implementation produced a problem when it tried to query 30 records of a type in `ln.Record`.

PR #3239, opened 2025-10-27
🚸 Enable passing `parquet_kwargs` to `Artifact.from_dataframe()` for configuring `pyarrow.ParquetWriter`

falexwolf:
Before | After
--- | ---
image | image
/ | image

Addresses:

- https://github.com/laminlabs/pfizer-lamin-usage/issues/216

### DIFF
--- a/lamindb/models/artifact.py
+++ b/lamindb/models/artifact.py
@@ -284,6 +284,8 @@ def process_data(
     if memory_rep is not None:
         path = settings.cache_dir / f"{provisional_uid}{suffix}"
         logger.important("writing the in-memory object into cache")
+        if parquet_kwargs is None:
+            parquet_kwargs = {}
         write_to_disk(data, path, **parquet_kwargs)
         use_existing_storage_key = False
 


lamindb/models/artifact.py
```diff
@@ -282,12 +283,10 @@ def process_data(
     # in case we have an in-memory representation, we need to write it to disk
     if memory_rep is not None:
         path = settings.cache_dir / f"{provisional_uid}{suffix}"
-        if isinstance(format, dict):
-            format.pop("suffix", None)
-        else:
-            format = {}
         logger.important("writing the in-memory object into cache")
-        write_to_disk(data, path, **format)
```
falexwolf, line 290: @Koncopd, I suspect that this is from many years ago. I don't think this makes sense and I didn't find a test for a dict-like `format` argument.

lamindb/models/artifact.py
```diff
@@ -282,12 +283,10 @@ def process_data(
...
+        if parquet_kwargs is None:
+            parquet_kwargs = {}
+        write_to_disk(data, path, **parquet_kwargs)
```
Koncopd, line 289: i don't think it is correct to call them `parquet_kwargs`, `write_to_disk` is not only for `parquet`.

Koncopd: i would call it `writer_kwargs` or something like that.

lamindb/models/artifact.py
Koncopd: i need to check. I remember there was something for `anndata`.

lamindb/models/artifact.py
falexwolf: If the abstraction stays the way it is this could, at this level in the calling cascade, have a generic name.

However, at the present time, the only configurable write we use is the parquet writer. And if we introduced a generic name we could not type it properly (e.g. with the signature of `ParquetWriter`).

Once we support the same e.g. for zarr I see two options:

1. the code has been refactored and dataframe/parquet is its own thing and we want to keep this name
2. the code still shares the same abstractions across all formats and we'll have a user-facing `zarr_kwargs` in addition to the `parquet_kwargs`

So, in case 2 we'll be facing the question of maintaining integrity of `zarr_kwargs` vs. `parquet_kwargs` and it might be easier to keep each of these objects around, type them with `ParquetWriterKwargs` and `ZarrKwargs` or something like this, and then pass them to the dedicated writer.

All this being said: I think we should keep this explicit name right now and work towards refactoring the code base towards at least separating the call cascades for in-memory and on-disk/in-cloud.

lamindb/models/artifact.py
falexwolf: The logic for anndata is handled upstream.

lamindb/models/artifact.py
Koncopd: But then we will have to support `parquet_kwargs` also and drag it along for a long time. I think this is a mistake and we can use a more general name from the start.

PR #3242, opened 2025-10-28
🚸 Also introduce `csv_kwargs` for `Artifact.from_dataframe()`

falexwolf:
Is follow-up to:

- https://github.com/laminlabs/lamindb/pull/3239

Also reorganizes the `records/sheets` tests because they test some of this.

### DIFF
--- a/tests/core/test_schema.py
+++ b/tests/core/test_schema.py
@@ -56,8 +56,6 @@ def test_schema_from_values():
         schema = ln.Schema.from_values(
             ["weird_name"], field=ln.Feature.name, dtype="float"
         )
-    with pytest.raises(ValidationError):
-        ln.Schema.from_values(["name"], field=ln.Feature.name, dtype="float")
 
 
 def test_schema_from_records(df):


Koncopd:
Why can't we have single `writer_kwargs` for both things (and more in the future). I think it is better than adding endless different `_kwargs`.

falexwolf:
> Why can't we have single writer_kwargs for both things (and more in the future). I think it is better than adding endless different _kwargs.

I thought because of this https://github.com/laminlabs/lamindb/pull/3239#discussion_r2466995689

image

The `writer_kwargs` could neither be easily typed nor documented + they're **not** self-explaining API. Something like `parquet_kwargs` or `csv_kwargs` is self-explaining and can be easily typed and documented.

The `write_kwargs` would be some custom lamindb thing that could mean anything.

I'd agree if we'd expect 20 different writers per data structure; but we'll be having maybe 10. And each of these 10 deserves their own type and docs. So, I think it's clearer if it has a dedicated name.

Also in the internal codebase I expect things to not go through the same functions anymore.

If it was a `writer` argument that could be `ParquetWriter`, `CsvWriter`, etc. I'd agree with you. But given it's a nested complicated thing, I'm pretty sure the code is easier to read & understand if we have clear self-explaining names.

All this being said: I'm also on the edge.

falexwolf:
Here:

image

falexwolf:
I'll keep it this way now. We can sleep on it for another night before the next release.

lamindb/models/record.py
```diff
@@ -402,7 +400,7 @@ def to_artifact(self, key: str = None) -> Artifact:
             key=key,
             description=f"Export of sheet {self.uid}{description}",
             schema=self.schema,
-            format=format,
```
falexwolf, line 405: @Koncopd, I hadn't spotted this breakage in the `parquet_kwargs` PR because all my search for `format` didn't yield anything informative.

Now, if I want to search for how configuring a csv export is used I'll search for `csv_kwargs` and immediately find this test case.

I'm saying this because `format` was clearly an instance of an overly generic name and usage pattern.

Koncopd: I am not even sure what this `format` is. I know we have `format` for `anndata` `zarr` vs `h5ad`, but it is not this, right?

falexwolf: It was abused in this way. I think we should remove the `format` kwarg entirely because we never officially documented and supported it. But that's for another time.

PR #3249, opened 2025-10-29
🐛 Fix edge cases in `to_dataframe()` with features

Koncopd:
On `lamindata` both were failing:
```python
ln.Artifact.to_dataframe(include="features")
ln.Artifact.to_dataframe(features="queryset")
```
This fixes it and also fixes https://github.com/laminlabs/pfizer-lamin-usage/issues/490

### DIFF
--- a/lamindb/models/query_set.py
+++ b/lamindb/models/query_set.py
@@ -430,7 +430,7 @@ def get_feature_annotate_kwargs(
             )
         features = list(set(feature_names))  # remove duplicates
 
-    feature_qs = Feature.using(qs.db if qs is not None else None).filter(
+    feature_qs = Feature.using(None if qs is None else qs.db).filter(
         dtype__isnull=False
     )
     if isinstance(features, list):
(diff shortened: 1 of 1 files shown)


Koncopd:
This is pretty slow though.

falexwolf (approved):
Looks good!

falexwolf:
Just the test needs to be fixed.

falexwolf:
Actually, rather than querying `lamindata` we should create a unit test that maps this failure mode.

Also we should keep the `features=True == includes="features"` working as it was and also test the legacy call `features=True`.

Koncopd:
> Actually, rather than querying lamindata we should create a unit test that maps this failure mode.

It is really hard to reproduce all the problems in a unit test. This is why i use `lamindata`.

falexwolf:
> It is really hard to reproduce all the problems in a unit test. 

We still have to attempt it. I'm only aware of a single problem, which is the `dtype=None` issue. That one is easy to reproduce.

What we can't reproduce in a unit test is not well tested.

I can take over and create the test cases if you tell me what the problems are. But the `lamindata`-based test is not an option I think (as an integration test, sure, but I don't think we need it as an integration test; we need a unit test in `unit-core`).

Koncopd:
As discussed, i am merging it and @falexwolf can work on optimizing and adding test cases for this.

PR #3256, opened 2025-10-31
🐛 Deal with corrupted state in which no `is_latest` record is found

falexwolf:


falexwolf:
Still need to add a test.

PR #3272, opened 2025-11-10
⬆️ No longer upper-bound pyarrow by `22.0`

Zethson:
Required for Python 3.14.


falexwolf:
Great!

Koncopd:
But was the problem with spatialdata resolved?

sunnyosun:
curators tests aren't run in this PR, so we can't confirm the error is gone.

Zethson:
Yes, we should run them but I can vouch that this specific pyarrow error is fixed with the latest SpatialData release. Any other SpatialData issues should not be related to this PR.

PR #3277, opened 2025-11-11
🚚 Deprecate the `data` argument of `Artifact` in favor of `path` and polish class docstrings

falexwolf:
The name choice `data` is reminiscent from when we still allowed other types in the default `Artifact()` constructor.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ Highlights:
 - **integrations** → [vitessce](https://docs.lamin.ai/vitessce), [nextflow](https://docs.lamin.ai/nextflow), [redun](https://docs.lamin.ai/redun), and [more](https://docs.lamin.ai/integrations)
 - **extendable** → create custom plug-ins based on the Django ORM
 
-If you want a GUI, you can connect your LaminDB instance to LaminHub and close the drylab-wetlab feedback loop: [lamin.ai](https://lamin.ai).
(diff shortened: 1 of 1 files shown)


lamindb/models/artifact.py
```diff
@@ -1411,9 +1409,17 @@ def __init__(
             return None
         # now proceed with the user-facing constructor
         if len(args) > 1:
-            raise ValueError("Only one non-keyword arg allowed: data")
+            raise ValueError("Only one non-keyword arg allowed: path")
 
-        data: str | Path = kwargs.pop("data") if len(args) == 0 else args[0]
+        if "data" in kwargs:
+            warnings.warn(
+                "`data` argument will be renamed to `path` in a future release.",
```
Zethson: But it's already been renamed here, right? It will be removed in a future release.

```suggestion
                "`data` argument was renamed to `path` and will be removed in a future release.",
```

so something like this?

falexwolf: Yes, thanks for spotting! This was tab complete from AI ☺️

Koncopd:
So we don’t allow creating artifacts from in-memory objects anymore?

falexwolf:
We do, but not through that constructor.

falexwolf:
Internals have to be re-factored and this is long overdue.

But when I just read the Artifact docs page like a new user, the `data` argument was so strikingly wrong that I wanted to get this out of the way before the internal refactor.

falexwolf:
We had already updated the type annotation a long time ago; I don't know why `data` as a name still stuck around.

Issue #2222, opened 2024-11-27
Enable CI for forks


Zethson:
### Add a description

Works in bionty now but not yet in lamindb. See https://github.com/laminlabs/lamindb/pull/2198 where it just failed early.

PR #3281, opened 2025-11-12
✨ Add more lightning integration with an example

jorenretel:
I recently was experimenting with a lightning checkpointing and save_config callback that writes to lamin. This integrates nicely with the lightning CLI, which is really the nicest way to use lightning in my opion. 

Probably this integration needs a  bit more polishing. Some things to figure out:

1. I probably went overboard a bit with using the lamin features with all the pruning of saved checkpoints. It likely becomes a bit  untransparent what is going on because I use “versions” of the same file to keep the top k best checkpoints and the latest checkpoint around. But to a user it would be way more obvious to really see all files under a given prefix (for instance when topk=5, you would expect 5, or maybe 6 when the latest checkpoint is not amount top k, files). The only question is how to name these files. In principle the labels in lamin already tell exactly what you need to know (things like: this model was the best, or the second best, or the last, whatever), but still, you can not have 5 files in a directory all called model.ckpt. So in principle <random-string>.ckpt (or step123-epoch456.ckpt) would be fine.  But then at the same time, for downstream applications (see next point) you would prefer some stable naming that will  give you the same  5 file names in the end of each experiment/version (like latest.ckpt, top1.ckpt, top2.ckpt etc.) 

2. We almost immediately discovered that deployment routines need semantic paths on s3.  For instance, sagemaker expects to get a prefix on s3 under which all the files live (possibly nested) that it will copy to a specific path in the deployment container, which the machine learner that creates the code in that container can use again to re-instantiate the model.

Both these points are related obviously and have something to do with the relationship between keys on lamin and paths expected by routines/users using these artefacts again. I feel like you guys probably already have some idea what a good solution would like like.

### DIFF
--- a/lamindb/examples/lightning/README.md
+++ b/lamindb/examples/lightning/README.md
@@ -1,0 +1,109 @@
+# lamin-lightning
+
+A Python package that integrates [LaminDB](https://lamin.ai) with [PyTorch Lightning](https://lightning.ai/pytorch-lightning) to streamline model checkpointing.
+
+## tl;dr
+Take a look at the [example](lamindb/examples/lightning) directory for a full example of how
+to use this package with PyTorch Lightning CLI.
+
+In a environment with pytorch, lightning and lamindb installed:
+
+```bash
+cd lamindb/examples/lightning
+python main.py fit --config lightning_config.yml
+```
+
+## Quick Start
+
+```python
+from lightning import Trainer
+from lamindb.integrations.lightning.checkpoint import LaminCheckpoint
+
+checkpoint_callback = LaminCheckpoint(
+    lamin_instance="your/instance",
+    lamin_root_dir="checkpoints",
+    monitor="val_loss",
+    save_top_k=3
+)
+
+trainer = Trainer(callbacks=[checkpoint_callback])
+```
+
+## Using LaminCheckpoint
+
+The `LaminCheckpoint` extends Lightning's `ModelCheckpoint` to save checkpoints to your LaminDB instance.
+
+### Configuration Options
+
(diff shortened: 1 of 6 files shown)


Zethson:
Thank you very much! I'll have a proper look soon.

The CI is failing because forks do not have access to required credentials. There's a chance that eventually we need to recreate the PR and then credit it to you.

falexwolf:
Thanks, Joren! Awesome!

> The CI is failing because forks do not have access to required credentials.

Right, we should just have given you write access to repo (I'll do this now just in case).

But I think we should also be able to allow the CI run in this case?

> There's a chance that eventually we need to recreate the PR and then credit it to you.

Let's try to avoid this! ☺️ -- Should be feasible.

falexwolf:
You have write access to lamindb now, Joren!

falexwolf:
Hm, I don't find a way to give the fork CI run access to secrets. I thought it's possible. Hm.

I suggest we merge this onto a lamindb branch and continue from there!

falexwolf:
We could also merge into main and then make a new PR to refactor -- let's just give it a little bit of time because we just made 1.16.0 and a bug fix release might be necessary and that's nicer to do from main than the `release` branch. ☺️

Zethson:
Dear @jorenretel ,

I'll continue your PR in https://github.com/laminlabs/lamindb/pull/3425 that you'll automatically co-author. As discussed above, our CI currently requires access to various secrets which forked repositories cannot get. Resolving this is tracked in #2222 .

I'll keep you posted in the other PR and will eventually also ask you for feedback. Thank you very much for your work already! Thanks to your ideas, the new implementation will surely be much better than before.

PR #3282, opened 2025-11-12
✨ Read-only queries of public instances through shadow SQLite clones

Zethson:
This PR enforces querying public LaminDB instances through local SQLite clones instead of remote PostgreSQL connections. When calling `Registry.connect("account/public-instance")`, the system now automatically downloads and connects to a cached SQLite copy of the instance, providing efficient read-only access without requiring database credentials. Private instances continue to use direct database connections as before, preserving existing behavior for authenticated access.


lamindb/models/sqlrecord.py
```diff
@@ -627,6 +628,40 @@ def using(
             instance=instance,
         )
 
+    @staticmethod
+    def _synchronize_clone() -> str:
```
Koncopd: i don't think this should be here.

Zethson: First we make it work, then we make it pretty. The logic is currently still wrong.

lamindb/models/sqlrecord.py
```diff
@@ -659,6 +662,41 @@ def connect(
...
+
+            Args:
+                storage_root: The storage root path of the (target) instance
+            """
+            cloud_db_path = UPath(storage_root) / ".lamindb" / "lamin.db"
+            local_sqlite_path = ln_setup.settings.cache_dir / cloud_db_path.path.lstrip(
+                "/"
+            )
+            local_sqlite_path.parent.mkdir(parents=True, exist_ok=True)
+            cloud_db_path_gz = UPath(str(cloud_db_path) + ".gz", anon=True)
```
falexwolf: If we want this to work for private storage locations, too, we need to use `create_path()` but that's for the future.

lamindb/models/sqlrecord.py
```diff
@@ -659,6 +662,41 @@ def connect(
...
+            local_sqlite_path_gz = Path(str(local_sqlite_path) + ".gz")
+
+            try:
+                cloud_db_path_gz.synchronize_to(
+                    local_sqlite_path_gz, error_no_origin=True, print_progress=True
+                )
+            except (FileNotFoundError, PermissionError):
+                try:
+                    cloud_db_path.synchronize_to(
```
falexwolf: Is there ever a use case in which we'd not want to compress the sqlite file? Testing both for compressed and uncompressed makes things complicated.

Ideally we only support the compressed sqlite file. The compression takes no time in comparison to IO and the sqlite file isn't queryable on S3 anyway. So I don't see the point of allowing the uncompressed file.

lamindb/models/sqlrecord.py
```diff
@@ -659,6 +662,41 @@ def connect(
...
+                        local_sqlite_path, error_no_origin=True, print_progress=True
+                    )
+                except (FileNotFoundError, PermissionError):
+                    logger.debug("Clone not found. Falling back to normal access...")
+                    return None
+            else:
+                with gzip.open(local_sqlite_path_gz, "rb") as f_in:
+                    with open(local_sqlite_path, "wb") as f_out:
+                        shutil.copyfileobj(f_in, f_out)
+                local_sqlite_path_gz.unlink()
```
falexwolf: Shouldn't you keep the `local_sqlite_path_gz` around to avoid re-download? 🤔 I don't really understand because it seems to already work.

lamindb/models/sqlrecord.py
```diff
@@ -659,6 +662,41 @@ def connect(
 
                 ln.Record.connect("account_handle/instance_name").search("label7", field="name")
         """
+
+        def _synchronize_clone(storage_root: str) -> str | None:
```
Koncopd: why is it created dynamically?

lamindb/models/sqlrecord.py
```diff
@@ -693,28 +729,56 @@ def connect(
             source_modules = set(  # noqa
                 [mod for mod in iresult["schema_str"].split(",") if mod != ""]
             )
-            # this just retrives the full connection string from iresult
-            db = update_db_using_local(iresult, settings_file)
+
+            # Try to connect to a clone if targeting a public instance but fall back to normal access if access failed
+            db = None
+            if (
+                "_jwt" in iresult["db_user_name"]
```
Koncopd: why does it check for jwt? it has to check for public, no? We want public to go through clone, not jwt.

lamindb/models/sqlrecord.py
```diff
@@ -693,28 +729,56 @@ def connect(
...
+
             # access_db can take both: the dict from connect_instance_hub and isettings
             into_db_token = iresult
         else:
             isettings = load_instance_settings(settings_file)
             source_modules = isettings.modules
-            db = isettings.db
+            db = None
+            if "jwt" in isettings.db and isettings.dialect == "postgresql":
+                db = _synchronize_clone(isettings.storage.root_as_str)
```
Koncopd, line 759: same here, why does it sync for jwt?

tests/core/test_queryset.py
```diff
@@ -327,3 +329,31 @@ def test_encode_lamindb_fields_as_columns():
         "name": "__lamindb_record_name__",
         "created_by": "__lamindb_record_created_by__",
     }
+
+
+def test_connect_public_clone_instance():
+    env = os.environ
+    env["LAMIN_TESTING"] = "true"
+
+    # testuser1 is admin of lamindata but testuser2 is not
```
falexwolf: I thought that read-level collaborators and non-collaborators should get the sqlite clone, but write-/admin-collaborators should get the postgres connection string because they might want to write.

If we test it with `testuser2` we have to ensure that they are a non-collaborator, and in the future (when expanding to private instances), a read-collaborator.

lamindb/models/sqlrecord.py
Zethson: The function?

No need for that. I had moved it internally because you didn't like where I had it before (just as a normal function of the class). I'm not sure where to put it but am open to suggestions.

falexwolf:
One more issue: it seems that the cache location isn't shared across instances. Why is that? Our cache is organized by storage location and not by instance so that it doesn't matter from which instance I load an artifact.

But I know for sure that I already downloaded the sqlite file for `cellxgene` when connected to `laminlabs/lamindata` as the current instance.

image

But now I know call `ln.Artifact.connect()` again from my local test sqlite instance and it re-syncs.

Can you double-check that the sqlite file is in the canonical cache, which would in this case mirror the default storage location of `cellxgene`?

[Edit] This is likely what happens: https://github.com/laminlabs/lamindb/pull/3282#discussion_r2541884988

tests/core/test_queryset.py
Zethson: `testuser2` is not a collaborator. I'll update the comment.

lamindb/models/sqlrecord.py
falexwolf: I think this is indeed a bug. Now with more experiments it seems that the cache is gone after re-starting the Python session. It just keeps re-downloading

image

falexwolf: Also explains this: https://github.com/laminlabs/lamindb/pull/3282#issuecomment-3552535011

lamindb/models/sqlrecord.py
Zethson: I'll double check this but I couldn't find a case where it included `public` even when logged out.

falexwolf:
I suggest that you take our rather comprehensive transfer tests, @Zethson, and run them through the SQLite file.

The entire test suite here is only to check read-only access to transferred objects: https://github.com/laminlabs/lamindb/blob/main/tests/storage/test_transfer.py

Another good test would be to save the small immuno dataset to `lamindata` (if it's not already there) and then run this test: https://github.com/laminlabs/lamindb/blob/734367caa7b4bff5216f4aca1d8e43fe88bb8b0a/tests/core/test_artifact_describe_to_dataframe.py#L129-L148

The devil is in the detail when it comes to ascertaining that all data in the SQLite file is 100% correct.

falexwolf:
You're right @Zethson, there is an issue with `.describe()` on that artifact in the original instance.

image

lamindb/models/sqlrecord.py
Koncopd: you have to log out and then re-cconnect. Using clone for jwt is incorrect, jwt is issued to collaborators or org memebers.

lamindb/models/sqlrecord.py
Koncopd: Just as a separate function?

Zethson:
> I suggest that you take our rather comprehensive transfer tests, @Zethson, and run them through the SQLite file.
> 
> The entire test suite here is only to check read-only access to transferred objects: https://github.com/laminlabs/lamindb/blob/main/tests/storage/test_transfer.py

These tests were already run on CI using our clones. Initially, there were failures but it should be good now.

> Another good test would be to save the small immuno dataset to lamindata (if it's not already there) and then run this test: https://github.com/laminlabs/lamindb/blob/734367caa7b4bff5216f4aca1d8e43fe88bb8b0a/tests/core/test_artifact_describe_to_dataframe.py#L129-L148

Also LGTM

falexwolf:
Well then let's merge this and close the loop in the docs! ☺️

falexwolf:
The team can play with this tomorrow and during the weekend and by Monday/Tuesday we could be ready for launch because no hiccups with firewalls anymore?

falexwolf:
Documentation for how to create a clone best go here?

- https://github.com/laminlabs/lamin-docs/blob/main/docs/setup.ipynb

And in the auto-generated CLI docs?

Or maybe just in the auto-generated CLI docs is enough. 🤔

falexwolf:
Probably just CLI docs is enough because we only create confusion if we make this part of the setup process I guess.

Zethson:
Yeah but I haven't added a `sync-clone` command to the CLI yet. I'll do that very soon. I figured out why the script didn't behave as expected. The subprocess closed too early and not all changes could be committed in time.

falexwolf:
Ok! We should also think about the right design for the `sync-clone` command. One could make an argument that it should not be at the root level of the command hierarchy similar to how `settings` is it's own command group.

We could have an `admin` group or similar to avoid cluttering the namespace UX for the regular user.

image

falexwolf:
After a long transition period `migrate` could also only be available from within the `admin` name space.

PR #3302, opened 2025-11-20
♻️ Add `query_ulabels()`, `query_features()`, `query_schemas()`, etc. to all `HasType` registries

falexwolf:


lamindb/models/has_parents.py
```diff
@@ -32,7 +32,7 @@
 # also len of QuerySet can be costly at times
 def _query_relatives(
     records: BasicQuerySet | list[HasParents],
-    attr: Literal["children", "parents", "records"],
+    attr: Literal["children", "parents"] | str,
```
Koncopd, line 35: this type doesn't make any sense to me, it should be just `str` then.

lamindb/models/has_parents.py
```diff
@@ -43,12 +43,12 @@ def _query_relatives(
...
-    if attr == "records":
-        attr_filter = "type__id__in"
-    elif attr == "children":
+    if attr == "children":
         attr_filter = "parents__id__in"
-    else:
+    elif attr == "parents":
         attr_filter = "children__id__in"
+    else:
+        attr_filter = "type__id__in"
```
Koncopd, line 51: if it is the same for everything else (is this correct @falexwolf ?), then why not keep `Literal["children", "parents", "records"]` and just pass `"records"`?

falexwolf: Because it's not the same. This was a bug here.

falexwolf: See all the examples in this diff!

lamindb/models/has_parents.py
falexwolf: We can spell out `projects`, `references`, `features`, `schemas`, `records`, `ulabels` in the next change.

PR #3308, opened 2025-11-22
🚸 Make adding optional features convenient through `schema.add_optional_features()`

falexwolf:
Given an existing `schema` the user journey is now as follows.

Before | After
--- | ---
https://lamin.ai/laminlabs/lamindata/transform/XLajjnyN3lEX0000 | https://lamin.ai/laminlabs/lamindata/transform/XLajjnyN3lEX0001
image | image
image | 

The simplification is possible because adding and removing _optional_ features does not change the integrity of existing datasets and sheets.

Also in this PR: docs improvements.

Before | After
--- | ---
image | image
image | image

### DIFF
--- a/lamindb/models/schema.py
+++ b/lamindb/models/schema.py
@@ -233,7 +233,7 @@ class Schema(SQLRecord, HasType, CanCurate, TracksRun):
 
             schema = ln.Schema([ln.Feature(name="required_feature", dtype=str).save()]).save()
 
-        A schema that constraints feature identifiers to be a valid feature names::
+        A schema that constrains feature identifiers to be a valid feature names::
 
             schema = ln.Schema(itype=ln.Feature)  # uses Feature.name as identifier type
 


lamindb/models/schema.py
```diff
@@ -1158,6 +1160,25 @@ def optionals(self) -> SchemaOptionals:
         """
         return SchemaOptionals(self)
 
+    def add_optional_features(self, features: list[Feature]) -> None:
+        """Add optional features to the schema."""
+        assert self.name is not None, "schema must have a name to add optional features"
```
Zethson, line 1165: Shouldn't this be a ValueError? Also shouldn't we suggest a resolution -> do xyz to add a name to schema

lamindb/models/schema.py
```diff
@@ -1158,6 +1160,25 @@ def optionals(self) -> SchemaOptionals:
...
+        self.features.add(*features)
+        self.optionals.add(features)
+        self.save(print_hash_mutation_warning=False)
+
+    def remove_optional_features(self, features: list[Feature]) -> None:
+        """Remove optional features from the schema."""
+        assert self.name is not None, (
```
Zethson, line 1172: See above.

lamindb/models/schema.py
```diff
@@ -1229,3 +1250,4 @@ class Meta:
 delattr(Schema, "validated_schemas")  # we don't want to expose these
 delattr(Schema, "composite")  # we don't want to expose these
 delattr(Schema, "composite_id")  # we don't want to expose these
+delattr(Schema, "slot")  # we don't want to expose these
```
Zethson, line 1253: I think what would be better is a single comment for all of the attributes that we're removing justifying why we're removing them instead of these comments.

lamindb/models/schema.py
```diff
@@ -179,21 +179,21 @@ def add(self, features: Feature | list[Feature]) -> None:
...
+    .. note::
 
-    To create a schema, at least one of the following arguments must be passed:
+        To create a schema, at least one of the following parameters must be passed:
 
-    - `features` - a list of :class:`~lamindb.Feature` records
-    - `itype` - a registry or field to constrain feature identifiers, e.g., `ln.Feature` or `bt.Gene.ensembl_gene_id`
-    - `slots` - a dictionary mapping slot names to :class:`~lamindb.Schema` objects, e.g., for an `AnnData`, `{"obs": Schema(...), "var.T": Schema(...), "obsm": Schema(...)}`
-    - `is_type=True` - a schema type, e.g., `ln.Schema(name="ProteinPanel", is_type=True)`
+        - `features` - a list of `Feature` objects
```
Zethson, line 188: Just for me to understand, why did you remove the sphinx ref here?

lamindb/models/schema.py
falexwolf: Thanks for the review! That's helpful!


This comment falls in our discussion about assert vs raising another error from about a year ago (I can find the link on Slack/GitHub, but don't have it at hand)

TLDR:

A ValueError or similar needs its dedicated test. The assert works without a test.

If it's likely that an error will be raised by a user, I'd say we should always craft a nice error UX and raise a typed error with a tested error message

I thought about it as an integrity check: it would be an edge case if a user tried to mutate an unnamed schema, because these are typically working in the background (aren't user defined in the first place, but implicitly inferred by lamindb during validation).

Writing this I realise that it's too easy for a user to accidentally create unnamed schemas... I guess one should actually require it so that they show up on the UI, and are queryable by name, and have auto-lookup 🤔

And writing this I realise that the way the implementation is done here works perfectly fine for unnamed schemas. I did a refactor half way that made this requirement unnecessary.

TODO: remove the assert and make it harder to accidentally make unnamed schemas

lamindb/models/schema.py
falexwolf: Sure, this will all be gone in lamindb 2.

lamindb/models/schema.py
falexwolf: The first reason is that Sphinx refs are helpful when they point to another page, bug a sphinx ref to the same page looks very verbose.

The second is that in several places in the docs, we use `Feature` stylized to mean the python class, e.g. in that line where `itype=ln.Feature` is explained. I find the contrast between the bold highlighted code-style `Feature` and the regular code-style `Feature` somewhat irritating.

My rule of thumb: I think simple styling should be the default so that the overall appearance is as homogeneous as possible and fat hyperlinks should be used when they reference another page for the first time in some context. ☺️

lamindb/models/schema.py
falexwolf: Done: https://github.com/laminlabs/lamindb/pull/3309/commits/5ce29ccb0ae97e93ec73f43285e0a0c918e0d4b8

With this, the unnamed schemas are also not an actual problem. So, things can stay as they are on that front.

PR #3310, opened 2025-11-23
👷 Also run core unit tests against an SQLite database

falexwolf:
While we ran almost all our guides on SQLite databases, we didn't run the more comprehensive unit tests on them.

This PR changes this and reveals bugs for the SQLite backend.

### DIFF
--- a/tests/core/test_schema.py
+++ b/tests/core/test_schema.py
@@ -385,17 +385,12 @@ def test_schema_components(mini_immuno_schema_flexible: ln.Schema):
     ).save()
     assert var_schema == var_schema2
 
-    try:
+    with pytest.raises(InvalidArgument) as error:
         ln.Schema(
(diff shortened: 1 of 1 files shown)


lamindb/models/_feature_manager.py
```diff
@@ -691,6 +722,14 @@ def filter_base(
                     f"currently not supporting `{comparator}`, using `__icontains` instead"
                 )
                 comparator = "__icontains"
+            if connections[feature._state.db].vendor == "sqlite" and comparator in {
```
Koncopd, line 732: `_state.db` can be `None`, no?

falexwolf: Not for something that's used in a query; all kinds of things will error then. But yes, it's legit that this whole thing should check that this `feature._state.adding is False`, which is what we already do in many places.

PR #3314, opened 2025-11-24
✨ Add `ln.DB` to conveniently query all entities in any instance

Zethson:
Added `QueryDB` class for querying remote LaminDB instances without switching the global connection context. Enables read-only access to multiple database instances simultaneously by providing a clean interface that bundles querysets for different registries.

Example:
```python
db = ln.DB("laminlabs/cellxgene")
db.Artifact.filter()
```

This feature was refined in the following PRs:

- https://github.com/laminlabs/lamindb/pull/3325
- https://github.com/laminlabs/lamindb/pull/3354
- https://github.com/laminlabs/lamindb/pull/3377

### DIFF
--- a/lamindb/models/query_set.py
+++ b/lamindb/models/query_set.py
@@ -4,8 +4,10 @@ import re
 from collections import UserList
 from collections.abc import Iterable
 from collections.abc import Iterable as IterableType
+from importlib import import_module
 from typing import TYPE_CHECKING, Any, Generic, NamedTuple, TypeVar
 
+import lamindb_setup as ln_setup
 import pandas as pd
 from django.core.exceptions import FieldError
 from django.db import models
@@ -1264,23 +1266,31 @@ class QueryDB:
         Returns:
             QuerySet for the specified registry scoped to this instance.
         """
-        from importlib import import_module
-
         class_name_base = "".join(word.capitalize() for word in name.split("_"))
         if class_name_base.endswith("s"):
             class_name_base = class_name_base[:-1]
 
-        for schema_name in ["lamindb"] + list(setup_settings.instance.modules):
+        owner, instance_name = self._instance.split("/")
+        instance_info = ln_setup._connect_instance._connect_instance(
+            owner=owner, name=instance_name
+        )
+        available_schemas = ["lamindb"] + list(instance_info.modules)
+
+        for schema_name in available_schemas:
             try:
                 schema_module = import_module(schema_name)
                 for attr in dir(schema_module.models):
                     if attr.lower() == class_name_base.lower():
                         model_class = getattr(schema_module.models, attr)
-                        return model_class.connect(self._instance)
+                        queryset = model_class.connect(self._instance)
+                        setattr(self, name, queryset)
+                        return queryset
             except (ImportError, AttributeError):
                 continue
 
-        raise AttributeError(f"Registry '{name}' not found in installed schemas")
+        raise AttributeError(
+            f"Registry '{name}' not found in installed modules for instance '{self._instance}'."
+        )
 
     def __repr__(self) -> str:
         return f"QueryDB('{self._instance}')"
--- a/tests/core/test_querydb.py
+++ b/tests/core/test_querydb.py
@@ -1,8 +1,9 @@
 import lamindb as ln
+import pytest
 
 
 def test_querydb_multiple_instances():
-    """Test accessing multiple instances simultaneously."""
+    """Accessing multiple instances simultaneously must work."""
     cellxgene = ln.QueryDB("laminlabs/cellxgene")
     lamindata = ln.QueryDB("laminlabs/lamindata")
     qs1 = cellxgene.artifacts.filter(suffix=".h5ad")
@@ -11,6 +12,19 @@ def test_querydb_multiple_instances():
 
 
 def test_querydb_bionty():
-    """Test querying a record that is not from LaminDB like bionty."""
+    """Querying a record from bionty must work."""
     cxg = ln.QueryDB("laminlabs/cellxgene")
     assert len(cxg.genes.filter(symbol__startswith="TP53")) > 0
+
+
+def test_querydb_missing_module():
+    """Attempting to access an attribute that comes from a missing module must error."""
+    db = ln.QueryDB("laminlabs/lamin-site-assets")  # instance without bionty
+
+    with pytest.raises(AttributeError) as e:
+        db.genes.first()
+
+    assert (
+        "Registry 'genes' not found in installed modules for instance 'laminlabs/lamin-site-assets'."
+        in str(e.value)
+    )


Zethson:
> Last comment: Is it intuitive to instantiate db = QueryDB(instance="laminlabs/cellxgene") as opposed to e.g. the weird alternative db = QueryDB.connect("laminlabs/cellxgene"). I mean clearly the latter is odd in many ways. And the former is a pattern that many libraries have. I'm just wondering about the absence of "connect". Maybe overthinking this.

I think that the former is fine but don't have a strong opinion on this. If you're unsure, we can ask the rest of the team.

falexwolf:
Darn. Now I have a lot of conflicts... 😆 

Just tried to push in the same moment.

falexwolf:
Given your commit is trivial, can I revert it and push my refactor?

Zethson:
Go for it. I'm off for tonight and fine with you finishing and merging it.

falexwolf:
Thanks!

falexwolf:
What we still have to somehow model test is the accessibility of `bionty` in case the default instance doesn't have bionty configured. I think we already get a nice error then.

Zethson:
Autocomplete doesn't work. I think we need to also overwrite `__dir__`

Zethson:
This seems to break `view_parents` which looks at the wrong instance:

```
effector_t_cell = cxg_db.cell_types.get("3nfZTVV4")
effector_t_cell.view_parents(distance=2, with_children=True)
```

crashes with

```
AttributeError                            Traceback (most recent call last)
Cell In[13], line 1
----> 1 effector_t_cell.view_parents(distance=2, with_children=True)

File ~/code/lamindb/lamindb/models/has_parents.py:114, in HasParents.view_parents(self, field, with_children, distance)
    111 if not isinstance(field, str):
    112     field = field.field.name
--> 114 return view_parents(
    115     record=self,  # type: ignore
    116     field=field,
    117     with_parents=True,
    118     with_children=with_children,
    119     distance=distance,
    120 )

File ~/code/lamindb/lamindb/models/has_parents.py:290, in view_parents(record, field, with_parents, with_children, distance, attr_name)
    282     df_edges_children = _df_edges_from_parents(
    283         record=record,
    284         field=field,
   (...)    287         attr_name=attr_name,
    288     )
    289     # Rename the columns to swap source and target
--> 290     df_edges_children = df_edges_children.rename(
    291         columns={
    292             "source": "temp_target",
    293             "source_label": "temp_target_label",
    294             "source_record": "temp_target_record",
    295             "target": "source",
    296             "target_label": "source_label",
    297             "target_record": "source_record",
    298         }
    299     )
    300     df_edges_children = df_edges_children.rename(
    301         columns={
    302             "temp_target": "target",
   (...)    305         }
    306     )
    307 if df_edges_parents is not None and df_edges_children is not None:

AttributeError: 'NoneType' object has no attribute 'rename'
```

falexwolf:
> This seems to break view_parents which looks at the wrong instance:

This is likely a problem with `Registry.connect()` that needs to be fixed in another PR, right?

falexwolf:
But we have static types, don't we? This shouldn't be necessary:

> Autocomplete doesn't work. I think we need to also overwrite __dir__

🤔

Zethson:
> But we have static types, don't we? This shouldn't be necessary:
> 
> > Autocomplete doesn't work. I think we need to also overwrite **dir**
> 
> 🤔

typehints aren't for runtime though. You can't tab complete typehints.

falexwolf:
> typehints aren't for runtime though. You can't tab complete typehints.

Apologies for being unclear. With "types" I meant we the static class attributes. The type _hints_ give context for auto-completion, but the auto-completion itself should be based on the class attributes, which I referred to as "types".

I'm referring to these attributes:

```python
class QueryDB:

    artifacts: QuerySet
    collections: QuerySet
    transforms: QuerySet
    runs: QuerySet
    users: QuerySet
    storages: QuerySet
    features: QuerySet
    ulabels: QuerySet
    records: QuerySet
    schemas: QuerySet
```

Why shouldn't they be auto-completable (irrespective of their type annotations).

Koncopd:
>This is likely a problem with Registry.connect() that needs to be fixed in another PR, right?

I am pretty sure the problem is not in `.connect()`, which just returns a `QuerySet`.

lamindb/models/query_set.py
```diff
@@ -1193,3 +1193,99 @@ def filter(self, *queries, **expressions) -> QuerySet:
...
+    transforms: QuerySet
+    runs: QuerySet
+    users: QuerySet
+    storages: QuerySet
+    features: QuerySet
+    ulabels: QuerySet
+    records: QuerySet
+    schemas: QuerySet
+
+    if setup_settings._instance_exists and "bionty" in setup_settings.instance.modules:
```
Koncopd, line 1226: should this check all modules actually?

falexwolf:
> I am pretty sure the problem is not in .connect(), which just returns a QuerySet.

Also sorry for being unclear here. I simply meant it's unrelated to this PR. It's an issue with evaluating a `view_parents()` after `.connect()` formerly known as `.using()`. It's likely an issue with propagating the connection through the `view_parents()` logic.

Zethson:
They're not full class attributes though. Just annotations.

```
print(hasattr(QueryDB, 'artifacts'))  # False
print('artifacts' in QueryDB.__annotations__)  # True
```

that's because they don't have a value so by default they only get added to annotations.

lamindb/models/query_set.py
Zethson: what do you mean? `setup_settings.instance.modules` is just the registered modules. This is not an expensive check.
Or are you saying that we should have a generalized version that does this for all modules and not just "hardcoded" for bionty and wetlab?

falexwolf:
> that's because they don't have a value so by default they only get added to annotations.

Right, we're not setting a value. Shouldn't we set a value during `__init__` and then this would be more idiomatic than overwriting `__dir__`?

falexwolf:
Wait, I just double checked. In VSCode auto-complete works.

image

So what is going wrong where in this reasoning chain?

Yesterday I was very surprised to read that:

> Autocomplete doesn't work. I think we need to also overwrite __dir__

Zethson:
It didn't work for me. Let me double check that.

Zethson:
image

I don't see anything without overwriting `__dir__`. The overwrite enables it for me.

falexwolf:
And what if you use my example? You also seem to be on VSCode whose auto-complete is based on the static type information.

In Jupyter, if I remember correctly, it might indeed only check `__dir__` and then render auto-complete from that, ignoring the static information.

In my memory, when studying the VSCode behavior 1 or 2 years ago, it ignored `__dir__`. This seems to have changed given your patch changes the behavior for you.

Zethson:
Your example works and I wonder whether this has something to do with where things are defined and imported. 

> In Jupyter, if I remember correctly, it might indeed only check __dir__ and then render auto-complete from that, ignoring the static information.

Without `__dir__` it also doesn't work for the Python REPL.

> Right, we're not setting a value. Shouldn't we set a value during __init__ and then this would be more idiomatic than overwriting __dir__?

`__dir__` is idiomatic to me but we can also do that

Zethson:
Actually, I went back to `__dir__` because setting it to a default value messes with the `get_attr` and increases code complexity then. We're using `__dir__` as intended and Claude also keeps suggesting to overwrite `__dir__`.

falexwolf:
Sounds good!

falexwolf:
Just to actually write it down -- it's been bugging me since we started the discussion about `QueryDB`.

Prisma offers these two patterns:

DB object `prisma` | `connect()` on table `user`
--- | ---
image | image

Query examples based on the DB object look like this:

image

This is 100% consistent with our API except for the plural `s` that we have on the attribute.

I think JS clients also use the singular convention and are similar. Postgrest would also use singular if the table name is singular (which many violate, but singular is considered idiomatic).

I think we _have to use_ plural on `artifact.records` etc. because otherwise people will think it's a `ForeignKey`.

So, I think we have to use plural also for these query sets even though it clashes with the nice singular convention.

The one way to justify singular is to consistently use singular:

```python
cellxgene = ln.QuerySet("laminlabs/cellxgene")
cellxgene.record.filter(...)
cellxgene_record = ln.Record.connect("laminlabs/cellxgene")
```

And do _not_ build on the analogy: `artifact.records`, `artifact.diseases`, etc.

The Prisma examples are from the Python client docs: https://prisma-client-py.readthedocs.io/en/stable/

falexwolf:
More perspectives on this.

In REST API design and UI design the plural is clearly favored, even though in SQL and ORM design the singular is idiomatic.

I don't think we can force users to use the plural vs. singular convention when making record types on the UI:

image

So the record types will often have a plural `s`, e.g. in the `Workflows` example. Sometimes they'll "behave like types" and sometimes "they'll behave like folders". And folders are also typically named with a plural `s`.

All of this makes me think that `QueryDB` should be thought of as a high-level interface in which the plural `s` is intuitive.

What do you think, @Zethson? Do you have an opinion on this?

Zethson:
We were already pretty close to the singular pattern in the first iteration of this PR where users were accessing the classes like:

```python
cxg = ln.QueryDB("laminlabs/cellxgene")
cxg.Artifact.filter()
```

which I found intuitive considering that people are used to a "module.Model.filter()" pattern. I'm not a fan of the whole singular vs plural handling that we now also need in `__get__attr__`.

I don't think that we can force users to do what we desire and we should always assume that they do the opposite. I'm in favor of ensuring that there's no "magic" here and what people see on the UI is also what they'd use in the API and vice versa.

falexwolf:
> I don't think that we can force users to do what we desire and we should always assume that they do the opposite. I'm in favor of ensuring that there's no "magic" here and what people see on the UI is also what they'd use in the API and vice versa.

Yes, absolutely.

> We were already pretty close to the singular pattern in the first iteration of this PR where users were accessing the classes like

Wasn't the initial design a no-go for these reasons:

(1) What about users misinterpreting the `Artifact` object as a class simply because it's capitalized?

(2) What about users trying to do things like this by analogy with the `Artifact` class:

```python
db.Artifact()
db.Artifact.load()
db.Artifact.cache()
db.Artifact.open()
```

(3) Is there any precedent for a class attribute that's capitalized that's not a class? pydantic, Django, and others have these constructs:

```python
class A:
    class B:
        pass
```

But clearly that's not what we have. We have a "bundle of query sets".

> which I found intuitive considering that people are used to a "module.Model.filter()" pattern. 

So, I disagree until you have a good response to the 3 issues. ☺️ 

> I'm not a fan of the whole singular vs plural handling that we now also need in __get__attr__.

Exactly. Doing this properly would require some more thinking. One could use the `ManyToMany` accessors that were the main source of the design inspiration: `artifact.diseases`, `artifact.ulabels`, etc. -- these all have the plural accessor well-defined, and not derived from some string casting. But this is more work and so maybe out-of-scope for this PR.

Zethson:
I kinda don't see users doing 2) because currently that would only work with Artifact objects and not with the Artifact class. No, there's not really a precedent for that and it'd also feel a bit off. I'm aware of that.

All right so summa summarum, I think the PR is ready to be merged right?

falexwolf:
I think so, yes! Considering all trade offs, I think this is a good quality of life improvement!

PR #3319, opened 2025-11-26
🐛 Fix `view_parents()` and `query_relatives()` for non-default instances

Zethson:
Discovered in https://github.com/laminlabs/lamindb/pull/3314#issuecomment-3575440309

Fixed database routing for records queried from remote instances via `connect()` . `view_parents()`, `query_parents()`, and `query_children()` now correctly query the same database as the record instead of defaulting to the local instance.

Before | After
--- | ---
Screenshot 2025-11-21 at 12 44 24 | Screenshot 2025-11-21 at 12 44 07


tests/core/test_has_parents.py
```diff
@@ -41,3 +42,31 @@ def test_view_lineage_circular():
...
+
+    if ct:
+        parents = ct.query_parents()
+        assert parents.db == "laminlabs/cellxgene"
+
+        children = ct.query_children()
+        assert children.db == "laminlabs/cellxgene"
+
+
+def test_view_lineage_connected_instance():
```
Zethson, line 65: This test already passed on main but it's still good to have it.

PR #3325, opened 2025-12-01
🚸 Polish `QueryDB`

Zethson:
- Refactored QueryDB to use uppercase singular registry names (e.g., cxg.Artifact instead of cxg.artifacts) for consistency with LaminDB conventions. 
- Added validation to prevent direct instantiation of registry classes from QueryDB, guiding users to use query methods like .filter() and .get() instead. 
- Added explicit type annotations for all core LaminDB, bionty, and wetlab registries to enable proper IDE autocomplete and Sphinx documentation generation.

### DIFF
--- a/lamindb/models/query_set.py
+++ b/lamindb/models/query_set.py
@@ -1269,9 +1269,21 @@ class QueryDB:
 
         Query records from a remote instance::
 
-            cellxgene = ln.QueryDB("laminlabs/cellxgene")
-            artifacts = cellxgene.Artifact.filter(suffix=".h5ad")
-            records = cellxgene.Record.filter(name__startswith="cell")
+            cxg_db = ln.QueryDB("laminlabs/cellxgene")
+            artifacts = cxg_db.Artifact.filter(suffix=".h5ad")
+            records = cxg_db.Record.filter(name__startswith="cell")
+
+            cxg_db.artifacts.filter(
+                suffix=".h5ad",
+                description__contains="immune",
+                size__gt=1e9,  # size > 1GB
+                cell_types__in=[
+                    cell_types.b_cell,
+                    cell_types.t_cell,
+                ],
+            ).order_by("created_at").to_dataframe(
+                include=["cell_types__name", "created_by__handle"]  # join with additional info
+            ).head()
     """
 
     Artifact: QuerySet[Artifact]  # type: ignore[type-arg]


falexwolf:
Ok, this is better than allowing it! Ultimately we need a proper lookup based on Python objects, not string transformation.

---

Very important: The `QueryDB` object should be a first class citizen. Can you please add it to the root level docs?

image

I'd say it should be all the way at the top?

image

Under "Query & connect to an instance"?

It should also be contrasted with `lamin connect`.

The `ln.connect()` might become more prominent in the future but for now should remain under "Basic utilities".

---

The docs for QueryDB should clearly show the `.artifacts`, `.records`, etc. attributes similar to how we have them here:

image

This is the analogy that users should have in mind.

falexwolf:
I think the following insight holds true:

While the M2M fields like `Artifact.records` should have the `s` and can be customized on a per-entity basis (e.g. `.linked_records` vs. `.records`) and needs to be contrasted with the singular foreign key, this argument clearly doesn't hold true for `QueryDB`.

`QueryDB` exposes a read-only interface to an entity. And that means lowercase-singular as in Prisma would, a priori, make the most sense. The analogy with M2M fields like `Artifact.records` is misleading.

We could still say the `QueryDB` experience should build on parallelism with the M2M fields, as originally argued. But realizing the above more clearly I'd say the singular is as good or might be better.

Unfortunately we need to get this right the first time, before we make the 1.17 release.

Zethson:
> QueryDB exposes a read-only interface to an entity. And that means lowercase-singular as in Prisma would, a priori, make the most sense. The analogy with M2M fields like Artifact.records is misleading.

I read what you wrote but I feel like lowercase singular is yet another pattern then. While I found it intuitive to have `querydb.Artifact` because of `ln.Artifact`, I don't quite see where else we have the pattern `something.artifact`. Therefore, I'd like to avoid that.

falexwolf:
Right. And I forgot that there might be `module1.Experiment` and `module2.Experiment` and then the name should be `module1_experiments` and `module2_experiments` just like one would do it for `artifact.module1_experiments` and `artifact.module2_experiments`.

(We did this for `artifact.clinicore_biosamples` and `artifact.wetlab_biosamples`.)

So, the plural inferred from the `related_name` on `Artifact` as the guiding rule, and hopefully at some point programmatic, is the best convention.

Unless we introduce a namspace along the lines of `db.wetlab.biosample` and `db.clinicore.biosample`, which is how Robrecht did it in the first LaminR `DBClient` design. 🤔 

Hmmmmm. These are unfortunately very hard design questions that we have to nail to avoid things feeling clunky.

falexwolf:
Maybe to define what we know: We cannot keep casting strings, that'd be much too unreliable, right? It works for a prototype but this can't be the production code long term. We are agreed on this, right?

Hence: Can we think through how the accessor name would be obtained programmatically? Would we loop through the attributes of `Artifact`?

---

> I don't quite see where else we have the pattern something.artifact

Yes, we don't have it. But many other ORMs and SQL have it. I'm still not sure. This can't be a hand-waving decision.

`QueryDB` will be the central way to query another instance. It has to be really pretty and it has to make sense. 

Sorry. As usual we didn't expect this complexity. This was meant to be simple and it turns out the implementation is simple but the design questions are hard.

Zethson:
> Maybe to define what we know: We cannot keep casting strings, that'd be much too unreliable, right? It works for a prototype but this can't be the production code long term. We are agreed on this, right?

Yes, of course. Always terrible to work with raw string representations.

> Hence: Can we think through how the accessor name would be obtained programmatically? Would we loop through the attributes of Artifact?

Reporting back very soon.

Zethson:
@falexwolf this PR now includes a prototype of automatic discovery of the registries so that we don't have to hardcode them anymore. However, this does not give us the docs that you're imagining. I think that we can only get those if we dynamically build the attributes section based on the `__doc__` of the attributes. While likely possible, this whole thing is exploding a bit in complexity. I can probably complete this tomorrow but I wonder whether the price that we're paying is getting a bit high.

(ignore the failing test - trivial to add back)

falexwolf:
Thanks! I'll take an in depth look tomorrow morning.

lamindb/__init__.py
```diff
@@ -1,5 +1,15 @@
 """A data framework for biology.
 
+Query & connect to an instance
+==============================
```
falexwolf, line 4: Great!

I'd just also add the `lamin connect` command with a sentence that says that this configuring an environment for writes & tracking.

And `QueryDB` is a pure tool to query instances.

lamindb/__init__.py
```diff
@@ -40,16 +50,15 @@
    Transform
    Run
 
-Records, labels, features & schemas
```
falexwolf, line 43: This was a conscious change 2 days ago! ☺️

falexwolf: Background: https://laminlabs.slack.com/archives/C04FPE8V01W/p1764106618090609?thread_ts=1746869196.094099&cid=C04FPE8V01W

falexwolf:
> However, this does not give us the docs that you're imagining.

I think that should just be a tweak in `lndocs`. If there are attributes, they should be discovered and rendered.

It should be easy to include them here if for whatever reason they're not included. https://github.com/laminlabs/lndocs/blob/a2d2b76e3a35c824e390a8dbf7e0984d12f7ebee/lndocs/lamin_sphinx/__init__.py#L838

And then we just need the registry as the type annotation and it'll be beautiful!

> this PR now includes a prototype of automatic discovery of the registries so that we don't have to hardcode them anymore

I like this!

> Unless we introduce a namespace along the lines of db.wetlab.biosample and db.clinicore.biosample, which is how Robrecht did it in the first LaminR DBClient design. 🤔

I spent the night going back and forth on this weighing the pros & cons that we discussed up there and in the previous PR.

I'm pretty convinced that singular and lower-case is the only canonical design among the three designs we considered. 

I think the analogy with the M2M fields is inaccurate and ultimately bites us. Unlike the plural, the singular is 100% well defined because it maps exactly on the SQL table name. The plural leads to weirdnesses with customized pluralizations (which you already have in the code now e.g. for ethnicities), which will guaranteed lead to bugs in custom modules with custom registries. There is no 100% reliable rule for pluralization.

In a nutshell: The pluralized M2M field does _not_ represent an entity, it represents a _relationship_. But we need an _entity_ in a database.

With the module namespace the way Robrecht had it in LaminR things are 100% analogous to working with the default instance.

I also now think that having _just a difference_ in capitalization (and _not also_ in pluralization) is a clear way to distinguish read vs. write access to _an entity_. It's clear that for read access no objects can be instantiated:

```python
ln.Artifact  # can be instantiated -> write access
db.artifact  # cannot be instantiated -> read access
```

For reads, things are parallel. Both of these read natural, and indeed more natural than if one of them was pluralized:

```python
ln.Artifact.filter()
db.artifact.filter()
ln.Artifact.get()
db.artifact.get()
```

The only downside of the singular is that it's a new pattern. But `QueryDB` is itself a new pattern and we're not becoming _more complicated_ than e.g. Prisma with this pattern (and Prisma has great design/UX). In fact, we're pretty much analogous then, which is a good thing.

Happy to dicuss on a call!

lamindb/__init__.py
Zethson: I also noticed that and it's a merge error. Not sure how I ended up there but I'll ensure that there are no changes.

Zethson:
> I think that should just be a tweak in lndocs. If there are attributes, they should be discovered and rendered.
> 
> It should be easy to include them here if for whatever reason they're not included. https://github.com/laminlabs/lndocs/blob/a2d2b76e3a35c824e390a8dbf7e0984d12f7ebee/lndocs/lamin_sphinx/__init__.py#L838
> 
> And then we just need the registry as the type annotation and it'll be beautiful!

I don't think that's possible. The attributes don't exist at Sphinx build time. They're only generated when an object was created and the user tries to access them. That's the cost of not hardcoding them.

> The plural leads to weirdnesses with customized pluralizations (which you already have in the code now e.g. for ethnicities), which will guaranteed lead to bugs in custom modules with custom registries. There is no 100% reliable rule for pluralization.

There's external dependencies for reliable plurals but obviously we don't want to introduce them just for that. But 1000%, the current plural handling is terrible. 

> Unlike the plural, the singular is 100% well defined because it maps exactly on the SQL table name.

Yes.

> I also now think that having just a difference in capitalization (and not also in pluralization) is a clear way to distinguish read vs. write access to an entity. It's clear that for read access no objects can be instantiated:

It is a clear way but is it a clear way for new users? Imagine you were to look at our docs for the first time. You might be confused why suddenly the same thing is now lowercase. For an experienced user, this correct distinguishment might make sense but I'm a bit worried about "first impressions".

I suggest that we quickly ask the DB team for feedback and then I implement it. Any singular implementation also has less complexity which is great.

falexwolf:
~I'll wait for the docs to build~ ☺️ 

Really curious how it's going to look

falexwolf (approved):
Pending the docs look pretty to you, I'd say: go and merge!

Code looks good to me! Thank you!

falexwolf:
Ok, seeing!

image

Can we have the `Attributes` heading that we have in the other class docs? It's likely stripped because `lndocs` doesn't add it for very simple classes but here we need it. 

So, we need an exception for `QueryDB` to get this.

(The good rule would be to only print this header on one-page-per-class pages, but I couldn't figure this out yet.)

falexwolf:
Once we have the "Attributes" section also "Examples" can shine. ☺️ 

I think it'd be great if you just copy & pasted a few queries from the `cxg` guide. So that people see that `.filter()` can be chained etc.

lamindb/models/query_set.py
```diff
@@ -1197,94 +1242,160 @@ def filter(self, *queries, **expressions) -> QuerySet:
...
+            f"Use {self._name}.filter(), {self._name}.get(), etc. to query records."
+        )
+
+    def __getattr__(self, attr):
+        return getattr(self._qs, attr)
+
+
 class QueryDB:
-    """Convenient access to QuerySets for every entity in a LaminDB instance.
+    """Convenient queries for all registries of an instance.
```
falexwolf: How about: "Query any registry of any instance."

Or: "Query any registry of any LaminDB instance."

I think the "convenient" doesn't add much: any abstraction should be "convenient".

I think saying "Queries" instead of "Query" is probably also fine, but verbal style would be better because this class isn't a registry to store things, this is a class that allows you to do something right away. We typically use verbal style for such logic.

Zethson: > I think the "convenient" doesn't add much: any abstraction should be "convenient".

True, these "filler" words are noise. Thanks.

> I think saying "Queries" instead of "Query" is probably also fine, but verbal style would be better because this class isn't a registry to store things, this is a class that allows you to do something right away. We typically use verbal style for such logic.

I also agree here, thanks.

> Query any registry of any instance.

Went with this.

PR #3339, opened 2025-12-04
✨ Test all clones work

Zethson:
Might be overkill - I'll think about it.

### DIFF
--- a/tests/core/test_queryset.py
+++ b/tests/core/test_queryset.py
@@ -358,7 +358,7 @@ def test_connect_public_clone_instance():
         pass
 
 
-def test_clones_work():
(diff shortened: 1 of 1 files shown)


falexwolf:
I think this is good to test as part of integration tests prior to releasing. But not as part of the unit tests. This will run ages and clearly doesn't test logic inside lamindb but factors related to the maintenance of these instances.

In that sense they're also unrelated to releasing lamindb. A cron job could be the right thing to do. Maybe as part of `lamindb-dispatch` given we'll have this problem regularly that we want to ascertain the integrity of certain assets.

Zethson:
Yeah, that's also been my thinking.

PR #3346, opened 2025-12-05
🔧 Default organism should be `None`, not `"human"`

falexwolf:
Superseded by:

- https://github.com/laminlabs/lamindb/pull/3350

Imposing a global organism can lead to unexpected side effects.

Instead, by default, no organism should be set.

See:

- https://github.com/laminlabs/bionty/pull/335


Koncopd:
`django.db.utils.ProgrammingError: relation "public.hubmodule_dbwritelog" does not exist`
You ran that exactly when i was reinstalling `dbwritelog`.

It is actually annoying that this can break all write operations in an instance for some time.

PR #3350, opened 2025-12-07
🚸 Allow validation of genes to pass without indicating organism

sunnyosun:
This is a temporary solution that allows non-human genes to pass validation when an organism isn't passed, if they are already in the registry. A refactor that upstreams this behavior will come.

### DIFF
--- a/lamindb/curators/core.py
+++ b/lamindb/curators/core.py
@@ -1478,8 +1478,8 @@ class CatVector:
                 values_array, field=self._field, **filter_kwargs, mute=True
             )
             validated_labels, non_validated_labels = (
-                values_array[validated_mask],
-                values_array[~validated_mask],
+                values_array[validated_mask].tolist(),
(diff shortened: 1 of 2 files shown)


lamindb/curators/core.py
```diff
@@ -1582,9 +1596,22 @@ def _validate(
...
+
+        values_validated: list[str] = []
+
+        # first validate against the registry with perfect matches
+        if (
+            hasattr(registry, "organism")
+            and valid_inspect_kwargs.get("organism") is None
+            and valid_inspect_kwargs.get("source") is None
+        ):
+            values_validated = registry_or_queryset.filter(
```
falexwolf, line 1608: Is this how validation is also done within `_inspect()`?

sunnyosun: No, that will be a bigger refactor to change inspect as inspect is case insensitive!

PR #3354, opened 2025-12-08
✨ QueryDB accessor & documentation improvements

Zethson:
Add schema namespaces to `QueryDB` for cleaner module organization.

Registries from `bionty` and `wetlab` modules now require namespace access (e.g., `db.bt.Gene` instead of `db.Gene`). Core `lamindb` registries remain at top level (e.g., `db.Artifact`). Both full (`db.bionty`) and short (`db.bt`) namespace names are supported.

### DIFF
--- a/lamindb/models/query_set.py
+++ b/lamindb/models/query_set.py
@@ -1475,6 +1475,8 @@ class WetlabQueryDB(ModuleNamespace):
 class QueryDB:
     """Query any registry of any instance.
 
+    QueryDB exposes all available registries of LaminDB and modules like Bionty or Wetlab.
+
     Args:
         instance: Instance identifier in format "account/instance".
 
@@ -1486,6 +1488,8 @@ class QueryDB:
 
             artifacts = cxg.Artifact.filter(suffix=".h5ad")
             records = cxg.Record.filter(name__startswith="cell")
+
+            ECL = cxg.bionty.CellType.filter(name="enterochromaffin-like cell")
 
             cxg.Artifact.filter(
                 suffix=".h5ad",


lamindb/models/query_set.py
```diff
@@ -1364,6 +1363,98 @@ def __getattr__(self, attr):
...
+    def __dir__(self) -> list[str]:
+        """Return list of available registries in this schema module."""
+        base_attrs = [attr for attr in object.__dir__(self) if not attr.startswith("_")]
+        try:
+            schema_module = import_module(self._schema_name)
+            if hasattr(schema_module, "__all__"):
+                registries = set()
+                for class_name in schema_module.__all__:
+                    model_class = getattr(schema_module, class_name, None)
+                    if model_class and hasattr(model_class, "connect"):
```
falexwolf, line 1432: How about `issubclass(SQLRecord)` instead of `hasattr(model_class, "connect')`?

tests/core/test_querydb.py
```diff
@@ -4,63 +4,74 @@
...
+    cxg_db = ln.QueryDB("laminlabs/cellxgene")
     with pytest.raises(AttributeError) as e:
-        cxg.artifacts  # noqa: B018
-    assert "Registry names must be capitalized and singular." in str(e.value)
+        getattr(cxg_db, attr)
+    assert expected_msg in str(e.value)
 
 
 def test_querydb_cache():
     """Subsequent accesses must return cached wrapper."""
```
falexwolf, line 63: Can you add one test that asserts that the queryset caching works as expected?

Say you query `Artifact.filter()` once and then another time; the two results should be different.

I'm pretty sure what returns is a copy of a QuerySet but better double check with a small test.

falexwolf:
Docs example:

image

The "remote" word is unnecessary. The database instance could be anywhere.

I'd not use the `cxg_db` but just say `db` or `cxg` just because less complexity.

How about adding a simple query, too? Like match artifact on a key or something like this?

falexwolf:
Why don't these render? It'd be good if they did.

image

Also a note that QueryDB exposes the same registries as the modules (lamindb, bionty, wetlab) would be nice.

falexwolf:
All attributes show up twice! ☺️ 

image

lamindb/models/query_set.py
```diff
@@ -1364,6 +1363,98 @@ def __getattr__(self, attr):
         return getattr(self._qs, attr)
 
 
+class SchemaNamespace:
```
falexwolf: I'd call this `ModuleNamespace` because we say "module", not "schema". ☺️

lamindb/models/query_set.py
```diff
@@ -1364,6 +1363,98 @@ def __getattr__(self, attr):
...
+    """Namespace for accessing registries from a specific schema module.
+
+    Args:
+        query_db: Parent QueryDB instance.
+        schema_name: Name of the schema module (e.g., 'bionty', 'wetlab').
```
falexwolf: This should be "module_name".

falexwolf (approved):
This looks great! These are all just small cosmetic changes that I'm requesting.

falexwolf:
Now I see why docs for `QueryDB` were duplicated. They're manually listed twice, once at the root and once here:

image

And I'm confused what "black magic" was done to eliminate the duplication warnings. They should still appear given it's referenced twice. 🤔

PR #3387, opened 2025-12-15
📝 README v5.2

falexwolf:
Refines:

- https://github.com/laminlabs/lamindb/pull/3372


falexwolf:
Writing this I feel like we need to work much harder to justify this inconsistency in the return types:

image

Or not use `__getitem__`. 🤔 

Did we document alternative designs for this?

---

I think it can be justified though. If we think of the sheets UI or any app for a relational database, the value that is shown is the _record_ not just an identifier for it. So, the _record_ is the _actual_ value and if we write an identifier in a csv or a parquet file then it's just a proxy.

So, if you will, `get_values()` might be making the "wrong" choice; but of course much logic & UX relies on these string values.

PR #3389, opened 2025-12-15
[Superseded] Better re-use Django's `ObjectDoesNotExist` and `MultipleObjectsReturned`

falexwolf:
Superseded by:

- https://github.com/laminlabs/lamindb/pull/3396

### DIFF
--- a/lamindb/models/feature.py
+++ b/lamindb/models/feature.py
@@ -32,7 +32,6 @@ from lamindb.base.fields import (
 )
 from lamindb.base.types import Dtype, FieldAttr
 from lamindb.errors import (
-    DoesNotExist,
     FieldValidationError,
     IntegrityError,
     InvalidArgument,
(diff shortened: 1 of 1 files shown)


falexwolf:
@sunnyosun, do you have any idea why the refactor here would impact these two tests?

image

sunnyosun:
Because they are redefined here from `SystemExit`? https://github.com/laminlabs/bionty/blob/main/bionty/_organism.py#L11

sunnyosun:
This is the only place that manually throws DoNotExist in `from_source`: https://github.com/laminlabs/bionty/blob/main/bionty/models.py#L625

sunnyosun:
This test not raising error was due to: https://github.com/laminlabs/lamindb/blob/cdbf0e819407dd6a54322c4780a861125883f57a/lamindb/models/feature.py#L540-L541

So in fact it makes sense this test not raising any errors.

falexwolf:
Are you saying this should not raise an error and the test was wrong all along? 🤔

falexwolf:
I'd say this should raise an error rather than excepting the error. So, I made the code more stringent so that it actually errors.

This is for the complex case in which a filter condition is specified within a dtype.

PR #3405, opened 2025-12-24
🔥 Remove tracking from `collection.artifacts.all()`

Koncopd:
Calling `collection.artifacts.all()` doesn't trigger tracking the collection anymore.

### DIFF
--- a/lamindb/models/query_manager.py
+++ b/lamindb/models/query_manager.py
@@ -23,7 +23,6 @@ from django.db.models.lookups import (
     Regex,
     StartsWith,
 )
-from lamin_utils import logger
 from lamin_utils._lookup import Lookup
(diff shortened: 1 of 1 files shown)


lamindb/models/query_manager.py
```diff
@@ -298,7 +298,7 @@ def to_dataframe(self, **kwargs):
 
         For `**kwargs`, see :meth:`lamindb.models.QuerySet.to_dataframe`.
         """
-        return self.all().to_dataframe(**kwargs)
+        return super().all().to_dataframe(**kwargs)
```
Koncopd: @falexwolf i think `collection.artifacts.to_dataframe()` should not trigger tracking.

falexwolf: Ok!

Issue #2459, opened 2025-02-14
Error when calling class methods on a record


sunnyosun:
### Add a description

At the moment, some users are confused about why `ln.Artifact.get(...).df()` returns all the records without filtering.

We need to error in such cases so that class methods can no longer be called on a record.

falexwolf:
We now have a recipe for this in `Record.to_dataframe()`.

falexwolf:
I found that we don't need to do anything for class methods that work through a metaclass:

```python
>>> ln.ULabel.to_dataframe()
         uid            name  is_type  \
id                                      
4   JJ3d8a2v              na    False   
3   hCKrqeRC  SuspensionType     True   
2   pscLWZbk    BTBR-ob/ob-1    False   
1   EhmpAjJO         DonorId     True  
>>> label = ln.ULabel.get(name="na")
>>> label.to_dataframe()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'ULabel' object has no attribute 'to_dataframe'
```

We already have the desirable behavior. ☺️ 

We only need to deal with class methods that act through `@classmethod`.

falexwolf:
Resolved via:

- https://github.com/laminlabs/lamindb/pull/3411

PR #3411, opened 2026-01-01
🚸 Raise a clear error if calling `CanCurate` and similar registry class methods on objects

falexwolf:
```python
def test_validate_called_on_object_raises_error():
    """Test that calling validate() on an object raises TypeError."""
    label = ln.ULabel(name="test_label").save()
    with pytest.raises(TypeError) as error:
        label.validate(["test_value"])
    assert (
        "ULabel.validate() is a class method and must be called on the ULabel class, not on a ULabel object"
        in str(error.value)
    )
```

Needs:

- https://github.com/laminlabs/bionty/pull/340
- https://github.com/laminlabs/lndocs/pull/145

Addresses:

- https://github.com/laminlabs/lamindb/issues/2459
- https://github.com/laminlabs/pfizer-lamin-usage/issues/441

### DIFF
--- a/lamindb/base/utils.py
+++ b/lamindb/base/utils.py
@@ -38,14 +38,12 @@ class strict_classmethod:
 
     def __get__(self, instance, owner):
         if instance is not None:
-            # Called on an instance - raise immediately or return cached error raiser
+            # Called on an instance - raise immediately
             raise TypeError(
                 f"{owner.__name__}.{self.func.__name__}() is a class method and must be called on the {owner.__name__} class, not on a {owner.__name__} object"
             )
 
         # Called on the class - return bound method using MethodType
-        if owner is None:
-            owner = type(instance)
         return MethodType(self.func, owner)
 
 


falexwolf:
Locally `TypeError` is raised correctly. But it's not raised in the first CI run. 🤔

lamindb/base/utils.py
```diff
@@ -36,39 +36,13 @@ def wrapper(*args, **kwargs):
...
-        # Called on the class - return the bound classmethod
-        print("  -> Returning bound classmethod for class call")
-        return self.func.__get__(obj, objtype)
-
-    def __call__(self, *args, **kwargs):
-        print(f"DEBUG __call__ invoked for {self.func.__name__}")
-        return self.func(*args, **kwargs)
+def raise_error_if_called_on_object(cls, method_name: str):
+    """Raise an error if a classmethod is called on an object."""
+    if hasattr(cls, "_state"):
```
Koncopd: This is very ad hoc approach. Can it be made more general?

Koncopd: Like working not only for django objects.

Koncopd: Or at least it should be named `raise_error_if_called_on_django_object` or something like that.

Koncopd: strange, your initial general approach should work https://stackoverflow.com/questions/78928486/how-to-deduce-whether-a-classmethod-is-called-on-an-instance-or-on-a-class

Koncopd: I would rather check why the original approach fails here. Maybe a discrepancy between python versions?

Koncopd: ok, it looks like i get now why the original method didin't work from reading the comments in the stackoverflow question.

Koncopd: btw did you get the code from llm?

falexwolf: It's specifically targeted for registries. But I agree this should be clearer.

falexwolf: I'll refactor.

Koncopd: I think it is better to figure out why the descriptor approach fails here. It looks like the bets way to solve this.

Koncopd: Something probbaly overwrites `__get__` at some point.

falexwolf:
After some back and forth with Claude it seems that all ways that rely on Python's builtin `@classmethod` are hacks. The only reliable way to achieve a "strict classmethod" is to work through a custom decorator.

Given we already have the custom `class_and_instance_method` decorator, I think this is worthwhile.

Refs:

* Chat: https://claude.ai/chat/7da98481-c597-433b-ab39-85b5cb5fb20d
* Share: https://claude.ai/share/a6ef2dc1-860f-48fa-ba46-851343a3056c

lamindb/base/utils.py
falexwolf: Sorry, GitHub didn't display the bulk of your comments for me. Responding now.

falexwolf:
> I think it is better to figure out why the descriptor approach fails here. It looks like the bets way to solve this.

The main reason I dropped the "double descriptor approach" is that the code for it became very complicated, in particular if also addressing `__call()__` which was the reason for the test failure. It also had considerable import-time overhead.

lamindb/base/utils.py
falexwolf: Responded in the main thread because we're discussing the general architecture of how to achieve this.

lamindb/base/utils.py
```diff
@@ -35,4 +36,34 @@ def wrapper(*args, **kwargs):
...
+            def error_raiser(*args, **kwargs):
+                raise TypeError(
+                    f"{owner.__name__}.{self.func.__name__}() is a class method and must be called on the {owner.__name__} class, not on a {owner.__name__} object"
+                )
+
+            return error_raiser
+
+        # Called on the class - return bound method
+        @wraps(self.func)
+        def wrapper(*args, **kwargs):
```
Koncopd: this looks suspicious to me, it creates a wrapper dynamically on every invocation of `__get__`, this is not good.

btw this is how `classmethod` looks like in python per https://docs.python.org/3/howto/descriptor.html:
```python
import functools

class ClassMethod:
    "Emulate PyClassMethod_Type() in Objects/funcobject.c"

    def __init__(self, f):
        self.f = f
        functools.update_wrapper(self, f)

    def __get__(self, obj, cls=None):
        if cls is None:
            cls = type(obj)
        return MethodType(self.f, cls)
```

Koncopd: i would rather just copy it.

falexwolf: Thanks for the comment! I think this should be addressable.

Koncopd:
Yeah, i agree `strict_classmethod` approach is better than adding a descriptor on top.

lamindb/base/utils.py
falexwolf: I'm pushing a fix that addresses your important concern in a minute.

falexwolf:
I'll also resolve the same inefficiency for `class_and_instance_method()` in the next commit.

lamindb/base/utils.py
```diff
@@ -16,23 +18,40 @@ class class_and_instance_method:
...
-__all__ = ["doc_args", "deprecated", "class_and_instance_method"]
+    def __get__(self, instance, owner):
+        if instance is not None:
+            # Called on an instance - raise immediately or return cached error raiser
+            raise TypeError(
+                f"{owner.__name__}.{self.func.__name__}() is a class method and must be called on the {owner.__name__} class, not on a {owner.__name__} object"
+            )
+
+        # Called on the class - return bound method using MethodType
+        if owner is None:
```
Koncopd: i think it is not needed here, i assume this check is when the class method is called on the instance. We will have `instance = None` always.

falexwolf:
Thanks, @Koncopd! Your comment about inefficiency was very important!

lamindb/base/utils.py
falexwolf: You're right. Then the construction would also be parallel to the updated `instance_and_class_method()` decorator.

falexwolf:
> i think it is not needed here, i assume this check is when the class method is called on the instance. We will have instance = None always.

You're much more attentive than Claude, Sergei! 😄 

Also addressed the second comment: https://github.com/laminlabs/lamindb/pull/3411/commits/b311c3ecee650d9f7b24a36bd0b57f76f8859217

lamindb/base/utils.py
falexwolf: Addressed here: https://github.com/laminlabs/lamindb/pull/3411/commits/b311c3ecee650d9f7b24a36bd0b57f76f8859217

Koncopd:
Yeah, i think we always have to review llm-generated code very carefully. At least for now.

falexwolf:
Just so that all documentation is in one place: I found that we don't need to do anything for class methods that work through a metaclass:

```python
>>> ln.ULabel.to_dataframe()
         uid            name  is_type  \
id                                      
4   JJ3d8a2v              na    False   
3   hCKrqeRC  SuspensionType     True   
2   pscLWZbk    BTBR-ob/ob-1    False   
1   EhmpAjJO         DonorId     True  
>>> label = ln.ULabel.get(name="na")
>>> label.to_dataframe()
AttributeError: 'ULabel' object has no attribute 'to_dataframe'
```

We already have the desirable behavior. ☺️ 

We only need to deal with class methods that act through `@classmethod`. This is what this PR achieved.

Issue #2736, opened 2025-05-02
📝 Clickable return types in `Artifact.open()`


Koncopd:
for proper references of return types in docs

It is not pretty now, see for example the return types here https://docs.lamin.ai/lamindb.artifact#lamindb.Artifact.open

Zethson:
Intersphinx might be sufficient?
Alex said that intersphinx slowed down docs building but I have not observed that in my packages. We can cache the intersphinx downloads.

falexwolf:
We've been using intersphinx since day 1 just like for scanpy/anndata and yes, it's the right way of making these types appear:

<img width="684" height="94" alt="Image" src="https://github.com/user-attachments/assets/30d82660-6ae4-4141-9fcd-8ff1673a8dcd" />

> Alex said that intersphinx slowed down docs building but I have not observed that in my packages. We can cache the intersphinx downloads.

Indeed I observed this for very long lists of cross-linked projects, which is why we don't use it for `pandas` and `anndata`. But `Artifact.open()` is a clear case where the value add justifies adding 2 more intersphinx projects.

falexwolf:
I'm looking into this.

falexwolf:
I think our problem is that in addition to using intersphinx, we also use mock imports. If I just add e.g. polars and tiledbsoma to intersphinx the types still don't render.

I added the following two two lines to the linked section in `lndocs` and built docs locally:

```python
intersphinx_mapping = {
    "docs": ("https://docs.lamin.ai", None),
    "polars": ("https://docs.pola.rs/api/python/stable", None),
    "tiledbsoma": ("https://docs.tiledb.com/api/python/stable", None),
}
```

https://github.com/laminlabs/lndocs/blob/100bb7a72ac7055f4e70c3c2c29dc453ffa985f9/lndocs/lamin_sphinx/__init__.py#L106-L128

falexwolf:
It not working is unrelated to mock imports. Investigating.

- https://github.com/laminlabs/lndocs/pull/146

falexwolf:
It evidently works through a regular link:

<img width="785" height="638" alt="Image" src="https://github.com/user-attachments/assets/3cd91d0e-36d8-402a-8bb2-661d1be9c00b" />

It also seems unrelated to `TYPE_CHECKING`.

I'm halfway working through a list of explanations from Claude

- Chat: https://claude.ai/chat/e302421d-a512-436d-95be-760b991f0685
[...]

PR #3414, opened 2026-01-01
📝 Render return type annotations of `Artifact.open()` in docs

falexwolf:
Before | After
--- | ---
image | image

Needs:

- https://github.com/laminlabs/lndocs/pull/146

Addresses:

- https://github.com/laminlabs/lamindb/issues/2736

### DIFF
--- a/lamindb/models/artifact.py
+++ b/lamindb/models/artifact.py
@@ -2481,10 +2481,13 @@ class Artifact(SQLRecord, IsVersioned, TracksRun, TracksUpdates):
     ):
         """Open a dataset for streaming.
 
-        Works for `pyarrow` or `polars` compatible formats
-        (`.parquet`, `.csv`, `.ipc` etc. files or directories with such files),
-        `AnnData` (`.h5ad`, `.zarr`), `SpatialData` (`.zarr`),
-        generic `hdf5` and `zarr`, `tiledbsoma` objects (`.tiledbsoma`).
+        Works for the following object types (storage formats):
+
+        - `DataFrame` (`.parquet`, `.csv`, `.ipc` files or directories with such files)
+        - `AnnData` (`.h5ad`, `.zarr`)
+        - `SpatialData` (`.zarr`)
+        - `tiledbsoma` (`.tiledbsoma`)
+        - generic arrays (`.h5`, `.zarr`)
 
         Args:
             mode: can be `"r"` or `"w"` (write mode) for `tiledbsoma` stores,


lamindb/models/artifact.py
```diff
@@ -2467,14 +2467,14 @@ def open(
...
+        PyArrowDataset
+        | PolarsLazyFrame
+        | AnnDataAccessor
         | SpatialDataAccessor
         | BackedAccessor
         | SOMACollection
         | SOMAExperiment
         | SOMAMeasurement
-        | PyArrowDataset
-        | Iterator[PolarsLazyFrame]
```
falexwolf, line 2477: @Koncopd, the `Iterateor` here was a mistake if I look into how it's documented in the `arrays` guide and the polars docs?

falexwolf:
Docs render fine for me locally but not yet on CI. 🤔 

The missing dependencies are installed, the PR on `lndocs` got merged. Hm.

image

Trying another time.

lamindb/models/artifact.py
Koncopd: No, this is not a mistake, why? This indicates that the return type is a context manager giving polars lazyframe.

Koncopd: Could you restore please?

falexwolf: > No, this is not a mistake, why?

I've never seen context managers documented like this, but of course that might be my ignorance. If I look at our docs, I'd expect a return type `LazyFrame` for this pattern:

image

I find the `Iterator` confusing because it doesn't remind me of a context manager; I was close to confusing it with `Iterable`.

> Could you restore please?

To be consistent with the remaining docs, we should have the `DataFrame` types at the top. Of course we should restore `Iterator` if this is how other libraries are documenting return types of context managers. And then we add a comment because I'm sure not only I am confused about this! ☺️

falexwolf:
What I dislike about the new design is that `tiledbsoma.Collection` et al. are rendered without their `SOMA` prefixes. This can be confusing.

I'm almost tempted to go back to `autodoc_mock_import` for `tiledbsoma` for this reason. But the problem is that this breaks integrity of the remaining return type annotations, too.

lamindb/models/artifact.py
falexwolf: I also find it odd that we don't document `AnnDataAccessor` this way then, because the recommended way for it is via a context manager, too, right?

falexwolf:
As soon as I add `tiledbsoma` to mock imports it starts to look like this again:

image

Maybe this is clearer though?

lamindb/models/artifact.py
Koncopd: `AnnDataAccessor` is a context manager by itself, unlike the return value over the lazyframe. In any case just lazyframe is not a correct return type here.

falexwolf: Do you have any sort of reference to show how it's done elsewhere?

> AnnDataAccessor is a context manager by itself, unlike the return value over the lazyframe.

What does this mean? 😅 I find this pretty complicated and confusing and I'm pretty sure users will, too.

When I prompt Claude, it answers this:

image

* Chat: https://claude.ai/chat/a3fc81f2-e561-4f30-b52b-2584817b7e57
* Share: https://claude.ai/share/581a50a8-587d-42ef-bf6c-df4bdcf0e2c3

falexwolf:
Looking at it longer I think it's clearer to not delegate to intersphinx for these return types, but keeping the mock imports:

intersphinx | mock imports
--- | ---
image | image

So, I'm reverting this in lndocs: https://github.com/laminlabs/lndocs/commit/0008a002e121369982ddebf20fa1b972036c137e

lamindb/models/artifact.py
Koncopd: See this https://stackoverflow.com/questions/49733699/python-type-hints-and-context-managers

Koncopd: `AnnDataAccessor` is a class implementing context manager protocol, so when we return it, it is just of this type. Iterator type is used for lazyframe because it is returned by a function wrapped by `@contextmanager`. This is why there is a difference.

Koncopd: Claude is giving you more or less the same thing as in the stackoverflow question. I find the generator variant much more confusing.

falexwolf: Got it! Thanks for clarifying!

Koncopd: Maybe we should add this stackoverflow question to the comments near the return types somewhere?

PR #3424, opened 2026-01-05
🗃️ Rename `SQLRecord._branch_code` to `SQLRecord.branch_id` and remove `feature_ref_is_name` and `label_ref_is_name` fields from link models

falexwolf:
The `_branch_code` stems from a time when we didn't have a branch registry.

The `feature_ref_is_name` and `label_ref_is_name` fields stem from a time where we didn't identify semantic references of features and labels through their dtypes.

The change is backward compatible.

### DIFF
--- a/tests/core/test_queryset.py
+++ b/tests/core/test_queryset.py
@@ -1,16 +1,13 @@
 # .latest_version is tested in test_versioning.py
 
-import os
 import re
 import textwrap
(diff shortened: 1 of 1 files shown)


falexwolf:
Remaining failures will be resolved after re-installing `dbwrite` log by @Koncopd and fixing the security schema.

PR #3425, opened 2026-01-05
🚸 Overhaul the PyTorch `lightning` integration (v2)

Zethson:
This uses @jorenretel's excellent proposal for an improved `lightning` integration:

- https://github.com/laminlabs/lamindb/pull/3281

Improvements for the `Callback` class:

- subclasses `ModelCheckpoint`
- can be used with `lightning` CLI
- enables to save the best checkpoint and not just the last-written checkpoint
- saves checkpoints as semantic paths on storage for deployment

## Usage

See https://docs.lamin.ai/lamindb.integrations.lightning

image

image

A new `SaveConfigCallback` for more lower level control:

https://docs.lamin.ai/lamindb.integrations.lightning#lamindb.integrations.lightning.SaveConfigCallback

image

Use `save_lightning_features` to save Features which are automatically tracked when using the new ModelCheckpoint.

https://docs.lamin.ai/lamindb.integrations.lightning#lamindb.integrations.lightning.save_lightning_features

image

## Further improvements

This PR also fixes https://github.com/laminlabs/lamindb/issues/3460 by removing the uppercase constraint.


Zethson:
@falexwolf 

For deployment, SageMaker (and similar deployment systems) expect a single S3 prefix containing all model files:
```
s3://bucket/models/my_model/
├── model.ckpt
├── config.yaml
└── tokenizer/...
```

They pass this prefix to SageMaker, which copies everything underneath into the container.
Currently, when saving individual file artifacts with virtual keys (the default), files are stored at:
```
s3://bucket/.lamindb/abc123.ckpt
s3://bucket/.lamindb/def456.yaml
```

Each file gets its own uid-based path. There's no common prefix to give SageMaker.

For the Lightning integration, we want *both*:
1. **Per-checkpoint metadata** - track score, rank, is_best_model for each checkpoint so users can query "give me the best checkpoint"
2. **Deployment-ready folder** - a single S3 prefix containing the files for deployment

These conflict because:
- Per-checkpoint tracking → separate artifacts → scattered uid-based paths
- Deployment prefix → folder artifact → lose per-checkpoint granularity

Options that I considered:

1. **Duplicate data**: Copy files into a new folder artifact at deployment time → wasteful
2. **Folder-only**: Save checkpoints as folder artifact, update in place → lose ability to query individual checkpoints
3. **Add `virtual_key=False`**: Would need changes to Artifact class. `ln.settings.creation._artifact_use_virtual_keys` exists and I played around with enabling/disabling it on request but it gets messy with versioning.

What would you suggest? Is this a "post Callback problem" that needs to be resolved elsewhere?

falexwolf:
> For deployment, SageMaker (and similar deployment systems) expect a single S3 prefix containing all model files:

The "common prefix" problem is very common. We also face it when organizing large collections of parquet files to enable hive partitioning. I think we somewhat have a solution that enables distributed queries across "non-adjacent" storage keys. But unless @Koncopd thinks we can already do that we don't have full support for patterns like this:

```
/experimental_data/
├── year=2024/
│   ├── month=01/
│   │   ├── day=15/
│   │   │   ├── part-00000.parquet
│   │   │   ├── part-00001.parquet
│   │   │   └── part-00002.parquet
│   │   ├── day=16/
│   │   │   ├── part-00000.parquet
│   │   │   └── part-00001.parquet
│   │   └── day=17/
│   │       └── part-00000.parquet
│   └── month=02/
│       ├── day=01/
│       │   ├── part-00000.parquet
│       │   └── part-00001.parquet
│       └── day=02/
│           └── part-00000.parquet
└── year=2023/
    └── month=12/
        └── day=31/
            ├── part-00000.parquet
            └── part-00001.parquet
```

I think that currently the users needing this use folder-like artifacts. So, `experimental_data` would map on a 16-char folder uid as opposed to a 20-char artifact uid. But then we currently lose all versioning capability. So, that's not a very satisfactory solution.

I don't have very clear thoughts about it and if object-level versioning on the folder level is as feasible as @Koncopd says it is this might be redundant, but my sense is that enforcing a common uid prefix for a group of related artifacts is a meaningful capability. It might also be related to efforts of tracking folder-like structures in storage keys.

For your example, I could imagine something like this:

```
ln.Artifact("...", key="my_model/model.ckpt", group_by="my_model").save()
ln.Artifact("...", key="my_model/config.yaml", group_by="my_model").save()
```

For the Hive partitioning it'd look like this:
```
ln.Artifact("...", key="experimental_data/year=2024/dataset1.parquet", group_by="experimental_data/year=2024").save()
ln.Artifact("...", key="experimental_data/year=2024/dataset1.parquet", group_by="experimental_data/year=2024").save()
```

The `group_by` argument could be renamed to `storage_folder` or something similar. Users will experience folders just through the `/` structure in the `key` like they do in the S3 UI or any file path UI.

The whole topic definitely needs to be treated in another PR because it's independent of `lightning`. We could port this to a dedicated issue and discuss with @Koncopd whether this is feasible on a short time scale. It's definitely a high value capability because so many tools rely on common prefixes in storage (just git doesn't, where lamindb's zero order inspiration comes from).

Zethson:
Thanks! I think we're aligned because actually just a minute before you answered, I pushed https://github.com/laminlabs/lamindb/pull/3425/changes/0587ecb7529ab8655454a58d59679f40fc58aab9 which goes into the suggested direction. I might revert my commit and solve this issue later after we've implemented a generalized version.

@Koncopd any comment would be appreciated, please. I am happy to create a separate issue to track this.

Koncopd:
I also don't have any clear thoughts on that. Maybe we can have a call and discuss?

falexwolf:
Thanks, Lukas! Currently working through this.

> is no longer specific to a tracking framework like MLFlow or wandb

 I refactored the PR description but stumbled across this "Goal" (I rephrased). Wasn't the previous also agnostic to MLFlow and wandb?

falexwolf:
A general comment that I make for everybody. It's always better to paste screenshots from the rendered docs rather than writing out the examples in the PR description because then we can readily discuss the docs design and one often spots errors through that.

image

falexwolf:
> path_prefix stores checkpoints at predictable semantic paths (e.g., s3://bucket/deployments/my_model/epoch=0-val_loss=0.5.ckpt) for deployment workflows like SageMaker that need a known S3 prefix. This trades versioning for semantic paths - each checkpoint becomes a separate artifact, queryable via ln.Artifact.filter(key__startswith=path_prefix).

I think we need to make this clearer. Versioning artifacts is super important. S3 supports multiple versions under the same S3 key and we'll build dedicated support for it sooner or later.

If S3 bucket versioning is enabled and the user re-creates a different artifact with the same real key, AWS S3 already handles versioning.

If we now have a first use case that exposes `_key_is_virtual=False` in a public API like `path_prefix` (name to be discussed), we have to investigate in more detail the interplay of S3 versioning and LaminDB versioning.

Alternative strategies: One can also create a common folder for artifacts by creating a new storage location via `ln.Storage(...)` and then passing that storage location to the Artifact constructor: `ln.Artifact(..., storage=storage)`. This way one could satisfy the requirement for SageMaker, too. But before advocating the use of storage locations for such a frequently changing use case, we should also discuss UX concerns.

So, I believe using S3 bucket versioning is the way to go here in particular because the spec is in `fsspec` and I believe consistent for other common object storage providers. Just conventional file systems don't offer it.

lamindb/integrations/lightning.py
```diff
@@ -1,87 +1,392 @@
...
+    - logger_version (str): Version from the first Lightning logger.
+
+    Example::
+
+        from lamindb.integrations import lightning as ll
+
+        ll.save_lightning_features()
+    """
+    ln.Feature(name="is_best_model", dtype=bool).save()
+    ln.Feature(name="score", dtype=float).save()
```
falexwolf: Polluting the root namespace of the `Feature` registry with such a generic name is somewhat problematic.

It's a fallacy to think it's great if all collaborators in a LaminDB instance agree on a variable `score` with `dtype` `float`.

It's a very unspecific name that will be used for all kinds of scores. Aggregating data across this feature will then mix results from "model scores from training" and these other contexts.

Two solutions. Either we just use `run.params` with JSON values. People can query their model by querying the JSON params of the run.

```python
runs_with_score_greater_x = ln.Run.filter(params__score__gt=x)
ln.Artifact.filter(run=runs_with_score)
```

Or we prefix all these feature names. Or we move these feature names under a built-in feature type `__lamindb_lightning__` or similar.

lamindb/integrations/lightning.py
```diff
@@ -1,87 +1,392 @@
...
 
-class Callback(pl.Callback):
-    """Saves PyTorch Lightning model checkpoints to the LaminDB instance after each training epoch.
+    Extends Lightning's ModelCheckpoint with artifact creation & feature annotation.
 
-    Creates version families of artifacts for given `key` (relative file path).
+    Two modes are supported:
 
-    See also: :doc:`docs:mlflow` & :doc:`docs:wandb`.
+    1. **Versioned** (default): All checkpoints are versioned under ``key``.
```
falexwolf: See my other comment. I don't think we can trade versioning for semantic paths given versioning with semantic paths is considered a solved problem in the industry.

lamindb/integrations/lightning.py
```diff
@@ -1,87 +1,392 @@
...
+       Storage paths are managed by LaminDB (virtual keys).
+
+    2. **Deployment**: Set ``path_prefix`` to store checkpoints at predictable semantic
```
falexwolf: Let's discuss the `path_prefix` name once we have clarity on how we "always version".

PS: Please don't use double-back-ticks. We're not using them anywhere. ☺️

lamindb/integrations/lightning.py
```diff
@@ -1,87 +1,392 @@
...
-        path: A local path to the checkpoint.
-        key: The `key` for the checkpoint artifact.
-        features: Features to annotate the checkpoint.
+        key: The artifact key for checkpoints.
+            Checkpoints are versioned under this key when ``path_prefix`` is not set.
+        path_prefix: If set, store checkpoints at semantic paths under this prefix for deployment.
+            Each checkpoint becomes a separate artifact with predictable S3 paths.
+        features: Features to annotate checkpoints with.
+            Values can be static or None (auto-populated from trainer metrics/attributes).
+        dirpath: Local directory for checkpoints.
```
falexwolf: Is this name a lightning convention?

lamindb/integrations/lightning.py
```diff
@@ -1,87 +1,392 @@
...
+        save_last: Save a copy of the last checkpoint.
+        save_top_k: Number of best checkpoints to keep.
+        save_weights_only: Save only model weights (not optimizer state).
+        mode: One of "min" or "max" for monitor comparison.
+        auto_insert_metric_name: Include metric name in checkpoint filename.
+        every_n_train_steps: Checkpoint every N training steps.
+        train_time_interval: Checkpoint at time intervals.
+        every_n_epochs: Checkpoint every N epochs.
+        save_on_train_epoch_end: Run checkpointing at end of training epoch.
+        enable_version_counter: Append version to filename to avoid collisions.
```
falexwolf, line 135: Unless this is all derived from lightning, this long list of parameters appears very opinionated? Or is there some other precedent? 

If this is us imposing opinions on how to do model training then this might be problematic because "model training ops" is a solved problem with established authorities (like lightning and wandb) defining semantics. If that's the case let's critically think through whether we really want this.

lamindb/integrations/lightning.py
```diff
@@ -1,87 +1,392 @@
...
-            trainer = pl.Trainer(callbacks=[lamindb_callback])
+
+            # Semantic paths (for deployment)
+            callback = ll.Checkpoint(
+                key="deployments/my_model",
+                path_prefix="deployments/my_model/",
+                monitor="val_loss",
+                save_top_k=3,
+            )
+            # Files at: s3://bucket/deployments/my_model/epoch=0-val_loss=0.5.ckpt
```
falexwolf: This is a standard, isn't it? I like this. How about we only use this? My impression is that ML people would generally prefer this over the cryptic paths.

lamindb/integrations/lightning.py
```diff
@@ -1,87 +1,392 @@
...
+            # Query: ln.Artifact.filter(key__startswith="deployments/my_model/")
+
+            trainer = pl.Trainer(callbacks=[callback])
+
+        Using the CLI::
```
falexwolf, line 160: I like this!

Could be very nice to store the config as an input artifact for the training run.

lamindb/integrations/lightning.py
falexwolf: Ok, I see this is inherited from `lightning.ModelCheckpoint`! Then that's great and the only question is if they ever change these names. Because if they do we should use `**kwargs` for them rather than hardcode them into lamindb and then have compat issues were a certain version of lamindb has to be used with a certain version of lightning.

Due to the (current) database schema dependency of lamindb it's best to decouple from any version requirements that analytics-focused users might have.

falexwolf:
Thanks, Lukas! I see all of these (third bullet to be clarified) as valuable improvements:

image

High-level points for potential improvements:

I think the code here might excessively use lamindb's features for data that could just be JSON and thereby have low performance and is complicated to maintain. What if a user really doesn't care about features at all and just wants to track training runs and model artifacts? And, if anything, store all these parameters as JSON? That's the main sentiment from most users I talk to in the context of workflow and function tracking; just putting parameters into `run.params` is good enough for them and doesn't come with any complexity or performance hits.

There is no coupling whatsoever with lamindb's `Run` objects at this point. This might be intentional or not and it might be debatable if people should just `ln.track()` or not. I keep going back and forth on whether we should have a dedicated JSON field on `Artifact` beyond `_aux` and which name it should have. For a moment we had `Artifact.params` but there are downsides. If we really think that we can't treat checkpoint saving canonically as part of a run that creates an artifact, we could add JSON to `Artifact._aux["params"]` and expose it via `Artifact.params` as a property. But it feels odd. It feels like model training is a function/workflow/process that's being run with certain parameters and this is where the parameters should be stored as they are in all the other use cases and integrations (`nf-lamin`, `@ln.flow`, `@ln.step`, `ln.track()`, and the derived integrations with workflow managers). There is also the question of which params are run-level and which are checkpoint-level. Many params are run level and shared across checkpoints. The iteration/epoch param together with the chosen metric values are checkpoint specific. The whole thing can shine if we have a nice way to make a table that lists all artifacts produced in a run and then shows their itration/epoch data. Essentially auto-generate a "sheet" by grouping artifacts who share the same `artifact.run` just like a lamindb sheet are a bunch of records who share a `record.type`. We have the UI for this. And this works indeed much better if we use features.

To make progress it'd be super cool if you could show a bunch of runs with a bunch of checkpoint families and a bunch of queries / dataframe retrievals. You could simply keep playing in the `lamindata` instance where you already have examples for `mlflow` and `wandb`. This here would be a lightning-only example. Maybe this will reveal that using features is really meaningful and the only thing we have to do is move them under `__lamindb_lightning__` and then think hard about their life-cycle (what if we ever want to change a name of these features etc.).

The versioning issue. I think we might only want to support the canonical paths that lightning and downstream use cases expect and eliminate the `path_prefix` keyword. Then we'll also eliminate the quirk that you pass a key for a single model `"test/my_model.ckpt"` while the `lightning` user just expects `dirname` and auto-generated paths.

image

I think it'd also help to compare usage within LaminDB vs. usage with "vanilla S3" or any other canonical usage of the `lightning.ModelCheckPoint` class.

That's it for today! ☺️

falexwolf:
Just read Joren's original comment on this. ☺️ 

image

I think it comes down to the experiment that shows the value of feature annotations over the two alternatives:

1. JSON annotations via `run.params` and/or, for now, artifact._aux (we will create a public JSON field if we really like this in the long-run, that's a migration that just adds a column in the database, so will not be breaking old clients, we just have to be 💯 sure about the name) - we also have `record.extra_data` already for exactly the same purpose
2. Coding params and metrics in the path similar to this: `s3://bucket/deployments/my_model/epoch=0-val_loss=0.5.ckpt`

My sense is that there should be pre-built solutions that create paths like this `epoch=0-val_loss=0.5.ckpt`. It's almost like hive partitioning. In a way that's a very simple and very good way of doing it, but of course it's then hard to make a table to aggregate statistics.

wandb makes excessive use of their `Table` class for these purposes, which really is very similar to lamindb's sheets. So, we might have enough signal for actual value and will then definitely build the "sheets for artifacts/runs" UI as described in the comment above.

falexwolf:
It could also be worthwhile if frameworks like Hydra offer semantics on file paths.

https://hydra.cc/docs/intro/

I also found this post on using lightning with hydra: https://towardsdatascience.com/keep-track-of-your-experiments-with-hydra-b29937a99fc9/

Another helpful way to think about is imagining someone who just dumps parameters into an SQLite or parquet file.

In fact, I'm doing this myself here: https://github.com/laminlabs/arrayloader-benchmarks

It was just the easiest way to get a table showing runs vs. some variables (mostly because we currently have the sheets UI only for records and not for runs and I didn't want to create features but would actually be fine with just JSON).

image

Among the 1M ways of putting artifacts into a table and showing them next to some variables, what makes it most convenient for the user? Which API feels "just right without overhead"?

lamindb/integrations/lightning.py
Zethson: Yes. All of these are passed to the constructor:

```python
super().__init__(
            dirpath=dirpath,
            monitor=monitor,
            verbose=verbose,
            save_last=save_last,
            save_top_k=save_top_k,
            save_weights_only=save_weights_only,
            mode=mode,
            auto_insert_metric_name=auto_insert_metric_name,
            every_n_train_steps=every_n_train_steps,
            train_time_interval=train_time_interval,
            every_n_epochs=every_n_epochs,
            save_on_train_epoch_end=save_on_train_epoch_end,
            enable_version_counter=enable_version_counter,
        )
```

lamindb/integrations/lightning.py
Zethson: I don't know the final answer of course but I am pretty confident that these are stable. Lightning is so incredibly popular, they would break a huge amount of code if they renamed them without deprecation cycles.

lamindb/integrations/lightning.py
Zethson: I first thought that just making it explicit that these features are created like this is fine but you're totally right. Even then, these names are too generic and will overlap with too much other work in the instance.

> Or we move these feature names under a built-in feature type __lamindb_lightning__ or similar.

not a fan of this because it feels a bit hidden.

>  Either we just use run.params with JSON values. 

Same here. This feels too complex and we'd have a naming overlap in the Runs instead. (I just read your comment below but this is still my opinion).

> Or we prefix all these feature names.

Not pretty but pragmatic. For now, I'd opt for this but am happy to convinced that `run.params` is a better alternative.

Zethson:
> I refactored the PR description but stumbled across this "Goal" (I rephrased). Wasn't the previous also agnostic to MLFlow and wandb?

Yes, it was. I was trying to pick up all comments that we got from users which included this one. I'm pretty sure that it arouse because the docstring of the earlier version had a MLflow comment/example.

> Versioning artifacts is super important. 

I agree but sadly I only noticed quite late that our earlier discussion will not enable the desired behavior.

> If we now have a first use case that exposes _key_is_virtual=False in a public API like path_prefix (name to be discussed), we have to investigate in more detail the interplay of S3 versioning and LaminDB versioning.

I agree. 

> I think the code here might excessively use lamindb's features for data that could just be JSON and thereby have low performance and is complicated to maintain. What if a user really doesn't care about features at all and just wants to track training runs and model artifacts? 

If they don't care about features, they don't pass any Features and it's fine. The Features are currently only used if they've been saved explicitly before (naming TBD).

> And, if anything, store all these parameters as JSON? That's the main sentiment from most users I talk to in the context of workflow and function tracking; just putting parameters into run.params is good enough for them and doesn't come with any complexity or performance hits.

I'm surprised because I consider this pretty unintuitive to both do and query for. See above. I don't think that performance is a concern here because a lightning epoch is vastly more expensive than annotating an Artifact with Features.

> To make progress it'd be super cool if you could show a bunch of runs with a bunch of checkpoint families and a bunch of queries / dataframe retrievals.

OK.

> The versioning issue. I think we might only want to support the canonical paths that lightning and downstream use cases expect and eliminate the path_prefix keyword. 

I guess you're referring to `dir_path`? I guess we can overload what it does.

> Coding params and metrics in the path similar to this: s3://bucket/deployments/my_model/epoch=0-val_loss=0.5.ckpt
My sense is that there should be pre-built solutions that create paths like this epoch=0-val_loss=0.5.ckpt. It's almost like hive partitioning. In a way that's a very simple and very good way of doing it, but of course it's then hard to make a table to aggregate statistics.

I understand that people do this but I also feel like it's just a bandaid. I would have said that people should query for an Artifact and its Path using LaminDB and then deploy this specific Path (the folder).

> Among the 1M ways of putting artifacts into a table and showing them next to some variables, what makes it most convenient for the user? Which API feels "just right without overhead"?

I'm stumbling over this because this is the lightning & eventually mlflow/wand INTEGRATION. So in my eyes we want to leverage the strengths of all components. Enabling LaminDB to be a proper ML tracking framework is a huge undertaking and goes beyond this PR in my opinion. But I'll think about it and will write some things up.

lamindb/integrations/lightning.py
Zethson: I don't like it but yes people absolutely do this.

Edit: This is now the default if you use `dirpath`. Lightning Checkpoint writes the files out like this.

falexwolf:
Agreed re features! Except for the verbose long names at the root type. ☺️  Let's keep the whole features architecture except for that I think we should move them under a feature type.

We've been consistent in using `__lamindb_xxx__` as a way to indicate lamindb-generated objects, so I think that's a good way to not collide with whatever a user might want to do. It keeps the root name space free and we can make this look nice in the UI as we will for the 20 other `__lamindb_xxx__` things. These features will still come up in search and queries for the name. 

We also have to pay attention to the case in which "the user forgets to call the feature creation helper function" and or renames or deletes these features manually.

Re allowing only one semantic path way rather than optionality and looking into how versioning on S3 works for real paths is still critical to this "feeling right", I believe.

PS: Another thing we need to pay attention to is to actually pass features and not just names in the internal `add_values()` calls, because otherwise it will look up from the feature root name space (alternatively one could pass the `feature_type` but that would also need to be designed). I'll make a small change to allow this after the refactor that I currently have open.

Zethson:
Added several query examples in lamindata: https://lamin.ai/laminlabs/lamindata/transform/R6Eko32IfTJO0005

Both Features on Artifact and JSON params on Run support filtering:

```python
# Features on Artifact
ln.Artifact.filter(score__gt=0.5).to_dataframe()

# JSON params on Run
ln.Artifact.filter(run__in=ln.Run.filter(params__lr=0.01)).to_dataframe()
```

Could split it: run-level params (lr, max_epochs, monitor) on `run.params`, checkpoint-level metrics (score, is_best_model, model_rank) as Features on Artifact. The callback has access to both via the trainer. While this would make it more "correct", it would also require users to know two ways of performing queries. Our guides could cover this but guides aren't always read. I am still not convinced by dumping everything into Runs.

What do you think?

Here's my suggested TODO list now:

- [x] Investigate S3 versioning behavior and how it interplays with LaminDB. Investigate how other frameworks take care of this.
- [x]  Remove double-backticks from docstrings
- [x] Use `__lamindb_xxx__` for a subset or all Features **depending on your answer to my comment**. I am not perfectly convinced but a long Feature name prefix isn't pretty either.
- [x] Consolidate `path_prefix` into `dir_path` or something else that lightning's Checkpoint already exposes
- [ ] "Another thing we need to pay attention to is to actually pass features and not just names in the internal add_values() calls"
- [x] Store CLI config file as input Artifact for training Run
- [x] Make sister PRs on https://github.com/laminlabs/lamin-mlops/ that showcase it live
- [ ] Get user feedback

Does this sound good to you?

lamindb/integrations/lightning.py
falexwolf: > not a fan of this because it feels a bit hidden.

I'm convinced it won't feel hidden and that grouping features by feature types is the only way to make large numbers of features navigable. In the bigger teams "feature clashing" is constantly a problem.

I meant this as the answer to the comment: https://github.com/laminlabs/lamindb/pull/3425#issuecomment-3744464543

Zethson: Sorry, I think this was me misunderstanding you. What I was thinking of was like a `Kind` that we have for Artifacts. But Feature types make a lot of sense and are great.

falexwolf:
> What do you think?

Great examples! This makes it much easier to understand. (I also immediately spot ways to improve the UX for the features query.)

> I am still not convinced by dumping everything into Runs.

I was thinking to add run-level metadata to runs and artifact-level metadata to artifacts. I think this makes the long list of variables easier to navigate and the problem nicely decomposed. I'm leaning against duplicating annotations across runs and artifacts. If we use features and not JSON it feels like it should be "done properly". Given we just copy lightning semantics and these are established I'm not actually happy about and think it's correct. If, by contrast, it was just a random user-defined function whose signature will change tomorrow then JSON params would be the correct thing to use.

Given you're strongly pro features, let's keep the features. ☺️  I think one should then just as well use features to annotate runs.

Re TODOs: agreed!

> depending on your answer to my comment.

Clarified! ☺️ 

>  into dir_path or something else that lightning's Checkpoint already exposes

I think they call it `dirpath` and its part of the signature already.

lamindb/integrations/lightning.py
falexwolf: Yes, it will show nicely nested like a "folder" or "parent" in the UI. And people who work with lightning will immediately spot it. The internal `Literal` options for `Artifact.kind` are indeed hidden and meant to be hidden.

The one thing that feels not correct is the first line here:

```python
lightning_feature_type = ln.Feature.get(name="__lamindb_lightning__")
assert lightning_feature_type.is_type
features = lightning_feature_type.features.all()
```

Also users might start to edit the name field - so this is not good. One could lock the type object to protect against edits. 🤔 

Or we could use the protected `_dtype_str` field for this. It then serves as a "meta type" that qualifies what the type is. I think this is a good way to solve this. So, something like this:

```python
lightning_feature_type = ln.Feature(name="lamindb.lightning", description="Auto-generated features tracking lightning params & metrics", is_type=True)
lightning_feature_type._dtype_str = "__lamindb_lightning__"
lightning_feature_type.save()
```

falexwolf: And then this query would work:
```python
lightning_feature_type  = ln.Feature.get(dtype="__lamindb_lightning__")
```

This feels odd though but ad hoc I don't have a better idea. 🤔

Zethson: `ln.Feature.get(dtype="__lamindb_lightning__")` does not work though. Only `ln.Feature.get(_dtype_str="__lamindb_lightning__")` or `ln.Feature.get(name="Lamindb.Lightning")`

falexwolf: > ln.Feature.get(dtype="__lamindb_lightning__") does not work though. 

Are you sure? 🤔 That's what I added here with a test:

- https://github.com/laminlabs/lamindb/pull/3454

falexwolf: > ln.Feature.get(name="Lamindb.Lightning")

We should not enforce the capitalization. We should just print a warning and recommend camel case.

falexwolf: This test: https://github.com/laminlabs/lamindb/pull/3454/files#r2687345338

But at this point this is all just tweaking the UX around the life cycle and retrieval of this feature type. We can sleep another night on this.

Zethson: I'll check it later but the query

```python
ln.Feature.get(dtype="__lamindb_lightning__")
```

fails with

```
FieldError: Unknown field 'dtype'. 
```

I'll omit the full traceback but you know which one it is.

falexwolf: I see, it works for `filter()` but not for `get()`. Will fix this.

Zethson:
**On S3 (or more generally cloud versioning)**
When you overwrite a file, S3 doesn't delete the old one - it assigns a unique `VersionId` to each version. However, as far as I know, this is not enabled by default. While we can ensure that it's on for the instances that we create, this might not be true for all customer instances that use their own storage. I don't know whether we have a way of forcing specific storage settings. 

**I fear that always relying on non-virtual keys could break local storage usage where such versioning is surely absent.** We want to avoid too many if statements in our code.

S3 versioning is enabled at the bucket level, not per-object. Each PUT to the same key creates a new version with a unique `VersionId`. GET without `VersionId` returns latest, GET with `VersionId` returns specific version. DELETE just adds a "delete marker" but versions still exist. GCS and Azure have similar semantics. fsspec supports `version_id` for reads but we don't currently track this AFAIK.

Virtual keys already cause file accumulation (`.lamindb/abc123.ckpt`, `.lamindb/def456.ckpt`, ...) so semantic keys with S3 versioning don't introduce new cost concerns.

For the lightning integration specifically, key collisions aren't a real concern because lightning auto-generates unique filenames like epoch=0-val_loss=0.5.ckpt. Different epochs/metrics produce different keys, so we won't overwrite previous checkpoints.

Having both per-checkpoint queryability and deployment-ready paths at the same time needs LaminDB changes. The `group_by` proposal would let artifacts keep virtual keys while sharing a storage prefix.
We could also consider making it easier to move/copy artifacts that were created for deployment into specific folders: ` artifact.copy_to("s3://bucket/deploy/model/")`. However, it sounds like polluting the API. 

Claude always tends to pushing back here and suggests separating these concerns. It suggests not to mess with lineage and considers it fine that we have two options now that allow satisfying both needs to a large enough degree.

I generally still think that we, as LaminDB, should ensure that our versioning is robust and that the paths are simple to retrieve. MLEs that want to deploy a model can easily copy Artifacts to the desired location and we can consider how we can make this easier.
I don't think that non-virtual keys should become the default because they're a mess with local storage etc.

falexwolf:
> However, as far as I know, this is not enabled by default. While we can ensure that it's on for the instances that we create, this might not be true for all customer instances that use their own storage. I don't know whether we have a way of forcing specific storage settings.

One could monitor that likely even from `fsspec`. In most contexts outside lamindb users prefer to switch versioning on. For lamindb both bucket settings are OK.

> We want to avoid too many if statements in our code.

Agreed.

> I fear that always relying on non-virtual keys could break local storage usage where such versioning is surely absent. 

But that could be user expectation. You make a new folder and you have paths for params, so versioning is not critical, in particular because likely whatever you're doing is not production anyway.

> fsspec supports version_id for reads but we don't currently track this AFAIK.

We don't. Syncing this with lamindb versioning would be a task. My main question is what happens when you do

```python
open("sample.fasta", "w").write("123") 
ln.Artifact("sample.fasta", key="sample.fasta").save()
open("sample.fasta", "w").write("456") 
ln.Artifact("sample.fasta", key="sample.fasta").save()
```

We have bucket versioning switched on for `s3://lamindata` AFAIK and I'm pretty sure that S3 versioning and LaminDB versioning would be automatically in sync simply because versioning a file is a simple linear process. One would have to think about how robust it is and where it could go wrong and how a query for a version in lamindb would translate to the corresponding call in `fsspec`. But all of this should, IMO, not be terribly complicated and is needed anyway.

But you make the most important argument below anyway, which is why it's not absolutely necessary to build all this versioning-coupling right now if we can just be confident that we'll build it at some point this year.

> For the lightning integration specifically, key collisions aren't a real concern because lightning auto-generates unique filenames like epoch=0-val_loss=0.5.ckpt. Different epochs/metrics produce different keys, so we won't overwrite previous checkpoints.

Exactly.

That's why I think that in practice it's low probability that versioning is ever useful in the lightning context. And if it's needed S3 versioning does the job. And if somebody runs things locally, I'd hope that our API already raises a warning or even an error. That'd be easy to add.

---

If we decided to only support the auto-generated lamindb paths, I'd be fully bought in. But given you (and I) are convinced that the real paths are valuable I'm pretty convinced that simplifying to only supporting the real paths is more maintainable, leads to a simpler API, and leads to less user surprise / quicker adoption.

> they're a mess with local storage etc.

I don't think they're a mess. We'll have the renaming / path moving very soon and any Nextflow pipeline writes them. Tons of data is based on real paths and we should use them whenever there is true value you to them. If there is no value then the virtual paths are better; but it's not that they are "much better".

I'd like to make the "real storage keys" usage much more abundant also in light of Hive partitioning and potential integrations with technologies like Iceberg. So, this is not an edge case.

> Having both per-checkpoint queryability and deployment-ready paths at the same time needs LaminDB changes. The group_by proposal would let artifacts keep virtual keys while sharing a storage prefix.
We could also consider making it easier to move/copy artifacts that were created for deployment into specific folders:  artifact.copy_to("s3://bucket/deploy/model/"). However, it sounds like polluting the API.

Yes, let's not do this.

---

Here are reasons why I'm leaning towards not supporting virtual paths in the lightning integration:

1. I think we create risk with a complex API that supports two very different "usage modes" with opposite behaviors that we have to explain to a user, bloating the signature and the docs.
2. My guess is that most lightning users don't even want to bother learning about versioning that they can't use for deployment (where it'd be most important). They want to use Lamin for lineage and maybe annotation/validation. 
3. Most of our users are enterprise or biotech or techbio and working in the cloud and actually deploying stuff not just training something on a laptop or an academic cluster. Most compute already happens in the cloud and it will only get more.
4. We can't remove something we support later on, but we can easily support virtual paths later on if we somehow find that the real paths are too problematic.
5. What you describe ("lightning auto-generates unique filenames like `epoch=0-val_loss=0.5.ckpt`") aligns with my mental model of experiment tracking frameworks. This is a solved problem and a cohesive philosophy and somehow working against it with lamindb's virtual paths seems like fighting and unnecessary fight.

Let's go on a call and nail a decision. In the meantime, happy to read invalidation of these "pro real path" or more "pro virtual path" arguments if there some. ☺️

Zethson:
> What you describe ("lightning auto-generates unique filenames like epoch=0-val_loss=0.5.ckpt") aligns with my mental model of experiment tracking frameworks. This is a solved problem and a cohesive philosophy and somehow working against it with lamindb's virtual paths seems like fighting and unnecessary fight.

There's edge cases here though because if you use the same `dirpath` and your training run produces the same validation metric value, you'd still overwrite the output (on local storage or with cloud storage versioning disabled). It can be argued that the same validation metric value (rounded at least) might not come with the bit exact same model.

> One could monitor that likely even from `fsspec`. In most contexts outside lamindb users prefer to switch versioning on. 

How much trust are we putting into users to enable/disable the right storage settings? I don't think that's in our control so we should bake our solution accordingly. We can warn if it's disabled but users will do what users do.

> Everything else

Yeah, I think we're aligned. I'm just worried about users that create their own cloud Storage and don't have versioning on. We promise versioning (generally as part of our product) and "you should have read the docs" is the wrong answer when someone overwrote their checkpoints/deployment models. Training runs can be very expensive.

falexwolf:
> with cloud storage versioning disabled

I'd need to check but I think you'd get an error. S3 doesn't make it so easy to overwrite. Also it'd be very easy to throw an error in LaminDB if versioning is disabled on S3 or if you're on a local file system, wouldn't it?

So, I think, we can make this completely safe. 

> How much trust are we putting into users to enable/disable the right storage settings? I don't think that's in our control so we should bake our solution accordingly. We can warn if it's disabled but users will do what users do.

None. But see above!

> I'm just worried about users 

See above! ☺️

Zethson:
> Also it'd be very easy to throw an error in LaminDB if versioning is disabled on S3 or if you're on a local file system, wouldn't it?

Yes, but then users would have to go to their admin to get this changed. This takes time, effort and causes frustration. 

> I'd need to check but I think you'd get an error. S3 doesn't make it so easy to overwrite.

I think only if options like [ifnonematch](https://docs.aws.amazon.com/sdk-for-swift/latest/api/awss3/documentation/awss3/putobjectinput/ifnonematch/) are enabled. Otherwise, I think it just overwrites silently (also what Claude suggests).

falexwolf:
> Yes, but then users would have to go to their admin to get this changed. This takes time, effort and causes frustration.

I'd raise the error only if an overwrite is about to happen. Given how unlikely that is it'd only let the user realize that they'd need to set a new value for `dirname` or fix some other mistake they made.

Hence: I don't think they'd go to an admin with time, effort and frustration involved. It'd just be a simple error along the lines of "you'd overwrite this artifact, please choose a new dirname or pass `overwrite_versions=True`".

Please let me know if I'm missing something! If your argument is true then it's of course important.

> I think it just overwrites silently (also what Claude suggests).

Ok! That also makes sense given that often this is indeed desired. Thanks for checking! 🙏

Zethson:
> I'd raise the error only if an overwrite is about to happen. Given how unlikely that is it'd only let the user realize that they'd need to set a new value for dirname or fix some other mistake they made.
> 
> Hence: I don't think they'd go to an admin with time, effort and frustration involved. It'd just be a simple error along the lines of "you'd overwrite this artifact, please choose a new dirname or pass overwrite_versions=True".

All right then let's roll with that. Summa summarum: I'll only allow `dirpath` and non-virtual paths.
TODO Lukas, do not forget backwards compatibility changes because we've removed several arguments.

falexwolf:
> Summa summarum: I'll only allow dirpath and non-virtual paths.

Yes! I think that'll give rise to the prettiest, simplest interface this integration can have. 🤩 And it'll be closest to user expectations from lightning users. ☺️

lamindb/integrations/lightning.py
falexwolf: Are you sure you're rebased to main? It works for me for a query against `laminlabs/lamindata`:

image

lamindb/integrations/lightning.py
```diff
@@ -1,40 +1,389 @@
...
+    - logger_version (str): Version from the first Lightning logger.
+
+    Example::
+
+        from lamindb.integrations import lightning as ll
 
-    Creates version families of artifacts for given `key` (relative file path).
+        ll.save_lightning_features()
+    """
+    # lamindb.lightning is lowercase and would trigger a naming warning -> mute
```
Zethson: Not a fan of this whole section. I am tempted to even remove the warning because it's not the biggest of deals if people name their things lowercase. We're a bit opinionated.

lamindb/models/sqlrecord.py
```diff
@@ -258,25 +258,28 @@ class ValidateFields:
...
+        s: The string to check
     """
     if "[" in s:  # this is because we allow types of form 'script[test_script.py]'
         return True
     last_component = s.split(".")[-1]
 
-    if not last_component[0].isupper():
-        raise ValueError(
-            f"'{last_component}' should start with a capital letter given you're defining a type"
+    # Find the first alphabetical character
```
Zethson: As written above, I'm thinking of just getting rid of this. Too much complexity and noise

lamindb/integrations/lightning.py
falexwolf: Ok!

lamindb/models/sqlrecord.py
falexwolf: Ok!

.gitmodules
```diff
@@ -7,9 +7,9 @@
 [submodule "sub/bionty"]
 	path = sub/bionty
 	url = https://github.com/laminlabs/bionty
-[submodule "sub/wetlab"]
```
Zethson, line 10: Was at least necessary/helpful for me. Can revert if it breaks existing redirects of other people.

tests/core/test_track_script_or_notebook.py
```diff
@@ -382,6 +382,26 @@ def test_clean_r_notebook_html():
     orig_notebook_path.write_text(content.replace(get_shortcut(), "SHORTCUT"))
 
 
+def test_notebook_to_script_notebooknode_metadata(tmp_path):
```
Zethson, line 379: Just a merge from the other branch - will get updated/removed when it is merged.

lamindb/_finish.py
```diff
@@ -165,6 +165,7 @@ def notebook_to_script(  # type: ignore
     import jupytext
 
     notebook = jupytext.read(notebook_path)
+    notebook.metadata.clear()
```
Zethson, line 168: From other PR

Zethson:
@falexwolf would appreciate another review, please. See https://6978c1ff9fea2614b8b77c41--lamin-mlops.netlify.app/guide for what it looks like in action now.

I'll then follow up with Joren.

falexwolf:
I know you wanted to await the final feedback from @jorenretel, but I'd say we can have another PR that polishes this more based on his feedback. I don't find flaws with this.

I'd merge it now because it has good things in it that I keep seeing myself wanting like the removal of upper case restriction.

And also, importantly, because I don't want to create conflicts here when migrating from ipynb to markdown:

- https://github.com/laminlabs/lamin-mlops/pull/32

PR #3428, opened 2026-01-06
🐛 Fix broken imports from `hubmodule`

fredericenard:
This is fixing boken imports due to `hubmodule` refactoring (see [PR](https://github.com/laminlabs/laminhub/pull/4145)). Let me know if you find something else that needs to be fixed @Koncopd.


Koncopd (approved):
Thanks.

PR #3430, opened 2026-01-06
🗃️ Bionty v2 migrations

sunnyosun:
With https://github.com/laminlabs/bionty/pull/344, https://github.com/laminlabs/wetlab/pull/129

### DIFF
--- a/docs/storage/transfer-local-to-cloud.ipynb
+++ b/docs/storage/transfer-local-to-cloud.ipynb
@@ -30,10 +30,10 @@
    "source": [
     "def cleanup(artifact: ln.Artifact):\n",
     "    features_sets = artifact.schemas.all()\n",
-    "    experiments = artifact.experiments.all()\n",
+    "    compounds = artifact.compounds.all()\n",
     "    artifact.delete(permanent=True, storage=False)\n",
     "    features_sets.delete()\n",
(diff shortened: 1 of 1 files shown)


sunnyosun:
Failed tests are due to remote instances are not migrated.

lamindb/migrations/0150_squashed.py
```diff
@@ -1793,9 +1793,9 @@ class Migration(migrations.Migration):
             },
             bases=(
                 lamindb.models.sqlrecord.HasType,
+                lamindb.models.has_parents.HasParents,
```
sunnyosun, line 1796: This change order of the bases is necessary for initializing a new instance, otherwise, one gets this error: TypeError: Cannot create a consistent method resolution order (MRO) for bases Model, object, HasParents

lamindb/models/has_parents.py
```diff
@@ -146,9 +147,15 @@ def _query_ancestors_of_fk(record: SQLRecord, attr: str) -> SQLRecordList:
     return SQLRecordList(ancestors)
 
 
-class HasParents:
+class HasParents(models.Model):
```
sunnyosun: Had to revert this as I couldn't resolve django errors raised in bionty models with this change.

PR #3433, opened 2026-01-06
👷 Ensure `wetlab` version `2.0a1` is installed

falexwolf:
To get CI fully working again.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -173,7 +173,7 @@ def install_ci(session, group):
     if IS_PR or group == "docs":
         run(
             session,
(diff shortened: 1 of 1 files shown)


Koncopd:
This PR added installation of wetlab from submodules but didn’t remove the redundant previous installations of wetlab from individual groups.

falexwolf:
The submodule installation will only run on PRs and on docs, not on push events e.g. on the release branch.

I think it's good to indicate which groups require wetlab, and which don't by specifying the extra. That's what the release branch run will use.

I'm confident we'll find better ways of organizing this in the future.

Issue #3436, opened 2026-01-07
Support grouping artifacts under a common storage prefix without losing per-artifact metadata


Zethson:
### Add a description

Many deployment and data processing systems (e.g. SageMaker, Spark/Hive) require a *single storage prefix* containing all relevant files. Today, Lamindb stores artifacts under uid-based virtual keys:

```
s3://bucket/.lamindb/<uid1>.ckpt
s3://bucket/.lamindb/<uid2>.yaml
```

This makes it impossible to hand a common prefix to downstream tools that expect folder-like semantics:

```
s3://bucket/models/my_model/
├── model.ckpt
├── config.yaml
└── tokenizer/...
```

At the same time, users need *per-artifact granularity*:

* checkpoint-level metadata (score, rank, is_best_model)
* querying individual checkpoints (“give me the best checkpoint”)
* versioning at the artifact level

These two requirements currently conflict.

**Current limitations**

* Folder-like artifacts provide a common prefix but lose artifact-level versioning and metadata.
* Per-artifact storage preserves metadata but scatters files across unrelated prefixes.
* Disabling virtual keys (`virtual_key=False`) is brittle and interacts poorly with versioning.
* Workarounds like copying files into a deployment folder are wasteful and ad hoc.

**Proposed direction**

Introduce an explicit way to *group related artifacts under a shared storage prefix* while keeping them as independent artifacts in the database.

Example API sketch:

```python
ln.Artifact("model.ckpt", key="my_model/model.ckpt", group_by="my_model").save()
ln.Artifact("config.yaml", key="my_model/config.yaml", group_by="my_model").save()
```

This would result in:

```
s3://bucket/.lamindb/<group_uid>/my_model/model.ckpt
s3://bucket/.lamindb/<group_uid>/my_model/config.yaml
```

Key properties:

* A stable common prefix usable by deployment systems
* Each file remains a first-class Artifact with its own UID, metadata, and versioning
* Folder semantics are inferred purely from storage keys (`/`), consistent with S3 UX
* Generalizes to other use cases (e.g. Hive-partitioned parquet layouts)

The `group_by` argument could be renamed to something like `storage_folder` or `storage_prefix`.

falexwolf:
After aligning within a call among @Zethson @Koncopd and @falexwolf, conclusion is that we use real paths to solve problem and for now enable this with a private flag:

```python
ln.Artifact("...", key="my_model/model.ckpt", _is_virtual_key=False).save()
[...]

PR #3438, opened 2026-01-07
✨ Allow passing `_key_is_virtual` on `Artifact` creation

Koncopd:
Addresses:

- https://github.com/laminlabs/lamindb/issues/3436
- https://github.com/laminlabs/pfizer-lamin-usage/issues/412

This PR adds a private argument to `Artifact` that allows setting the private field `_key_is_virtual` on init:

```python
# writes to ln.setup.settings.storage.root / "myfile.csv"
artifact = ln.Artifact(path, key="myfile.csv", _key_is_virtual=False).save()
```

This is to start adopting more internal usage of real storage paths prior to deciding on a public API.


falexwolf (approved):
Thanks!

PR #3441, opened 2026-01-08
👷 Reflect permission utils refactoring in noxfile

falexwolf:


noxfile.py
```diff
@@ -152,10 +152,12 @@ def install_ci(session, group):
...
     elif group == "permissions":
-        run(
-            session,
-            "uv pip install --system --no-deps ./laminhub/backend/laminhub_rest/hubmodule",
-        )
+        cmds = "uv pip install --system --no-deps ./laminhub/backend/laminhub_rest/hubmodule"
+        cmds += "\nuv pip install --system -e ./laminhub/backend/utils"
+        cmds += "\nuv pip install --system -e ./laminhub/backend/central"
+        cmds += "\nuv pip install --system -e ./laminhub/backend/dbinstance"
+        cmds += "\nuv pip install --system -e ./laminhub/backend/aws"
```
Koncopd, line 159: `-e` is not needed here i think

Koncopd: Also better to move down

Koncopd: to line 188 where `./laminhub/backend` is installed.

falexwolf: Ok!

falexwolf: I just looked through it and I think this could be more thoroughly refactored, but I don't have enough knowledge to do it well. I'll leave that to another PR. This PR was to fix CI.

PR #3443, opened 2026-01-08
🗃️ Migrate most auxiliary fields to SQL columns

falexwolf:
Migrate auxiliary fields from the `_aux` JSON field to standard Django fields, truncate Schema UIDs, and rename `n` to `n_members`.

## Background

The auxiliary fields system stores certain model attributes in a JSON field (`_aux`) under the key `"af"`. Each field is indexed by a string number ("0", "1", etc.) and accessed via Python properties. This migration converts them to proper Django fields.

## Fields to Migrate

| Model | Current Property | New Field | Type | Nullable | Default |
|-------|------------------|-----------|------|----------|---------|
| **Artifact** | `_is_saved_to_storage_location` | `_save_completed` | `BooleanField` | No | `False` |
| **Run** | `cli_args` | `cli_args` | `CharField(max_length=1024)` | Yes | `None` |
| **Feature** | `default_value` | `default_value` | `JSONField` | Yes | `None` |
| **Feature** | `nullable` | `nullable` | `BooleanField` | Yes | `None` |
| **Feature** | `coerce_dtype` | `coerce` | `BooleanField` | Yes | `None` |
| **Schema** | `coerce_dtype` | `coerce` | `BooleanField` | Yes | `None` |
| **Schema** | `flexible` | `flexible` | `BooleanField` | Yes | `None` |
| **Schema** | `n` (rename) | `n_members` | `IntegerField` | Yes | `None` |

### Nullability Rationale

- **`_save_completed`**: Internal tracking field, always has a meaningful value.
- **`cli_args`**: Only set when run is invoked from CLI.
- **`default_value`**: Most features don't have a default value.
- **`nullable`, `coerce`, `flexible`, `n_members`**: These are nullable because Features and Schemas can be "type-like" (`is_type=True`), in which case these settings don't apply. `NULL` means "not applicable".

For non-type records (`is_type=False`), the migration sets concrete values:
- `nullable`: from `_aux` or `True` (default)
- `coerce`: from `_aux` or `False` (default)
- `flexible`: from `_aux` or `n_members < 0` (legacy default)
- `n_members`: keeps existing value (renamed from `n`)

For type records (`is_type=True`), all these fields are set to `NULL`.

## Schema UID Truncation

Since lamindb 1.5, Schema UIDs are 16 characters (previously 20). This migration:
1. Truncates existing 20-char UIDs to 16 chars
2. Reduces `max_length` from 20 to 16

## Fields Remaining as Auxiliary given they're not converged

These fields will stay in `_aux["af"]`:
- **Schema** `optionals` (`_aux["af"]["1"]`) - list of optional feature UIDs
- **Schema** `_index_feature_uid` (`_aux["af"]["3"]`) - UID of index feature

### DIFF
--- a/lamindb/migrations/0172_alter_artifact__save_completed_alter_feature_coerce_and_more.py
+++ b/lamindb/migrations/0172_alter_artifact__save_completed_alter_feature_coerce_and_more.py
@@ -11,58 +11,79 @@ class Migration(migrations.Migration):
     ]
 
     operations = [
-        migrations.AlterField(
-            model_name="artifact",
-            name="_save_completed",
-            field=lamindb.base.fields.BooleanField(blank=True, default=None, null=True),
-        ),
-        migrations.AlterField(
-            model_name="feature",
-            name="coerce",
-            field=lamindb.base.fields.BooleanField(blank=True, default=None, null=True),
-        ),
-        migrations.AlterField(
-            model_name="feature",
-            name="default_value",
-            field=lamindb.base.fields.JSONField(blank=True, default=None, null=True),
-        ),
-        migrations.AlterField(
-            model_name="feature",
-            name="nullable",
-            field=lamindb.base.fields.BooleanField(blank=True, default=None, null=True),
-        ),
-        migrations.AlterField(
-            model_name="run",
-            name="cli_args",
-            field=lamindb.base.fields.CharField(
-                blank=True, default=None, max_length=1024, null=True
-            ),
-        ),
-        migrations.AlterField(
-            model_name="schema",
-            name="coerce",
-            field=lamindb.base.fields.BooleanField(blank=True, default=None, null=True),
-        ),
-        migrations.AlterField(
-            model_name="schema",
-            name="flexible",
-            field=lamindb.base.fields.BooleanField(blank=True, default=None, null=True),
-        ),
-        migrations.AlterField(
-            model_name="schema",
-            name="n_members",
-            field=lamindb.base.fields.IntegerField(blank=True, default=None, null=True),
-        ),
-        migrations.AlterField(
-            model_name="schema",
-            name="uid",
-            field=lamindb.base.fields.CharField(
-                blank=True,
-                db_index=True,
-                default=None,
-                editable=False,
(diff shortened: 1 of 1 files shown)


lamindb/models/artifact.py
```diff
@@ -1403,6 +1400,8 @@ def labels(self) -> LabelManager:
     """The creator of this artifact."""
     _overwrite_versions: bool = BooleanField(default=None)
     """See corresponding property `overwrite_versions`."""
+    _save_completed: bool = BooleanField(default=False, db_default=False)
```
Koncopd: This is not correct, the current field can be None or bool, where None indicates no write to storage is needed. So it is not correct to use this name and change the type without changing the logic.

Koncopd:
There is a problem with changes for artifact, commented.

lamindb/models/artifact.py
falexwolf: Thanks! Changing to nullable and ensuring that logic stays exactly as it is.

Koncopd: The name is also problematic, because it doesn’t really indicate that the writing to storage is completed, not saving.

Koncopd: We can change the logic ofc so that this indicates saving if you think it is better.

falexwolf: I went back and forth about this and I think the crucial thing that's to be conveyed is that the `artifact.save()` operation fully completed. And yes, that's because of the storage component but I'd say this name is fine and not too ambiguous with a small docstring. The previous name, despite its extreme length and verbosity wasn't very intuitive either. It's hard to name because we can't say "upload". This name is at least quite a bit simpler than the previous without sacrificing interpretability.

I've just migrated all our instances. Unless you have a killer good name, let's keep it like this. If you have one, we can run another migration tomorrow. I hope we don't. 😅

Koncopd: I still think the new name is confusing. _write_completed would be better in my opinion.

falexwolf: A save and a write event are the same thing, aren't they? 😅

PR #3445, opened 2026-01-09
🗃️ Refactor `_storage_completed` mechanism and remove unique constraints on type names

falexwolf:
This follows with more consolidation of changes made here:

- https://github.com/laminlabs/lamindb/pull/3443

### DIFF
--- a/lamindb/migrations/0175_squashed.py
+++ b/lamindb/migrations/0175_squashed.py
@@ -4580,6 +4580,7 @@ class Migration(migrations.Migration):
                     models.ManyToManyField(
                         related_name="users",
                         through="lamindb.ArtifactUser",
+                        through_fields=("user", "artifact"),
                         to="lamindb.artifact",
                     ),
                 ),
--- a/tests/storage/test_artifact_zarr.py
+++ b/tests/storage/test_artifact_zarr.py
@@ -37,7 +37,7 @@ def test_zarr_upload_cache(get_small_adata):
 
     assert ln.Artifact.get(path=artifact.path) == artifact
 
-    assert artifact._storage_completed
+    assert artifact._storage_completed is None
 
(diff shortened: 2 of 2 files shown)


falexwolf:
@Koncopd, a brief double check?

lamindb/models/artifact.py
```diff
@@ -1713,6 +1706,30 @@ def overwrite_versions(self) -> bool:
         """
         return self._overwrite_versions
 
+    @property
+    def _storage_completed(self) -> bool | None:
```
Koncopd, line 1710: `None` is here to avoid unnecessary database hits for cases (registering existing paths) when no write to a storage is needed.

Koncopd:
tqdm is here https://github.com/laminlabs/lamindb/blob/eef301905ba12b2c96921188051ec1c3dbd2b532/lamindb/migrations/0101_alter_artifact_hash_alter_feature_name_and_more.py#L11

falexwolf:
But this should never run. Was always there.

falexwolf:
Oh, I see.

lamindb/models/artifact.py
```diff
@@ -2901,7 +2912,7 @@ def save(
         # the saving / upload process has been successful, just mark it as such
         # maybe some error handling here?
         if flag_complete:
-            self._save_completed = True
+            self._storage_completed = True
```
falexwolf: @Koncopd, can we set this to `None` rather than `True`? Because then we have much sparser data and an easier query. Less load on the JSON field. Essentially we only use this "_storage_completed: False" as an indicator for failed uploads, but we never indicate successful uploads.

This also makes it 💯 a case for the JSON field as opposed to a true SQL column.

Koncopd: Discussed, i am fine with it.

Issue #3427, opened 2026-01-05
Return layers_keys, obs_keys, and obsm_keys if only available


mojtababahrami:
Hey! 
I have a feature request for the `MappedCollection` module which I think is quite useful when scaling to many datasets with different schemas and structures. Now, MappedCollection tries to return the layers_keys, obs_keys, and obsm_keys from every anndata storage file in its `__getitem__`, which requires all the underlying anndata files to have them all. However, when scaling to many heterogeneous datasets for large-scale training (e.g. scBaseCount + CellxGene + Tahoe), every dataset has a different set of layers and obs/obsm keys stored. It would be great if the MappedCollection module only returned the keys that are available for each sample and ignored them if not available. It should be quite easy to check on the fly within the `__getitem__` like below for obsm_key and similarly for others:

```
def __getitem__(self, idx: int):
....
        if self.obsm_keys is not None:
            for obsm_key in self.obsm_keys:
                if obsm_key in store["obsm"].keys():    # <<<<<<  Like this  !
                    lazy_data = store["obsm"][obsm_key]
                    out[f"obsm_{obsm_key}"] = self._get_data_idx(lazy_data, obs_idx)
....
```

falexwolf:
Thanks for this, @mojtababahrami! 

This sounds very reasonable and easily feasible. Could you attempt a PR - it would be part of the upcoming release (lamindb v2 on Sunday? ☺️

mojtababahrami:
Hey @falexwolf 
I have created the PR #3446 for this.
I have also added support to read layer_keys from `.raw` too, which is similarly helpful when extending to many heterogeneous datasets.

falexwolf:
Wow, that's great! @Koncopd will review it next week! ☺️

Koncopd:
This was implemented - https://github.com/laminlabs/lamindb/pull/3446 https://github.com/laminlabs/lamindb/pull/3582

PR #3446, opened 2026-01-09
✨ Improve data retrieval in `MappedCollection`

mojtababahrami:
- Return layers_keys, obs_keys, and obsm_keys if only available
- Add support for reading layer_keys in `.raw`

Addresses #3427

### DIFF
--- a/lamindb/core/_mapped_collection.py
+++ b/lamindb/core/_mapped_collection.py
@@ -184,6 +184,11 @@ class MappedCollection:
                 X = store["X"]
                 store_path = self.path_list[i]
                 self._check_csc_raise_error(X, "X", store_path)
+                if isinstance(X, ArrayTypes):  # type: ignore
+                    n_obs_storage = X.shape[0]
+                else:
+                    n_obs_storage = X.attrs["shape"][0]
+                indices_storage = np.arange(n_obs_storage)
                 if self.filtered:
                     indices_storage_mask = None
                     for obs_filter_key, obs_filter_values in obs_filter.items():
@@ -191,8 +196,11 @@ class MappedCollection:
                             obs_filter_values = list(obs_filter_values)
                         elif not isinstance(obs_filter_values, list):
                             obs_filter_values = [obs_filter_values]
-                        obs_labels = self._get_labels(store, obs_filter_key)
-                        obs_filter_mask = np.isin(obs_labels, obs_filter_values)
+                        if obs_filter_key in store["obs"]:
+                            obs_labels = self._get_labels(store, obs_filter_key)
+                            obs_filter_mask = np.isin(obs_labels, obs_filter_values)
+                        else:
+                            obs_filter_mask = np.full(n_obs_storage, False)
                         if pd.isna(obs_filter_values).any():
                             obs_filter_mask |= pd.isna(obs_labels)
                         if indices_storage_mask is None:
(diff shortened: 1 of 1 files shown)


Zethson:
Thank you very much!

Sorry about the CI. Secrets currently aren't available for forks - it's on our roadmap to fix that.

Koncopd:
Hi, @mojtababahrami,thank you for the PR. Unfortunately it can't be merged as it is. Please see my comments.

lamindb/core/_mapped_collection.py
```diff
@@ -376,31 +377,49 @@ def __getitem__(self, idx: int):
...
+                if lazy_data is None:
+                    continue
                 out[layers_key] = self._get_data_idx(
                     lazy_data, obs_idx, self.join_vars, var_idxs_join, self.n_vars
                 )
             if self.obsm_keys is not None:
                 for obsm_key in self.obsm_keys:
-                    lazy_data = store["obsm"][obsm_key]
-                    out[f"obsm_{obsm_key}"] = self._get_data_idx(lazy_data, obs_idx)
+                    if obsm_key in store["obsm"].keys():
```
Koncopd: This just can't be done on every index rertrieval. Getting keys from `hdf5` or `zarr` storage is an expensive operation, so this should be cached on init instead and then checked.

lamindb/core/_mapped_collection.py
```diff
@@ -376,31 +377,49 @@ def __getitem__(self, idx: int):
...
-                        cats = self._cache_cats[label][storage_idx]
-                        if cats is None:
-                            cats = []
-                    else:
-                        cats = None
-                    label_idx = self._get_obs_idx(store, obs_idx, label, cats)
-                    if label in self.encoders and label_idx is not np.nan:
-                        label_idx = self.encoders[label][label_idx]
-                    out[label] = label_idx
+                    if label in store["obs"].keys():
```
Koncopd: Same here, this just kills the overall performance completely.

lamindb/core/_mapped_collection.py
mojtababahrami: I see. I didn't know it's an expensive retrieval. Yes we can definitely cache them in init. I'll do that

mojtababahrami:
Hey @Koncopd 
I have added the required caching. 

By the way, I was not able to run the tests for it locally. I get this error:
`RuntimeError: Failed to load instance laminlabs/lamin-dev, please check your permissions!`
But I think it should pass on github ci/cd. There shouldn't be a breaking change

Koncopd:
Thank you, i will check soon.

Koncopd:
Sorry for the delay, i will revisit this soon.

Koncopd:
Hello, @mojtababahrami , sorry, i have a lot things going, i will dedicate some time to this soon.

mojtababahrami:
Hey @Koncopd 
That would be great! 
I have noticed that I had to add a few more lines to handle when keys are not available. 
As I can not run the tests, I was not able to thoroughly test it. But it should work, I think.

Koncopd:
@mojtababahrami ok, megred, i will add tests and some improvements separately.

mojtababahrami:
Thanks a lot!

falexwolf:
Ohhhh, this is great! Thank you guys! 🙏

PR #3452, opened 2026-01-10
🗃️ Rename `RootBlock` to `Block` with a column `key` akin to `Artifact` and `Transform`

falexwolf:
For parallelism with `Artifact` and `Transform` in case `Block` becomes a first-class citizen in the API.

### DIFF
--- a/lamindb/models/sqlrecord.py
+++ b/lamindb/models/sqlrecord.py
@@ -2163,15 +2163,21 @@ class SQLRecordInfo:
                 core_module_fields.append(field)
 
         def _get_related_field_type(field) -> str:
-            field_type = (
-                field.related_model.__get_name_with_module__()
-                .replace(
-                    "Artifact", ""
-                )  # some fields have an unnecessary 'Artifact' in their name
-                .replace(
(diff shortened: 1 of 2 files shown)


Koncopd:
`RootBlock` is hard-coded in RLS.

falexwolf:
Good to know!

falexwolf:
I'll update laminhub accordingly!

Koncopd:
What is the new name, just `lamindb_block`? Were there any other changes like this?

Koncopd:
I will quickly make a change.

falexwolf:
Ok! You can simply remove it!

falexwolf:
It inherits from `SQLRecord` meanwhile like a first-class citizen. All "attached blocks" are the same as they were.

falexwolf:
> What is the new name, just lamindb_block? 

Yes!

> Were there any other changes like this?

No!

Koncopd:
Done here https://github.com/laminlabs/laminhub/pull/4202

falexwolf:
Thanks!

PR #3457, opened 2026-01-13
🐛 Fix str behavior of `Artifact.from_dataframe()`

Zethson:
Noticed while looking at https://github.com/laminlabs/lamindb/pull/3455 again when it got merged.

Pretty sure that my changed code would break if a str is passed and not a Pathlike. This is fixed and tested in this PR.

### DIFF
--- a/lamindb/models/artifact.py
+++ b/lamindb/models/artifact.py
@@ -1963,7 +1963,8 @@ class Artifact(SQLRecord, IsVersioned, TracksRun, TracksUpdates):
             if path.suffix == ".parquet":
                 import pyarrow.parquet as pq
 
-                artifact.n_observations = pq.read_metadata(path.open("rb")).num_rows
+                with path.open("rb") as f:
+                    artifact.n_observations = pq.read_metadata(f).num_rows
             else:
                 # csv/tsv/others have no metadata and would require a full expensive read
                 artifact.n_observations = None


lamindb/models/artifact.py
```diff
@@ -1959,7 +1959,7 @@ def from_dataframe(
             artifact.n_observations = len(df)
         else:
             # must be a str or path
-            path = df
+            path = UPath(df)
```
Koncopd: Do you assume here that this can be only a local path? If not, i.e. it can be for example a remote path in lamin-mnaged storage, then it should be `create_path(df)` instead.

Zethson: I assumed that it's a local path here but I don't see a reason why it couldn't be a remote path.

I'll take a look again. Thanks!

Koncopd: Ok, if you converge on a local path, please indicate it in the docstring.

lamindb/models/artifact.py
```diff
@@ -1959,11 +1959,11 @@ def from_dataframe(
             artifact.n_observations = len(df)
         else:
             # must be a str or path
-            path = df
+            path = create_path(df)
             if path.suffix == ".parquet":
                 import pyarrow.parquet as pq
 
-                artifact.n_observations = pq.read_metadata(path).num_rows
+                artifact.n_observations = pq.read_metadata(path.open("rb")).num_rows
```
Koncopd: This is very probably should use context manager, i.e. `with path.open(...) as f`, because i am not sure that `pq.read_metadata` closes properly.

PR #3458, opened 2026-01-13
🔒️ Introduce safe-guards in updating the space of artifacts

Koncopd:
This PR:

- adds a mechansim to track changed fields for existing/saved records
- forbids changing `space` for existing artifacts in space-managed storages.

It prepares a PR that moves datasets from one space-associated storage location to another when moving the artifacts across spaces.

### DIFF
--- a/lamindb/models/artifact.py
+++ b/lamindb/models/artifact.py
@@ -1902,8 +1902,7 @@ class Artifact(SQLRecord, IsVersioned, TracksRun, TracksUpdates):
 
         Args:
             df: A `DataFrame` object or a `UPathStr` pointing to a `DataFrame` in storage, e.g. a `.parquet` or `.csv` file.
-            key: A relative path within default storage,
-                e.g., `"myfolder/myfile.parquet"`.
+            key: A relative path within default storage, e.g., `"myfolder/myfile.parquet"`.
             description: A description.
             revises: An old version of the artifact.
             run: The run that creates the artifact.
@@ -2412,6 +2411,7 @@ class Artifact(SQLRecord, IsVersioned, TracksRun, TracksUpdates):
             data: A file path or in-memory dataset object like a `DataFrame`, `AnnData`, `MuData`, or `SpatialData`.
             run: `Run | bool | None = None` The run that creates the artifact.
                 If `False`, suppress tracking the run.
+                If `None`, infer the run from the global run context.
             format: `str | None = None` The format of the data to write into storage.
                 If `None`, infer the format from the data.
 


lamindb/models/artifact.py
```diff
@@ -2833,6 +2835,11 @@ def delete(
         """
         super().delete(permanent=permanent, storage=storage, using_key=using_key)
 
+    @property
+    def _is_in_auto_storage(self) -> bool:
```
falexwolf: Can we think harder about how to name the "automatically created uid-prefixed storage locations"?

This is a very important concept. Most storage locations will be auto-generated because it's a pain to deal with bucket creation and semantic names. Most SaaS products use this, too, of course.

So, we need a name for an "auto-generated storage location". It's also important to convey to users that these storage locations are "auto-generated by lamindb".

LaminDB has a "generator" for "storage locations" to serve:

* storage access permission management
* region resolution
* other use cases that require keeping data under a common prefix in a bucket or file system

So, I think, in this PR, we could update the `lamindb.storage` doc with a definition of this and a good name. And then create a property on `Storage` and/or `StorageSettings` akin to `is_managed_by_hub` or `is_cloud`.

Then we don't need a private property.

---

And having written all this I realize that this property is meant to indicate the `".lamindb/"` folder location. 🙈 - So the comment is for another time, I guess.

I would call this property "_has_auto_generated_key" or something similar. It's a bit of a pity that this here is so complicated:
```
self._real_key is None and (self._key_is_virtual or self._key is None)
```

The second part here feels like a design mistake:
```
self._key_is_virtual or self._key is None
```

Because if `self._key is None` I'd say `self._key_is_virtual` should be implied but it never occured to me. What happens if virtual keys are switched off and somebody tries to create an artifact without passing a `key`? I'm pretty sure they run into an error.

lamindb/models/artifact.py
```diff
@@ -2858,6 +2865,11 @@ def save(
 
                 artifact = ln.Artifact("./myfile.csv", key="myfile.parquet").save()
         """
+        if self._field_changed("space_id") and self._is_in_auto_storage:
+            raise ValueError(
```
falexwolf, line 2866: I would issue a warning if the artifact is not in `.lamindb/`. But more importantly I don't understand why we only consider artifacts stored under `.lamindb` 🤔

lamindb/models/artifact.py
```diff
@@ -2858,6 +2865,11 @@ def save(
...
+                "Space cannot be changed after creation for artifacts in auto storage."
```
falexwolf: The language is very confusing if we don't document what "auto storage" is.

lamindb/models/artifact.py
falexwolf: IMO the whole space-storage integrity coupling applies to any storage location (be it auto-generated or not) and whether data is under `.lamindb/` or under semantic storage keys.

One could allow people to bypass that integrity if we think there is low risk. But then we should do something like:

```python
with ln.permissions.by_pass_storage_space_map:
     ...
```

(Not relevant now. But it's conceivable that in certain situations people don't care about managing access at the storage level because there is no security concern, just a governance concern. And then it's convenient to move the artifact record into a restricted space while keeping it in a more broadly accessible storage location.)

lamindb/models/sqlrecord.py
```diff
@@ -1086,6 +1120,7 @@ def save(self: T, *args, **kwargs) -> T:
         if pre_existing_record is not None:
             init_self_from_db(self, pre_existing_record)
         else:
+            # TODO: refactor to use TRACK_FIELDS
```
falexwolf, line 1123: Yes, I really like what you did with TRACK_FIELDS. That's a simple & elegant solution. Much better than the ad hoc key and name change tracking.

lamindb/models/artifact.py
Koncopd: >IMO the whole space-storage integrity coupling applies to any storage location (be it auto-generated or not) and whether data is under .lamindb/ or under semantic storage keys.

We discussed this, right? And concluded that there are no problems with changing space if the artifact is already not in the storage attached to the space or even there but not in the auto storage key path / auto path / auto storage.

lamindb/models/artifact.py
Koncopd: >I'm pretty sure they run into an error.

No, in this case it will be created in the auto generated path, as with `_key_is_virtual=True`. This is why this check is as it is now here and in some other places.

lamindb/models/artifact.py
Koncopd: Right, but i am not sure how else to name it. Also where should this be documented?

lamindb/models/artifact.py
falexwolf: Let's re-discuss on a call tomorrow. My assumption is that the storage location is in the space.

lamindb/models/artifact.py
Koncopd: This is not relevant anymore

lamindb/models/artifact.py
```diff
@@ -2859,6 +2861,16 @@ def save(
 
                 artifact = ln.Artifact("./myfile.csv", key="myfile.parquet").save()
         """
+        # note that storage is not editable after creation
+        if (
+            self._field_changed("space_id")
+            and (artifact_storage := self.storage).instance_uid is not None
+            and artifact_storage.space_id == self._original_values["space_id"]
```
Koncopd, line 2864: @falexwolf i have changed the check to this.

falexwolf: This looks good to me! But shouldn't we run the exact same lines in `Artifact("...", storage=storage, space=space)`?

Koncopd: It already ignores the storage in this case.

Koncopd:
@falexwolf can i merge this?

lamindb/models/artifact.py
falexwolf: I see that you manually checked but I meant that we call the same code so that we don't have implicit entanglement of validation.

It'd be much easier to maintain and assert correctness if the same function is called.

tests/core/test_artifact_basics.py
```diff
@@ -1248,3 +1248,28 @@ def test_save_url_with_virtual_key():
...
+    artifact = ln.Artifact.get(key="test_space_change.tsv")
+    artifact.space = space
+    with pytest.raises(ValueError) as err:
+        artifact.save()
+    assert (
+        "Space cannot be changed because the artifact is in the storage location of another space."
+        in err.exconly()
+    )
+
+    artifact.delete(permanent=True, storage=True)
```
falexwolf: I don't think you need `storage=True` here.

lamindb/models/artifact.py
falexwolf: Clarified on a call. Please a make component in both parts of the code to cross-reference each other.

lamindb/models/artifact.py
```diff
@@ -2119,7 +2126,8 @@ def from_mudata(
 
         Args:
             mdata: A `MuData` object.
-            key: A relative path within default storage, e.g., `"myfolder/myfile.h5mu"`.
+            key: A relative path within default storage,
```
falexwolf: @Zethson @Koncopd can you two please check your linter settings?

We don't want to go back and forth.

This just changed hours ago in Lukas's PR causing lot's of noise in the diffs.

Koncopd: yep, it is not linter, i overwrote the file by accident. fixed

PR #3466, opened 2026-01-15
🍱 Support CELLxGENE 7.0.0

Zethson:
We simplify CELLxGENE integration by removing pinned ontology version support in favor of curating against latest versions.

In the future, we support users to
1. Validate with BioHub's `cellxgene-schema` tool
2. Then use LaminDB's CELLxGENE integration to ingest validated artifacts & register metadata

This is achieved by
- Remove pinned ontology version handling
- Fix behavior for pinned sources (not sure where it got lost - was already broken before https://github.com/laminlabs/lamindb/pull/3516) & adding a test that always runs
- Simplify test suite using a Visium CELLxGENE dataset

Remaining TODOs
- [ ] Update guide


sunnyosun:
You are unblocked via https://github.com/laminlabs/bionty/pull/359.

falexwolf:
Finalizing this now.

falexwolf:
That was easy. Mini conflict, everything passes. Merging! 😄

PR #3481, opened 2026-01-21
📝 Better return typing of `feature.__getitem__`

sunnyosun:


lamindb/models/_feature_manager.py
```diff
@@ -920,7 +921,15 @@ def get_values(self, external_only: bool = False) -> dict[str, Any]:
         """
         return get_features_data(self._host, to_dict=True, external_only=external_only)  # type: ignore
 
-    def __getitem__(self, feature: str) -> Any | dict[str, Any]:
+    def __getitem__(
+        self, feature: str
+    ) -> (
+        DtypeStr
```
ap--: `DtypeStr` is wrong, since it's just the literals that are supported to chose the dtype. I think these should be the types as returned,

It should be a union of the actual dtypes.

https://github.com/laminlabs/lamindb/blob/9f27746d15540bf409b86628dae025b612450d63/lamindb/base/types.py#L56-L67

so `int | float | str | datetime | date | UPath`

sunnyosun: Ah, thanks for catching it!

PR #3483, opened 2026-01-22
🐛 Revert `artifact` changes on failed `.replace()` with no storage corruption

Koncopd:


Koncopd:
Not sure this is a good idea now.

Issue #3480, opened 2026-01-21
`track()` RepresenterError


Zethson:
### Add a description

As of this year, I regularly get errors like:

```python
---------------------------------------------------------------------------
RepresenterError                          Traceback (most recent call last)
Cell In[4], line 2
      1 # track this notebook/script run so that all checkpoint artifacts are associated with the source code
----> 2 ln.track(params=MODEL_CONFIG, project=ln.Project(name="MLflow tutorial").save())

File ~/code/lamindb/lamindb/core/_context.py:545, in Context._track(self, transform, project, space, branch, features, params, new_run, path, pypackages)
    543     if description is None:
    544         description = self._description
--> 545     self._create_or_load_transform(
    546         description=description,
    547         transform_ref=transform_ref,
    548         transform_ref_type=transform_ref_type,
    549         transform_type=transform_type,  # type: ignore
    550     )
    551 else:
    552     if transform.kind in {"notebook", "script"}:

File ~/code/lamindb/lamindb/core/_context.py:779, in Context._create_or_load_transform(self, description, transform_ref, transform_ref_type, transform_type, key)
    773 source_code_path = ln_setup.settings.cache_dir / self._path.name.replace(
    774     ".ipynb", ".py"
    775 )
    776 if (
    777     self._path.exists()
    778 ):  # notebook kernel might be running on a different machine
--> 779     notebook_to_script(description, self._path, source_code_path)
    780     _, transform_hash, _ = hash_file(source_code_path)
    781 else:

File ~/code/lamindb/lamindb/_finish.py:168, in notebook_to_script(title, notebook_path, script_path)
    165 import jupytext
    167 notebook = jupytext.read(notebook_path)
--> 168 py_content = jupytext.writes(notebook, fmt="py:percent")
    169 # remove global metadata header
    170 py_content = re.sub(r"^# ---\n.*?# ---\n\n", "", py_content, flags=re.DOTALL)

File ~/miniforge3/envs/lamindb/lib/python3.12/site-packages/jupytext/jupytext.py:488, in writes(notebook, fmt, version, config, **kwargs)
    485     update_jupytext_formats_metadata(metadata, fmt)
    487 writer = TextNotebookConverter(fmt, config)
--> 488 return writer.writes(notebook, metadata)

File ~/miniforge3/envs/lamindb/lib/python3.12/site-packages/jupytext/jupytext.py:247, in TextNotebookConverter.writes(self, nb, metadata, **kwargs)
[...]

PR #3488, opened 2026-01-26
🐛 Fix PyYAML serialization of NotebookNode

Zethson:
Fixes https://github.com/laminlabs/lamindb/issues/3480

It is not clear to me why I only recently started encountering this issue but I did update my environment. However, I think we should have encountered this in our CI if it were only about the most recent versions.

## Problem

`ln.track()` fails with `RepresenterError: ('cannot represent an object', {'display_name': '...', 'language': 'python', 'name': 'python3'})` when converting notebooks to scripts.

## Cause

`NotebookNode` (from nbformat) is a `dict` subclass, but PyYAML's `SafeDumper` only does exact type matching in `yaml_representers` - it doesn't fall back to parent classes in the MRO. This is a known issue: https://github.com/yaml/pyyaml/issues/142

When jupytext calls `yaml.safe_dump()` to serialize notebook metadata, nested `NotebookNode` objects (like `kernelspec`) fail to serialize.

## Fix

Register a custom representer for `NotebookNode` that serializes it as a plain dict.

### DIFF
--- a/tests/core/test_track_script_or_notebook.py
+++ b/tests/core/test_track_script_or_notebook.py
@@ -377,23 +377,22 @@ def test_clean_r_notebook_html():
 
 
 def test_notebook_to_script_notebooknode_metadata(tmp_path):
-    """Test that notebook_to_script handles NotebookNode metadata https://github.com/laminlabs/lamindb/issues/3480."""
+    """Test that notebook_to_script handles NotebookNode metadata.
+
+    https://github.com/laminlabs/lamindb/issues/3480
+    """
     import nbformat
-    import yaml
     from lamindb._finish import notebook_to_script
 
-    # Create and convert a minimal notebook to trigger representer registration
     nb = nbformat.v4.new_notebook()
+    nb.metadata["kernelspec"] = nbformat.NotebookNode({"display_name": "python3"})
     notebook_path = tmp_path / "test.ipynb"
     nbformat.write(nb, notebook_path)
-    notebook_to_script("Test", notebook_path)
 
-    # Now test that NotebookNode can be serialized (fails without discarding metadata)
-    data = {
-        "jupyter": {"kernelspec": nbformat.NotebookNode({"display_name": "python3"})}
-    }
-    result = yaml.safe_dump(data)
-    assert "display_name: python3" in result
+    # This would raise RepresenterError without metadata.clear()
+    result = notebook_to_script("Test", notebook_path)
+    assert result is not None
+    assert "NotebookNode" not in result
 
 
 class MockRun:


falexwolf:
Just seeing this: We have to be a little careful with this.

I'm trying to remember whether there was a conscious decision against stripping notebook metadata because it makes the source code hash essentially useless.

I understand that what you're doing here is serializing metadata as yaml at the top of `source_code`, right?

Let me read the code whether there are relevant comments.

falexwolf:
Yes, so to ensure the `hash` is meaningful we consciously strip all metadata:

https://github.com/laminlabs/lamindb/blob/1fb683ef2b8636aec11d1314caad2614de5b6ef8/lamindb/_finish.py#L169-L170

If there is an issue with writing the yaml, rather than fixing it, I'd say we should not write it in the first place. Wouldn't it be weird to create the custom logic that you're proposing just to strip it in the next line? This is what I'm talking about:

```python
import yaml
from nbformat import NotebookNode

# PyYAML's SafeDumper.represent_data() only checks yaml_representers for exact type match on mro[0].
# The MRO fallback loop only checks yaml_multi_representers, which is empty.
# So dict subclasses like NotebookNode fail to serialize even though dict is in yaml_representers.
# This is a known issue: https://github.com/yaml/pyyaml/issues/142
yaml.add_representer(
  NotebookNode,
  lambda dumper, data: dumper.represent_dict(data),
  Dumper=yaml.SafeDumper,
)
```

If you run into an issue then of course we have to fix the issue. I just hope with less complexity.

Zethson:
> If there is an issue with writing the yaml, rather than fixing it, I'd say we should not write it in the first place. Wouldn't it be weird to create the custom logic that you're proposing just to strip it in the next line? 

You're totally right. Actually, just clearing the metadata is good enough then. I've tested that the fix in this PR allows me to track notebooks again.

Zethson:
@falexwolf could you please have another look?

lamindb/_finish.py
```diff
@@ -165,6 +165,7 @@ def notebook_to_script(  # type: ignore
     import jupytext
 
     notebook = jupytext.read(notebook_path)
+    notebook.metadata.clear()
     py_content = jupytext.writes(notebook, fmt="py:percent")
     # remove global metadata header
     py_content = re.sub(r"^# ---\n.*?# ---\n\n", "", py_content, flags=re.DOTALL)
```
falexwolf, line 171: Do we still need this then? I think we should be able to remove this line now that we have line 168.

Otherwise looks great!

falexwolf: I think `notebook.metadata.clear()` is more elegant than the regex. I just wasn't aware back then.

Zethson: From what I know, yes.

 Jupytext injects its own jupytext.text_representation metadata regardless of metadata.clear().

falexwolf: Ok! That would have been a nice comment but we'll have it here now on GitHub if we ever wonder again ☺️

falexwolf: I'll keep this conversation open so that it'll be easy to find in the future!

PR #3489, opened 2026-01-26
💚 Fix white-space formatting in tests for new Rich behavior

Zethson:
Fixes whitespace changes introduced in https://github.com/Textualize/rich/blob/master/CHANGELOG.md#1430---2026-01-24

### DIFF
--- a/tests/core/test_record_sheet_examples.py
+++ b/tests/core/test_record_sheet_examples.py
@@ -161,12 +161,12 @@ id,uid,name,treatment,cell_line,preparation_date,project,__lamindb_record_uid__,
     assert artifact.features.describe(return_str=True).endswith("""\
 └── Dataset features
     └── columns (7)
-        cell_line           bionty.CellLine         HEK293T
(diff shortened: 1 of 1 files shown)


falexwolf:
Upper-bounding rich resolves the issue. I think this could work for a while because we only need the upper bound in the `dev` dependencies.

If rich resolves this or if we somehow make our tests less sensitive to white space can be figured out down the road.

falexwolf:
Can you move the upper bound to the dev dependencies and then merge so that we unblock the other PRs? ☺️

Zethson:
I'm very close to fixing the tests - just give me a few more minutes, please.
This is new (and more correct) Rich behavior so this is here to stay.

falexwolf:
> I'm very close to fixing the tests - just give me a few more minutes, please.
This is new (and more correct) Rich behavior so this is here to stay.

Perfect! 🙇 

If you canonically fix this is much better of course!

PR #3499, opened 2026-01-29
📝 Move lazy upload out of arrays guide and into `from_lazy()` docstring + unit tests

falexwolf:
This doesn't mean we'll not show this in a guide, but I will be refactoring `arrays.ipynb` to be exclusively about queryies/streaming/slicing along with reformatting.

### DIFF
--- a/tests/storage/test_streaming.py
+++ b/tests/storage/test_streaming.py
@@ -300,7 +300,6 @@ def test_from_lazy():
 
 
 def test_from_lazy_cloud():
-    """Covers from_lazy docstring: local (zarr.open) and cloud (FsspecStore)."""
     previous_storage = ln.setup.settings.storage.root_as_str
     ln.settings.storage = "s3://lamindb-test/storage"
     lazy = ln.Artifact.from_lazy(


falexwolf:
FYI @Koncopd, I'm now working on restructuring the whole arrays.ipynb. This is just a first step.

falexwolf:
@Koncopd, was there a bug in the guide?

image

Can you please check?

tests/storage/test_streaming.py
```diff
@@ -299,6 +299,26 @@ def test_from_lazy():
     artifact.delete(permanent=True, storage=True)
 
 
+@pytest.mark.parametrize("use_fsspec", [False, True])
+def test_from_lazy_local_and_cloud(use_fsspec):
+    """Covers from_lazy docstring: local (zarr.open) and cloud (FsspecStore)."""
+    lazy = ln.Artifact.from_lazy(
```
Koncopd, line 305: this should not use a local path

tests/storage/test_streaming.py
```diff
@@ -299,6 +299,26 @@ def test_from_lazy():
...
+        suffix=".zarr", overwrite_versions=True, key="stream_test.zarr"
+    )
+    if use_fsspec:
+        store = zarr.storage.FsspecStore.from_url(lazy.path.as_posix())
+        group = zarr.open(store, mode="w")
+        group["ones"] = np.ones(3)
+    else:
+        store = zarr.open(lazy.path, mode="w")
```
Koncopd: this won't even work with a cloud path (or at least it didn't work with the zarr version at the time of writing this in the guide), so should be removed

tests/storage/test_streaming.py
```diff
@@ -299,6 +299,26 @@ def test_from_lazy():
...
 
 
+@pytest.mark.parametrize("use_fsspec", [False, True])
+def test_from_lazy_local_and_cloud(use_fsspec):
+    """Covers from_lazy docstring: local (zarr.open) and cloud (FsspecStore)."""
+    lazy = ln.Artifact.from_lazy(
+        suffix=".zarr", overwrite_versions=True, key="stream_test.zarr"
+    )
+    if use_fsspec:
+        store = zarr.storage.FsspecStore.from_url(lazy.path.as_posix())
```
Koncopd: this uses zarr v3, but unit-storage uses zarr v2

docs/arrays.ipynb
```diff
@@ -686,89 +685,6 @@
...
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {
-    "tags": [
-     "hide-output"
-    ]
-   },
-   "outputs": [],
-   "source": [
-    "lazy = ln.Artifact.from_lazy(suffix=\".zarr\", overwrite_versions=True, key=\"mydata.zarr\")\n",
```
Koncopd, line 713: We still need to show this in the guide, otherwise it is just impossible to find in the docs. I mean it is higly notrivial thing that needs to be demostrated. Can we keep it in some form in the guide please?

falexwolf:
Thanks for clarifying!

> Can we keep it in some form in the guide please?

Yes, happy to show it somewhere. Just not in a guide on querying datasets. ☺️ 

We'll need to create a guide on "creating artifacts" or repurpose `tutorial` for this.

tests/storage/test_streaming.py
```diff
@@ -299,6 +299,23 @@ def test_from_lazy():
     artifact.delete(permanent=True, storage=True)
 
 
+def test_from_lazy_cloud():
+    """Covers from_lazy docstring: local (zarr.open) and cloud (FsspecStore)."""
+    previous_storage = ln.setup.settings.storage.root_as_str
+    ln.settings.storage = "s3://lamindb-ci/test-data"
```
Koncopd: why here and not in s3://lamindb-test/storage?

falexwolf: Sure, we can also use that one! I'll adapt!

Koncopd: doens't even work with s3://lamindb-ci/test-data

tests/storage/test_streaming.py
```diff
@@ -299,6 +299,23 @@ def test_from_lazy():
     artifact.delete(permanent=True, storage=True)
 
 
+def test_from_lazy_cloud():
+    """Covers from_lazy docstring: local (zarr.open) and cloud (FsspecStore)."""
```
Koncopd: docstring here doesn't makes sense

tests/storage/test_streaming.py
```diff
@@ -299,6 +299,23 @@ def test_from_lazy():
...
+    previous_storage = ln.setup.settings.storage.root_as_str
+    ln.settings.storage = "s3://lamindb-ci/storage"
```
Koncopd: should be s3://lamindb-test/storage, not s3://lamindb-ci/storage"

falexwolf:
Thanks for adding those commits, @Koncopd! 🙏

PR #3505, opened 2026-01-30
🚸 Warn about reducing the default value for `limit` in `.to_dataframe()` from 100 to 20

falexwolf:
This warns about making the default `limit` of `to_dataframe()` consistent with `.search()` results.

More importantly, it limits the default number of objects users see. 100 objects is on the high end when the desire is to "browse objects".

A reduction means that read logs are more sparsely populated with more relevant data, and users don't have to call `.head()` on every dataframe they query.

The downside is that sometimes users might run an actual work load against a dataframe and there might be edge cases when `100` was "still enough", but `20` is not. The typical analytics workload however will use much more than `100` objects.


falexwolf:
@ap--, this change here _could_ break a workflow but I consider it an edge case.

If you think it's better to delay to v3, then I'll not merge it.

ap--:
> @ap--, this change here _could_ break a workflow but I consider it an edge case.
> 
> If you think it's better to delay to v3, then I'll not merge it.

Would it be possible to keep the default until `2.3.0` and raise a FutureWarning for calls on `to_dataframe` that are larger than 20 but smaller than 100 and then switch the default in `2.3.0` ?

Then at least two minor versions had the warning that the default will switch.

falexwolf:
That's possible but because of how frequent the `to_dataframe()` call is for users it'd result in a LOT of warnings. With all the bad consequences for aesthetics and user-awareness for warnings in general. I don't think I want to burden users with this.

Given you're concerned I think the way I'd do it is add a warning banner to the docs saying that this 100 value will be reduced to 20 in lamindb v3. This will likely only be a year from now, but this is fine-tuning; not urgent. 100 is a reasonable default, too. ☺️ 

Thank you for commenting! 🙏

ap--:
> That's possible but because of how frequent the `to_dataframe()` call is for users it'd result in a LOT of warnings. With all the bad consequences for aesthetics and user-awareness for warnings in general. I don't think I want to burden users with this.

Hmm, I don't think this will be problematic. If your warning text is static, default emit behavior will only show the warning once. And additionally if limit is explicitly set, no warning is raised, so in your internal code all is good. If limit is not set, and the returned dataframe has more than 20 or less than 100 rows you emit a warning that the default will change. Users that used to_dataframe on records that had more than 100 rows should already be aware of the fact that there's a limit.

So it'll be at most one warning per script or notebook execution. And it'll go away in two minor versions. 🤷

falexwolf:
Thank you for expounding! I'll do what you say just not for v2.1.0 but a bit after the launch.

lamindb/models/query_set.py
```diff
@@ -1181,6 +1182,16 @@ def to_dataframe(
         if limit is not None and len(df) > limit:
             is_truncated = True
             df = df.iloc[:limit].copy()
+        # Keep the default limit at 100 until 2.6.0 while preparing users
+        # for the upcoming switch to 20.
+        if limit == 100 and 20 < len(df) < 100:
```
falexwolf, line 1187: Darn, this should have been <= 100 here.

PR #3516, opened 2026-02-03
🚸 Broaden curation to dtypes that are unions

Zethson:
It is now possible to define Union dtypes using the `|` character:

image

which can then also be validated against:

image

## implementation details

Union dtype features require per-registry `filter_kwargs`, field, and organism lookups since these are registry-specific. The existing `self._*` attributes are computed in `__init__` for only the first field, so validation methods now recompute these for each additional registry in the union.

### DIFF
--- a/docs/curate.md
+++ b/docs/curate.md
@@ -352,7 +352,7 @@ Since not all metadata is always stored within the dataset itself, it is also po
 !python scripts/curate_dataframe_external_features.py
 ```
 
-## Union of dtypes
+## Union dtypes
 
 Some metadata columns might validate against several registries.
 


tests/curators/test_curators_examples.py
```diff
@@ -518,6 +518,33 @@ def test_schema_mixed_ensembl_symbols(ccaplog):
     schema.delete(permanent=True)
 
 
+def test_schema_mixed_features(ccaplog):
+    """Test that union dtype features validate against multiple registries."""
+
+    mixed_feature = ln.Feature(
+        name="mixed_feature",
+        dtype="cat[bionty.Tissue.ontology_id|bionty.CellType.ontology_id]",
```
falexwolf, line 526: I think we also support and advocate in our docs:
```
dtype=[bt.Tissue.ontology_id, bionty.CellType.ontology_id]
```

This being said, the `|` notation might also be more intuitive here.

The string notation raises a warning meanwhile.

All of these are not primary concerns of this PR of course and could be polished another time!

falexwolf:
Could you expand tests so that these are covered?

image

image

The amount of nesting and logic grows yet again with this PR, which is OK for as long as the lines are covered.

I'm insisting because the method that does most of the work is now at a point where it's very hard for a human or an AI to assert correctness just because of long/nested it is. I think we'll thoroughly refactor the whole `curators/core.py` module at some point to make it easier to digest. But not now. ☺️

falexwolf:
Great, Lukas! 🙏 

One last ask: Can you add a screenshot to the PR description showing the examples in the context of the docs page that documents them? (Currently the PR doesn't update the docs.)

tests/curators/test_curators_examples.py
Zethson: Interesting. I haven't seen this and Claude couldn't find any solution from the `summary.md` either.

falexwolf: We showcased this syntax for several years (all of this dates back to an early trial in 2022) here: https://github.com/laminlabs/lamindb/blob/ec45fa12a3c0622391aa77f67b44cfa690d2a8e7/lamindb/models/feature.py#L479-L484

I now remember that I removed showcasing it rather recently because I was concerned that the `|` syntax would likely be the right thing.

I think the tests still exist though.

No need to add the `[A, B]` syntax back, we should probably just deprecate it and then remove it.

Zethson:
I simplified the code a bit more, improved the docs, and added screenshots.
Good to go @falexwolf ?

falexwolf:
Great! And it seems you were able to address the coverage issues! 😄 

image

Thanks!

PR #3520, opened 2026-02-04
📝 README v5.5: re-write the why? section

falexwolf:
With @Zethson and this [Gemini chat](https://gemini.google.com/app/7f315f8cb3399bd0).

Also see:

- https://github.com/laminlabs/lamin-docs/pull/399

And previous PR:

- https://github.com/laminlabs/lamindb/pull/3510

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,8 @@ But without **queryable & validated data** or with data locked in organizational
 Imagine building software without git or pull requests: an agent's quality would be impossible to **verify**.
 While code has git and tables have dbt/warehouses, biological data has lacked a framework for managing its unique complexity.
 
-LaminDB fills the gap. It is a lineage-native lakehouse that understands bio-registries and formats (`AnnData`, `.zarr`, …) based on the established open data stack:
+LaminDB fills the gap.
+It is a lineage-native lakehouse that understands bio-registries and formats (`AnnData`, `.zarr`, …) based on the established open data stack:
 Postgres/SQLite for metadata and cross-platform storage for datasets.
 By offering **queries, tracing & validation in a single API**, LaminDB provides the **context & memory** to turn messy, agentic biological R&D into a scalable process.
 


README.md
```diff
@@ -1,47 +1,58 @@
...
-With one API, you get: lakehouse, lineage, feature store, ontologies, bio-registries & formats.
+You get context & memory through a lineage-native lakehouse that understands bio-formats, registries & ontologies.
 
 <details>
 <summary>Why?</summary>
 
-Reproducing and understanding how datasets, models, and workflows were created is crucial to high-quality R&D — especially as agents increasingly contribute to it.
-At the same time, training models across thousands of datasets — from LIMS and ELNs to orthogonal assays and cross-team silos — is now a major learning opportunity, but requires queryable & validated data.
-Robustly scaled learning operations need something biology has lacked: an API-first data management framework comparable to git for code or warehouses for tables.
+(1) **Reproducing, tracing & understanding** how datasets, models & results are created is critical to quality R&D.
```
Zethson, line 9:
```suggestion
(1) **Understanding** how datasets, models & results were created is critical to quality R&D.
```

Simpler. We could try to cram reproduce in there because it matters a lot but I feel like it's easy to read like this.

README.md
```diff
@@ -1,47 +1,58 @@
...
+Without **context**, agents & humans make mistakes and **cannot close feedback loops** across data generation & analysis.
+Without **memory**, vast compute and intelligence are wasted on fragmented, non-compounding tasks.
```
Zethson:
```suggestion
Without **memory**, a lot of effort is often wasted on the wrong tasks.
```

Simpler

README.md
```diff
@@ -1,47 +1,58 @@
...
 
-LaminDB fills the gap with a lineage-native data lakehouse that understands bio-registries and formats (`AnnData`, `.zarr`, …).
-It provides queries across many datasets with enough freedom to maintain high-paced R&D while automating rich context on top of versioning, change management, and other industry standards.
+(2) Training & fine-tuning models with thousands of datasets — across LIMS, ELNs, orthogonal assays — is now a primary path to **scaling R&D**.
+But without **queryable & validated data** or with data locked in organizational & infrastructure siloes, it leads to **garbage in, garbage out** or is quite simply impossible.
```
Zethson, line 14:
```suggestion
However,  **unvalidated data** locked in organizational & infrastructure siloes is challenging to access efficiently.
```

This is a tricky sentence because it makes two statements:
1. Queries are useful
2. Validated data is useful.

Then the order is also off. Queries are mentioned first then validated data, but the "bad results" are in the opposite order - first garbage data and then impossible queries.

README.md
```diff
@@ -1,47 +1,58 @@
...
+
+Imagine building **software without git or pull requests**: an agent's quality would be impossible to verify.
```
Zethson: IDK maybe it's just me but I feel like it's too big of an intellectual jump here that git & PRs are the history & queryablity that you mentioned above and that this has the potential to verify agent quality.

README.md
```diff
@@ -1,47 +1,58 @@
...
+While code has git and tables have warehouses, biological data has lacked an open-source framework for managing its unique complexity.
+
+**LaminDB fills the gap.** It is a lineage-native lakehouse that understands bio-registries and formats (`AnnData`, `.zarr`, …) based on the established open data stack:
+Postgres/SQLite for metadata and cross-platform storage systems for big datasets.
+By offering **queries, tracing & validation in a single API**, LaminDB provides the **context & memory** to turn messy, agentic biological R&D into a scalable process.
```
Zethson, line 22:
```suggestion
By offering **validation, tracing & queries in a single API**, LaminDB provides the **context & memory** to turn messy, agentic biological R&D into a scalable process.
```

Feels a bit more intuitive to me. First you validate, then you trace where things actually came from, and finally you query.

README.md
```diff
@@ -1,47 +1,58 @@
...
-- **extensible** → create custom plug-ins based on the Django ORM
+- **distributed** → dedicated zero-copy and lineage-aware data sharing across infrastructure (databases & storage locations)
+- **integrations** → [git](https://docs.lamin.ai/track#sync-code-with-git), [nextflow](https://docs.lamin.ai/nextflow), [vitessce](https://docs.lamin.ai/vitessce), [redun](https://docs.lamin.ai/redun), and [more](https://docs.lamin.ai/integrations)
+- **extensible** → create custom plug-ins based on the Django ORM, the basis for LaminDB's registries
+
+GUI?
 
-If you want a GUI: [LaminHub](https://lamin.ai) is a data collaboration hub built on LaminDB similar to how GitHub is built on git.
+- [LaminHub](https://lamin.ai) is a data collaboration hub built on LaminDB similar to how GitHub is built on git.
+- It's easy to vibe code on top of the open-source stack. ☺️
```
Zethson:
```suggestion
```

Not a fan of emojis and not a fan of the statement either but that's just my opinion.

README.md
falexwolf: I see why you want this and I was going back and forth.

But isn't it important to mention traceability and reproducibility? These are all related but still distinct. Depending on who I talk to "reproducibility" (e.g. in notebooks) and "traceability" (in general) are the big catch words.

There is no common use of an "understandability" term: that's really high-level land.

I wish it was shorter but I'm worried that the important signal from "reproducing & tracing" is missed.

README.md
falexwolf: The focus is on the compounding nature of data & operations. And I think "vast compute and intelligence" is adequate today.

So, I'd prefer to keep as is.

README.md
falexwolf: It makes three statements:

1. Queries
2. Validated
3. Unified access

All of these three are necessary to scale models because if these aren't provided "it leads to **garbage in, garbage out** or is quite simply impossible"

> Then the order is also off.

The order in the opener is the way it is because people first worry about "queryability" (or more broadly one would say accessibility/findability but we want to stick with queriability because it's stronger and more differentiated from any searchable system).

Once they can access/query data some (but not all) people worry about whether it's validated.

And last realize that "to get that context from that LIMS I also have to pull in data from there" and that might not be possible because tool fragmentation is hard/impossible to scale even if the LIMS comes with MCP.

The second part of the sentence describes two failure modes: "it leads to **garbage in, garbage out** or is quite simply impossible"

Swapping the order is not possible stylistically, I fear. And "inability to query" or "non-unified access" both lead to the more dramatic "it simply doesn't work".

---

I think we'd lose quite a bit of signal with what you're suggesting, in particular there'd be no relation to "queryability" anymore, which is the first thing we mention everywhere.

README.md
falexwolf: > IDK maybe it's just me but I feel like it's too big of an intellectual jump here that git & PRs are the history & queryablity that you mentioned above and that this has the potential to verify agent quality.

The pull requests are a new aspect that we can debate whether it should appear here. But I think the fact that git stores a history of your code-like data so that you can trace back all changes commonly resonates and really is the same value that you get from full traceability of all changes to your lakehouse.

That hook has, IMO, worked for 4 years. It's always been that analogy.

If you have a suggestion to make it smoother/clearer: great! I can see how removing "pull requests" can help but I'd like to hint at branching which is important to then arrive at governance / change management.

README.md
falexwolf: The order currently starts with "query" because of this

> The order in the opener is the way it is because people first worry about "queryability" (or more broadly one would say accessibility/findability but we want to stick with queriability because it's stronger and more differentiated from any searchable system).

This is also on the landing page and in the top tagline above the "Why?" section.

image

I think you're coming from the "order of operations" (you gotta validate & annotate to query) rather from the "value hierarchy" where IMO query needs to be mentioned first. Or rather "validating data sounds like work" and "querying data" sounds like value.

README.md
falexwolf: Why?

Zethson: This is very much a personal opinion but I think that emojis are unprofessional. I also think that vibe coding has a negative connotation and suggesting it is unprofessional.

But just my 2cts

README.md
Zethson: > I think you're coming from the "order of operations" (you gotta validate & annotate to query) rather from the "value hierarchy" where IMO query needs to be mentioned first. Or rather "validating data sounds like work" and "querying data" sounds like value.

Correct.

README.md
falexwolf: Agreed. Will remove!

README.md
```diff
@@ -1,47 +1,56 @@
...
-Highlights:
+How?
 
 - **lineage** → track inputs & outputs of notebooks, scripts, functions & pipelines with a single line of code
-- **lakehouse** → manage, monitor & validate schemas; query across many datasets
+- **lakehouse** → manage, monitor & validate schemas for standard and bio-formats; query across many datasets
 - **feature store** → manage features & labels; leverage batch loading
 - **FAIR datasets** → validate & annotate `DataFrame`, `AnnData`, `SpatialData`, `parquet`, `zarr`, …
-- **LIMS & ELN** → manage experimental metadata, ontologies & markdown notes
+- **LIMS & ELN** → manage bio-registries, ontologies & markdown notes
```
Koncopd: is "bio-registry" a well-known term?

falexwolf: Yes. It's also on our landing page at lamin.ai

README.md
```diff
@@ -1,47 +1,56 @@
...
+You get context & memory through a lineage-native lakehouse that understands bio-formats, registries & ontologies.
 
 <details>
 <summary>Why?</summary>
 
-Reproducing and understanding how datasets, models, and workflows were created is crucial to high-quality R&D — especially as agents increasingly contribute to it.
-At the same time, training models across thousands of datasets — from LIMS and ELNs to orthogonal assays and cross-team silos — is now a major learning opportunity, but requires queryable & validated data.
-Robustly scaled learning operations need something biology has lacked: an API-first data management framework comparable to git for code or warehouses for tables.
+(1) **Reproducing, tracing & understanding** how datasets, models & results are created is critical to quality R&D.
+Without **context**, agents & humans make mistakes and **cannot close feedback loops** across data generation & analysis.
```
Koncopd: i would put "humans & agents", in this order.

README.md
Koncopd: I actually like Lukas’s variant more.

README.md
falexwolf: You're right! Fixing.

README.md
falexwolf: I do understand why! It's plain and simple (which is also what I prefer) but it misses IMO key information.

I could remove the "vast" because it's IMO not 100% necessary even though idiomatic. The imbalance is a hallmark for biological data.

While scaling laws hold in every domain where vast amounts of data existed and compute is the bottleneck, things are upside down in biology (except for sequence data and their simple properties where vast amounts indeed exist).

sunnyosun:
I think it in general looks good to me, just feel like there are a bit too many bullets here, can some of them be merged?

I think the `simple` one can probably be removed? and "reproducible" can be part of "lineage" or "FAIR"?
image

falexwolf:
> I think it in general looks good to me, just feel like there are a bit too many bullets here, can some of them be merged?

I also wish there were less bullets. We can merge "lakehouse & feature store", because it's more or less the same thing for us.

falexwolf:
Removed it. I don't think anything else can be removed.

README.md
```diff
@@ -1,49 +1,65 @@
...
+You get context & memory through a lineage-native lakehouse that understands bio-formats, registries & ontologies.
 
 <details>
 <summary>Why?</summary>
 
-Reproducing and understanding how datasets, models, and workflows were created is crucial to high-quality R&D — especially as agents increasingly contribute to it.
-At the same time, training models across thousands of datasets — from LIMS and ELNs to orthogonal assays and cross-team silos — is now a major learning opportunity, but requires queryable & validated data.
-Robustly scaled learning operations need something biology has lacked: an API-first data management framework comparable to git for code or warehouses for tables.
+(1) **Reproducing, tracing & understanding** how datasets, models & results are created is critical to quality R&D.
+Without **context**, humans & agents make mistakes and **cannot close feedback loops** across data generation & analysis.
```
namsaraeva, line 10:
```suggestion
Humans & agents are mistake-prone and **cannot close feedback loops** across data generation & analysis.
```

So that we don't repeat "without".

README.md
namsaraeva: I also like @Zethson's suggestion more, but if @falexwolf wants to keep his wording, I would definitely get rid of "vast". It reads much better without.

README.md
```diff
@@ -1,49 +1,65 @@
...
+Without **memory**, vast compute and intelligence are wasted on fragmented, non-compounding tasks.
 
-LaminDB fills the gap with a lineage-native data lakehouse that understands bio-registries and formats (`AnnData`, `.zarr`, …).
-It provides queries across many datasets with enough freedom to maintain high-paced R&D while automating rich context on top of versioning, change management, and other industry standards.
+(2) Training & fine-tuning models with thousands of datasets — across LIMS, ELNs, orthogonal assays — is now a primary path to **scaling R&D**.
+But without **queryable & validated data** or with data locked in organizational & infrastructure siloes, it leads to **garbage in, garbage out** or is quite simply impossible.
```
namsaraeva, line 14:
```suggestion
But without **queryable & validated data** or with data locked in organizational & infrastructure siloes, it leads to **unreliable results** or is quite simply impossible.
```

I find garbage in garbage out a bit aggressive.

README.md
```diff
@@ -1,49 +1,65 @@
...
+
+Imagine building software without git or pull requests: an agent's quality would be impossible to **verify**.
+While code has git and tables have warehouses, biological data has lacked an open-source framework for managing its unique complexity.
```
namsaraeva: Maybe add a couple of words on "unique complexity"? Something along the lines of different modalities, rich metadata etc.

README.md
namsaraeva: I agree with @Zethson here, too. I immediately thought of the "order of operations", but we might be biased.

README.md
falexwolf: The repetition is on purpose but I'm not attached to it.

But "context" needs to stay in the sentence.

README.md
falexwolf: Agreed on "vast" being unnecessary fluff. Removed.

README.md
falexwolf: Yes, I can see that. But don't have a better idiom. I can confirm however that industry & academic leadership uses this regularly. I didn't have this in my early drafts but Gemini suggested it and I thought it's punchy.

README.md
falexwolf: Yes, and no. We've had descriptions of that complexity for years on the docs and e.g. also here on the "seminal" blog post: https://blog.lamin.ai/problems

But after talking to people for so long I've heard so many times that "yeah, everybody knows metadata is a mess but nobody has a concrete solution for solving it" that it started to feel like unnecessary words to re-describe the specifics of the mess. This is why here I'd rather jump to the solution (the How? section and the API docs).

README.md
falexwolf: I think you're biased because you know how this stuff works ☺️

falexwolf:
Thank you all for the valuable feedback, @Zethson @Koncopd @sunnyosun @namsaraeva @fredericenard!

I'll merge it as is now and keep these two simplification ideas for the first paragraph under the Why? section in mind. Reading it with a few nights sleep in between might change perception.

PR #3476, opened 2026-01-18
🏗️ Lazy import proxy for `lamindb.models`

falexwolf:
See:

- https://github.com/laminlabs/lamindb/pull/2849

PR #3525, opened 2026-02-06
⚡ Speed-up lamindb import by lazily importing pandas, anndata & pyarrow

falexwolf:
Analysis: https://lamin.ai/laminlabs/lamindata/transform/77UArcsm8SY50006

The improvement is roughly 60% (1.2 sec vs. 3 sec on CI, 0.8 sec vs. 2 sec on my laptop) auto-connect to `laminlabs/lamindata`.

image

Before | After
--- | ---
https://github.com/laminlabs/lamindb/actions/runs/21747544115/job/62737070942 | https://github.com/laminlabs/lamindb/actions/runs/21787333086/job/62860826308?pr=3525
image | image

Already this a first optimization is a big deal for any CLI operation that relies on lamindb. But even for testing, scripting, running ETL-type workflows and notebooks it matters.

More performance and "start-up" optimization to follow.

Needs:

- https://github.com/laminlabs/bionty/pull/364

Auto-complete was fixed here:

- https://github.com/laminlabs/lamindb/pull/3534

### DIFF
--- a/tests/storage/test_streaming.py
+++ b/tests/storage/test_streaming.py
@@ -256,6 +256,7 @@ def test_anndata_open_mode():
     artifact = ln.Artifact(fp, key="test_adata.h5ad").save()
 
     with artifact.open(mode="r") as access:
+        # TODO: add back proper type checking once reset_django() is gone
         assert type(access).__name__ == "AnnDataAccessor"
         assert type(access).__module__ == "lamindb.core.storage._anndata_accessor"
     # can't open in write mode if not tiledbsoma


lamindb/models/save.py
```diff
@@ -32,6 +25,32 @@
 
     from .artifact import Artifact
 
+_storage_paths_cache = None
+
+
+def _storage_paths():
```
Koncopd: this really looks bad to me. Such cumbersome blocks of code to import a few things. And the usage of this is cumbersome.

falexwolf: Yes, this is an early version of the refactor. I just wanted to give it a try to see how complex it would be to remove pandas, anndata, pyarrow, etc.

I don't think any of these packages should be imported in a database client like lamindb and with a bit more work I think this should be doable.

No worries, this won't be merged in the current form.

I'll tinker with it a bit more during the weekend.

falexwolf: I've stared at this a bit longer and now think it's a nice way of scoping this `storage.paths` module.

If we didn't have the concern of `core.storage.paths` being "heavy", we could have done the same thing, just without caching:

```
from ..core.storage import paths as storage_paths 
```

This is nicer than importing all these helper functions at the root namespace IMO.

And what's done here now is a very simple way of an analogous `storage_paths` namespace in a cached way.

And in Python 3.7 there will be lazy modules and in a few ways one wouldn't need to worry about this anymore.

falexwolf:
At this point in the PR we reduced the import time from around 3 sec to 1.2 sec

Before | After
--- | ---
https://github.com/laminlabs/lamindb/actions/runs/21747544115/job/62737070942 | https://github.com/laminlabs/lamindb/actions/runs/21787333086/job/62860826308?pr=3525
image | image

This is a **huge** deal for any CLI operation that relies on lamindb.

But even for testing, scripting, running workflows and notebooks it's UX that changes quality of life and influences perception of the quality of the package.

---

The strategies used to ensure the lazy import in this PR are not yet all polished and 100% consistent. There is potentiall for making this very pretty.

Right now I'd say it's absolutely worth it to optimize the import time even further although this PR here would be the biggest win.

lamindb/core/storage/types.py
```diff
@@ -13,7 +12,12 @@
...
 
-MuData = TypeVar("MuData")
-SpatialData = TypeVar("SpatialData")
+if TYPE_CHECKING:
+    from anndata import AnnData
+    from mudata import MuData
+    from spatialdata import SpatialData
 
-ScverseDataStructures = AnnData | MuData | SpatialData
+    ScverseDataStructures = AnnData | MuData | SpatialData
```
falexwolf, line 12: Here is the full type @Zethson.

lamindb/core/storage/types.py
```diff
@@ -13,7 +12,12 @@
...
+else:
+    # AnnData | MuData | SpatialData; Any required for union with DataFrame in objects.py
+    ScverseDataStructures = Any
```
falexwolf, line 15: I'm still not quite sure why this would be necessary. Needs more thought.

lamindb/core/storage/types.py
Zethson: Yes I saw but this means that only when typechecking is the full type accessible. In other words, autocomplete and stuff probably won't be as smart?

lamindb/core/loaders.py
```diff
@@ -57,14 +59,18 @@ def load_fcs(*args, **kwargs) -> AnnData:
     return readfcs.read(*args, **kwargs)
 
 
-def load_tsv(path: UPathStr, **kwargs) -> pd.DataFrame:
```
Koncopd, line 60: why was the return type removed here?

lamindb/core/loaders.py
```diff
@@ -196,7 +238,7 @@ def load_to_memory(
     """
     filepath = create_path(filepath)
     suffix = extract_suffix_from_path(filepath)
-    loader = FILE_LOADERS.get(suffix, None)
+    loader = _get_file_loaders().get(suffix, None)
```
Koncopd: Maybe a class with `__gettattr__` instead of a function for such things?

lamindb/models/artifact.py
```diff
@@ -92,21 +67,73 @@
 from .storage import Storage
 from .ulabel import ULabel
 
+
+def _storage():
```
Koncopd: It is annoying that each module implements the same pattern again and again. Can we maybe centralize this somehow?

lamindb/models/save.py
```diff
@@ -380,7 +401,7 @@ def check_and_attempt_clearing(
     if hasattr(artifact, "_clear_storagekey"):
         try:
             if artifact._clear_storagekey is not None:  # type: ignore
-                delete_msg = delete_storage_using_key(
+                delete_msg = _storage_paths().delete_storage_using_key(
```
Koncopd: I don't like that this has to be called for access every time, just looks ugly, no?

falexwolf:
Thanks for review, @Koncopd. I wanted to address most of the things anyway by polishing this PR - just hadn't yet get around it. 😅 

Will come back here in just a bit!

lamindb/core/loaders.py
falexwolf: just an agent mistake

lamindb/core/loaders.py
falexwolf: I think the right design is that the whole loaders module is not eagerly loaded given it's a mere utility.

lamindb/models/artifact.py
falexwolf: Agreed.

lamindb/models/save.py
falexwolf: Yes, this should IMO look more like a module- or class-namespace, not like a function call. Will think how to refactor it.

If it looked like a "Storage class" or a "storage module", then I think it'd be OK.

lamindb/core/storage/types.py
falexwolf: Hmm, given ruff's behavior is that any type annotation that's _just a type annotation_ is moved under TYPE_CHECKING and the consequence being that many/most of our type annotations are actually under TYPE_CHECKING I think you're wrong.

"autocomplete and stuff" would not work for most of our type annotations if that were true.

This being said: I haven't paid close attention.

And also this being said: I don't like the `Any` alternative and I think this should be dealt with so that `Any` is avoided (https://github.com/laminlabs/lamindb/pull/3525#discussion_r2779508570).

lamindb/core/storage/types.py
falexwolf: One way of doing this is by moving this type into the `storage` module where I think it'd actually belong.

lamindb/core/loaders.py
falexwolf: restored the whole `loaders.py` from main and instead loading the whole module lazily instead

tests/storage/test_streaming.py
```diff
@@ -701,7 +495,7 @@ def test_compressed(gz_suffix):
     assert artifact.n_observations == 30
 
     with artifact.open() as store:
-        assert isinstance(store, AnnDataAccessor)
+        assert type(store).__name__ == "AnnDataAccessor"
```
Koncopd, line 504: why doesn't `isinstance` work?

falexwolf: That is a good question. It works for me locally, but gave issues on CI.

falexwolf:
@Koncopd @Zethson - at this point I think the code is better and evidently the package import time is much better opening all the opportunities discussed on Slack.

What looks "ugly" is the datasets module, but that should be thoughtfully re-done anyway; so I don't think it's worthwhile investing into a sophisticated scheme.

Please let me know if you find something that still concerns you!

Note: I just notice that I'm not quite clear how we handled the SpatialData type before given this was always lazy.

lamindb/core/storage/types.py
Zethson: Do you want to merge the PR with this? Have you checked autocomplete as a user?

lamindb/core/__init__.py
```diff
@@ -36,9 +36,30 @@
...
-from . import loaders, subsettings, types
+from . import subsettings
 from ._context import Context
-from ._mapped_collection import MappedCollection
 from ._settings import Settings
+
+
+def __getattr__(name: str):
+    # need to lazy import a few auxliary modules to maintain
+    # backward compat
```
Zethson: Nit: Please move this a line up

tests/storage/test_streaming.py
Koncopd: but then it has to be investigated properly and resolved instead of adding this hack, don't you agree?

lamindb/models/artifact.py
```diff
@@ -92,21 +69,70 @@
...
+    )
+
+
+# Cache the storage utils on first use
+_storage_cache: object | None = None
+
+
+# refactor this module to group logic that needs storage access in a class
+# in the future; then we don't need _s() anymore
+def _s():
```
Zethson, line 108: Having this be a shorter name doesn't mean that it doesn't exist anymore haha. Could we please consider giving this a longer and descriptive name? I prefer explicit over short/hidden.

Koncopd:
I have not checked the last changes yet but my lingering concern is the `isinstance` problem.

Zethson (approved):
Still have my reservations but let's go

tests/storage/test_streaming.py
Koncopd: we had multiple cases where such errors indicated genuine problems.

falexwolf: I think that in this case it should be alright because no use case ever will do this check.

> but then it has to be investigated properly and resolved instead of adding this hack, don't you agree?

Yes, of course, but I need to trade all the other things that we don't do during that time. Blocking the whole PR because of this detail seems exaggerated.

> we had multiple cases where such errors indicated genuine problems.

I think we have a single case and that's the Django reset problem which is a much bigger problem that's unresolved and should deserve the time that would otherwise go into this

This PR sets the basis for me investing time into the "re-connect django instead of reset django" problem again that'd eliminate the stale python class problem completely

This being said: somehow the lazy import mechanism seems to duplicate the AnnDataAccessor reference - I expect us to fix this sooner or later but I don't see risk from it; I'd much rather focus on the important problems

falexwolf:
> I have not checked the last changes yet but my lingering concern is the isinstance problem.

Responded rather extensively.

lamindb/core/storage/types.py
falexwolf: I have not, but see: https://github.com/laminlabs/lamindb/pull/3525#discussion_r2781858810

We're already using TYPE_CHECKING for most type hints.

Can you speak to why this ever worked?

> Note: I just notice that I'm not quite clear how we handled the SpatialData type before given this was always lazy.

We never had SpatialData in a non-lazy way because the package doesn't rely on it.

falexwolf: I now remember how you did it, @Zethson:

```python
MuData = TypeVar("MuData")
SpatialData = TypeVar("SpatialData")
ScverseDataStructures = AnnData | MuData | SpatialData
```

I had a brief conversation with AI on Friday night and it was adamant that this was worse than the `Any` and TYPE_CHECKING solution now.

falexwolf: Started a dedicated prompt on this:

I'm wondering about how to type classes that aren't part of the package dependencies



This here is one suggestion

```python
from __future__ import annotations
from typing import TypeVar
from anndata import AnnData
MuData = TypeVar("MuData")
SpatialData = TypeVar("SpatialData")
ScverseDataStructures = AnnData | MuData | SpatialData
```


These here is another suggestion

```python
from __future__ import annotations
from typing import TYPE_CHECKING, Any
if TYPE_CHECKING:
  from anndata import AnnData
  from mudata import MuData
  from spatialdata import SpatialData
  ScverseDataStructures = AnnData | MuData | SpatialData
else:
  # AnnData | MuData | SpatialData; Any required for union with DataFrame in objects.py
  ScverseDataStructures = Any
```

This type is located in `lamindb.core.storage` and then used in different parts of the module. Besides mypy type checking there is also a concern of auto-complete and IDE experience.

Here is the explanation why the previous solution was worse according to Gemini:

image

Now I'm curious whether Gemini will also suggest the string-based types that were already in this PR on Friday.

falexwolf: Ok, Gemini suggest to keep as is because

image

image

tests/storage/test_streaming.py
Koncopd: Sorry, i can't really agree here.

>somehow the lazy import mechanism seems to duplicate the AnnDataAccessor reference

This could very much mean that caching is just broken. The cause of this has to be understood.

lamindb/core/__init__.py
```diff
@@ -36,9 +36,30 @@
...
+    # none of them should have been eagerly imported in the first place
+    if name == "loaders":
+        import importlib
+
+        loaders = importlib.import_module(".loaders", package=__name__)
+        return loaders
+    if name == "storage":
+        import importlib
+
+        storage = importlib.import_module(".storage", package=__name__)
```
Koncopd, line 57: why `importlib.import_module` here? why not just `from . import storage`? This can be related to the problem with `isinstance`.

tests/storage/test_streaming.py
Koncopd: Having different classes on one import is a genuine problem in my opinion, it can cause problems in random places in the future.

Koncopd: i think 
`from .core.storage._anndata_accessor import AnnDataAccessor`
gives an object different in id from
`from .core.storage import AnnDataAccessor`
or something like that.

Koncopd: This is more problematic than you think because that would probably mean we also have two or more `AccessRegistry` objects and so on.

Koncopd:
Thank you for the response, but i can't agree that this is not a problem. Commented.

tests/storage/test_streaming.py
falexwolf: Ok, am investigating.

falexwolf:
Addressed in commit. Hopefully this will reveal what the problem is.

lamindb/core/__init__.py
falexwolf: Good point.

tests/storage/test_streaming.py
```diff
@@ -701,7 +508,11 @@ def test_compressed(gz_suffix):
     assert artifact.n_observations == 30
 
     with artifact.open() as store:
-        assert isinstance(store, AnnDataAccessor)
+        assert isinstance(store, AnnDataAccessor), (
+            f"Expected AnnDataAccessor, got {type(store).__name__}"
+            f" (module: {type(store).__module__}, expected from: {AnnDataAccessor.__module__},"
```
Koncopd: No, in that sense they are clearly the same if check is only done on names. I would comapre ids also: `id(AnnDataAccessor)` and `id(type(store))`. And also modules `AnnDataAccessor.__module__` and `type(store).__module__`, not by name, but via `is` or `id`.

falexwolf: Yes, evidently the ids are different but they don't help with debugging.

Maybe the next step helps.

Now is the first time that the agent is onto a real explanation:

image

falexwolf: Yes, evidently the ids are different but they don't help with debugging.

Maybe the next step helps.

Now is the first time that the agent is onto a real explanation:

image

falexwolf: Yes, evidently the ids are different but they don't help with debugging.

Maybe the next step helps.

Now is the first time that the agent is onto a real explanation:

image

tests/storage/conftest.py
```diff
@@ -4,6 +4,10 @@
 from time import perf_counter
 
 import lamindb as ln
+
+# Force storage module to load first to ensure consistent class identity
+# (avoids isinstance failure when storage is lazy-loaded via different paths)
+import lamindb.core.storage  # noqa: F401
```
Koncopd: But we care not only about this working correctly in our tests witht this specific setup, but in general, no?

falexwolf: But if this fixes it we know what the problem is.

falexwolf: I so far can't reproduce any of these problems locally, that's why I'm experimenting via CI

falexwolf:
I think the problem is just reset_django().

image

The code in this PR is fine. `reset_django()` is not.

Koncopd:
But we don't switch instances here.

falexwolf:
> But we don't switch instances here.

That doesn't matter. The only thing that matters is whether lamindb is imported (it leads to Django being setup against the in-memory database) and subsequently `init()` or `connect()` is called within the same Python process (which leads to django being reset).

Here is the import:

https://github.com/laminlabs/lamindb/blob/136055174daa72ee131654d061f982502f0ec25e/tests/storage/conftest.py#L6

That's why I'm so adamant about eliminating "reset_django()" entirely. It's the one piece of black magic in the code base.

falexwolf:
Let me convince you.

pyproject.toml
```diff
@@ -57,6 +57,7 @@ fcs = [
 ]
 dev = [
     # basic test
+    "setuptools>=81,<82",  # 82 removes pkg_resources; xarray-schema (spatialdata dep) still needs it
```
Koncopd: i guess this is not needed anymore.

falexwolf: yep

falexwolf:
Ok, just switching the import order in conftest.py is not enough because pytest collects all test files first.

Koncopd:
Even dynamic import was not enough.

falexwolf:
> Even dynamic import was not enough.

It was enough for the `AnnDataAccessor` case, but for some reason the `BackedAccessor` still failed. Of course one could now investigate what's specific to `BackedAccessor` but IMO this time is better spent eliminating `reset_django()` via:

- https://github.com/laminlabs/lamindb-setup/pull/1231

I'm now pretty convinced I can make this work by making the instance modules part of the local environment config; exactly like they are when configuring an environment default instance.

This would be low-complexity, would be perfectly robust, and would also speed-up everything. The only caveat with it is that it'll sometimes hard-error if the user wants to do something with modules that they hadn't configured in the environment. But that's, I think, OK, because it's not frequent.

I also spent time on this but think it's much more complicated and also has UX problems. And I don't even know whether I could get it to work in finite time:

- https://github.com/laminlabs/lamindb/pull/3476

Koncopd (approved):
Fine for me other than failing `isinstance` tests.

falexwolf:
@Koncopd & I just had a long call about the path to eliminating `reset_django()`. I think he is now convinced that with constraining the modules a robust & performant solution is possible: https://github.com/laminlabs/lamindb-setup/pull/1231#issuecomment-3629092892

The topic is overall more complicated because there is also the Lambda topic, but I'm convinced that we will get this under control.

---

One idea would be to keep this PR open until `reset_django()` is eliminated. The one reason I don't like it is because we'll have conflicts by the time.

So, I'll merge it now and promise that I'll fix `reset_django()` and with that the `AnnDataAccessor` class identity issue.

PR #3528, opened 2026-02-11
🐛 Scope feature query by schema member types in curator

Zethson:
```python
type_a = ln.Feature(name="TypeA", is_type=True, description="Type A features").save()
type_b = ln.Feature(name="TypeB", is_type=True, description="Type B features").save()

ln.Feature(name="assay name", type=type_a, dtype=str).save()
ln.Feature(name="assay name", type=type_b, dtype=str).save()

schema = ln.Schema(
    name="schema_a",
    features=[
        ln.Feature.get(name="assay name", type=type_a)
    ],
    flexible=True,
    otype="DataFrame",
).save()

df = pd.DataFrame({"assay name": ["exp1", "exp2"]})
tmp = Path(tempfile.mkdtemp()) / "test.tsv"
df.to_csv(tmp, sep="\t", index=False)

a = ln.Artifact(
    path=tmp,
    key="repro/test.tsv"
).save()
a.otype = "DataFrame"

curator = ln.curators.DataFrameCurator(a, schema)
curator.save_artifact()
```

used to throw `ValidationError: Ambiguous match for Feature 'assay name': found 2 features at depth 1 (under types: ['TypeA', 'TypeB'])`.

When validating columns for a flexible schema, the "columns" CatVector queries all features by name, causing ambiguous matches when the same feature name exists under different types. This is fixed by passing the Schema to CatVector and filtering the query to only feature types present in the Schema's members.

### DIFF
--- a/lamindb/curators/core.py
+++ b/lamindb/curators/core.py
@@ -19,6 +19,7 @@ import lamindb_setup as ln_setup
 import numpy as np
 import pandas as pd
 import pandera.pandas as pandera
+from django.db.models import Q
 from lamin_utils import colors, logger
 from lamindb_setup.core._docs import doc_args
 from lamindb_setup.core.upath import LocalPathClasses
@@ -1538,7 +1539,7 @@ class CatVector:
             if issubclass(registry, HasType):
                 if self._type_record is None:
                     # When we have a Schema with typed members,
-                    # scope the query to only the types present in the schema's members
+                    # scope the query to the types present in the schema's members (plus untyped features)
                     # to avoid ambiguous matches across different feature types.
                     qs = registry.filter()
                     if self._schema and self._schema.n_members:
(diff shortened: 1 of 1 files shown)


lamindb/curators/core.py
```diff
@@ -1535,7 +1538,21 @@ def _add_validated(self) -> tuple[list, list]:
...
+                        type_ids = {
+                            m.type_id
+                            for m in self._schema.members
+                            if m.type_id is not None
+                        }
+                        if type_ids:
+                            qs = registry.filter(
+                                Q(type_id__in=type_ids) | Q(type_id__isnull=True)
+                            )
+                    self._subtype_query_set = qs
```
Koncopd, line 1555: @Zethson where is this used?

Zethson: Just below what you're looking at:

```python
values_array = np.array(remaining_values)
                validated_mask = self._subtype_query_set.validate(
                    values_array, field=field, mute=True
                )
                validated_values, non_validated_values = (
                    list(values_array[validated_mask]),
                    list(values_array[~validated_mask]),
                )
                records = self._subtype_query_set.filter(
                    **{f"{field_name}__in": validated_values}, **filter_kwargs
                ).to_list()
                records = keep_topmost_matches(records)
```

if I understand your question correctly.

Koncopd: ok, yes, thanks

PR #3530, opened 2026-02-11
🐛 Fix DB snapshot sync

Zethson:
We must always sync the `lamindb.gz` file.

### DIFF
--- a/lamindb/models/sqlrecord.py
+++ b/lamindb/models/sqlrecord.py
@@ -522,9 +522,11 @@ def _synchronize_clone(storage_root: str) -> str | None:
         if cloud_db_path_gz.synchronize_to(
             local_sqlite_path_gz, error_no_origin=True, print_progress=True
         ):
-            with gzip.open(local_sqlite_path_gz, "rb") as f_in:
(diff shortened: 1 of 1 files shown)


lamindb/models/sqlrecord.py
```diff
@@ -514,20 +514,17 @@ def _synchronize_clone(storage_root: str) -> str | None:
...
-        cloud_db_path_gz.synchronize_to(
+        if cloud_db_path_gz.synchronize_to(
             local_sqlite_path_gz, error_no_origin=True, print_progress=True
-        )
-        with gzip.open(local_sqlite_path_gz, "rb") as f_in:
-            with open(local_sqlite_path, "wb") as f_out:
-                shutil.copyfileobj(f_in, f_out)
+        ):
+            with gzip.open(local_sqlite_path_gz, "rb") as f_in:
+                with open(local_sqlite_path, "wb") as f_out:
```
Koncopd: btw you can combine both into one `with` statement.

Zethson: Never done this actually but yeah

PR #3542, opened 2026-02-17
🗃️ Add a `_status_code: int` field to `Branch` and `BaseBlock`

falexwolf:
Needs:

- https://github.com/laminlabs/lamindb-setup/pull/1293

### DIFF
--- a/lamindb/migrations/0182_squashed.py
+++ b/lamindb/migrations/0182_squashed.py
@@ -190,6 +190,7 @@ class Migration(migrations.Migration):
         ("lamindb", "0179_v2_2_part_2"),
         ("lamindb", "0180_v2_2_part_3"),
         ("lamindb", "0181_v2_2_part_4"),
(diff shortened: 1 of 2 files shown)


Koncopd:
`created_by` with `current_user_id()` as a default value is problematic due to this overwrite
https://github.com/laminlabs/lamindb/blob/1cf9fab1f635eba720ed78206734f78f6cdbae97/lamindb/base/users.py#L26
The error happens during the migration on init of an instance before the user is registered.

At least this is what i see here https://github.com/laminlabs/lamindb/actions/runs/22115054047/job/63920886700?pr=3542

falexwolf:
> created_by with current_user_id() as a default value is problematic due to this overwrite

We have it everywhere, so this can't be a new problem, can it?

I'm currently trying to understand how it happens all of a sudden.

Koncopd:
No, this is new because before the default value was `None`.

PR #3550, opened 2026-02-18
Relax AnnData version constraint

Zethson:
A theislab user wanted to use lamindb with a more recent anndata version but was met with installation errors.


Koncopd:
No, please, could you just upper bound with the actual latest version please.

Koncopd:
Why bumping the lower bound though?

Zethson:
> Why bumping the lower bound though?

We had wanted to do that for a long time already (there's a slack thread on it) and by now, I actually fear that some users might use lamindb with an old version that we absolutely haven't tested in ages.

Koncopd:
Ok, got it, thank you.

PR #3628, opened 2026-04-03
changes in _finish.py to ensure tracking of large notebooks

ishitajain9717:
I think the issue was the following:
On big notebooks, save-to-disk can take a while. ln.finish() only waited for 30s for the file to look just saved, and on the second finish() after save and run again, if it still didn’t see a fresh file it could keep going, clear the in-memory run, and the next finish() threw TrackNotCalled even though we’d already run track()
What I tried, that might work:
Extended that “wait for save” loop (30 → 300 iterations, 5 min) so large .ipynb writes have time to finish.
On the retry path for Python notebooks, if the file still isn’t seen as saved, we now return "retry" instead of falling through — so we don’t drop the active run in that bad state.

### DIFF
--- a/lamindb/_finish.py
+++ b/lamindb/_finish.py
@@ -213,7 +213,8 @@ def check_filepath_recently_saved(filepath: Path, is_finish_retry: bool) -> bool
     # to upload of an outdated notebook
     # also see implementation for R notebooks below
     offset_saved_time = 0.3 if not is_finish_retry else 20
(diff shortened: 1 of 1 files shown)


ishitajain9717:
This was not an issue with lamin, but with laminR. It is being discussed here https://github.com/laminlabs/laminr/issues/262, because this issue from pfizer occured when using R kernel in Jupyter notebook, with laminR

PR #3659, opened 2026-04-22
🔒️ Obfuscate user secrets in params and source code

falexwolf:
Addresses:

- https://github.com/pfizer-collab/laminlabs/issues/835


Koncopd:
It is not clear to me if this hides also credentials in Postgres connection strings.

falexwolf:
> It is not clear to me if this hides also credentials in Postgres connection strings.

Checking. I believe we'll need to slowly expand test cases here.

PR #3666, opened 2026-04-28
📝 Better document collection vs folder and clarify that `ArtifactSet` can also be streamed

falexwolf:


docs/arrays.md
```diff
@@ -19,52 +19,60 @@ import lamindb as ln
...
 
-It is also possible to open a collection of cloud artifacts.
+### Streaming from a set of artifacts
 
-```python
-collection = ln.Collection.connect("laminlabs/lamindata").get(
-    key="sharded_parquet_collection"
-)
-```
+You can open several parquet files as a single dataset by calling `.open()` on the result of a query:
```
falexwolf, line 55: @Koncopd, this was very hard to discover in the docs. I wonder whether we should also some cross-reference it under `Artifact.open()`?

This triggered, I think, this issue:

- https://github.com/pfizer-collab/laminlabs/issues/893

PR #3673, opened 2026-05-05
✨ Move artifact to a new storage location on key change

Koncopd:
This PR enables updating `artifact.key` in a controlled way inside `artifact.save()` for non-virtual keys.

Example:
```python
import lamindb as ln
# Non-virtual artifact (real storage path key)
artifact = ln.Artifact("local_file.csv", key="data/local_file.csv", _key_is_virtual=False).save()
# Rename key (will move underlying object and ask for confirmation)
artifact.key = "data/local_file_renamed.csv"
artifact.save() # confirm moving the artifact to the new path
```

### DIFF
--- a/lamindb/models/artifact.py
+++ b/lamindb/models/artifact.py
@@ -1313,7 +1313,7 @@ class Artifact(SQLRecord, IsVersioned, TracksRun, TracksUpdates):
             ),
         ]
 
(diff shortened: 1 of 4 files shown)


falexwolf:
How is versioning handled for real keys? We need bucket versioning switched on then, right?

PR #3676, opened 2026-05-06
:memo: Update `CONTRIBUTING.md`

sheetalgiri:
I tried to install based on instructions in [Contributing.md](https://github.com/laminlabs/lamindb/blob/main/CONTRIBUTING.md) where I had issues running the nox install through and with incompatible laminci versions, and made changes addressing this.


falexwolf (approved):
Looks good to me!

PR #3677, opened 2026-05-07
🚸 Improve UX of auto-selected managed storage locations for a space

falexwolf:
Strengthens instance-ownership safeguards for storage operations by requiring `artifacts` and settings storage targets to be managed by the current instance

### DIFF
--- a/lamindb/models/artifact.py
+++ b/lamindb/models/artifact.py
@@ -595,9 +595,6 @@ def get_artifact_kwargs_from_data(
         "key": key,
         "size": size,
         "storage_id": storage.id,
-        # passing both the id and the object
-        # to make them both available immediately
-        # after object creation
         "n_files": n_files,
         "_overwrite_versions": overwrite_versions,  # True for folder, False for file
         "n_observations": None,  # to implement
--- a/lamindb/models/artifact.py
+++ b/lamindb/models/artifact.py
@@ -3102,11 +3102,6 @@ class Artifact(SQLRecord, IsVersioned, TracksRun, TracksUpdates):
             logger.warning("you are saving to a non-latest version of the artifact")
 
         access_token = kwargs.pop("access_token", None)
-        if self.storage.instance_uid != ln_setup.settings.instance.uid:
-            root_as_str = self.storage.root
-            raise ValueError(
-                f"Storage {root_as_str} exists in another instance ({self.storage.instance_uid}), cannot write to it from here."
-            )
 
         if self._field_changed("key", check_is_saved=False):
             new_key = self.key
@@ -3132,6 +3127,11 @@ class Artifact(SQLRecord, IsVersioned, TracksRun, TracksUpdates):
                     raise InvalidArgument(
                         "Cannot update the key of an artifact in a storage location that is not managed by an instance."
                     )
+                if self.storage.instance_uid != ln_setup.settings.instance.uid:
+                    root_as_str = self.storage.root
+                    raise ValueError(
+                        f"Storage {root_as_str} exists in another instance ({self.storage.instance_uid}), cannot write to it from here."
+                    )
                 old_key = self._original_values["key"]
                 if old_key is None:
                     raise InvalidArgument(
@@ -3150,6 +3150,11 @@ class Artifact(SQLRecord, IsVersioned, TracksRun, TracksUpdates):
             if self.storage.instance_uid is None:
                 raise InvalidArgument(
                     "Cannot update the suffix of an artifact in a storage location that is not managed by an instance."
+                )
+            if self.storage.instance_uid != ln_setup.settings.instance.uid:
+                root_as_str = self.storage.root
+                raise ValueError(
+                    f"Storage {root_as_str} exists in another instance ({self.storage.instance_uid}), cannot write to it from here."
                 )
             if not _handle_suffix_change_on_save(self):
                 return None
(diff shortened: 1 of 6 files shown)


lamindb/models/artifact.py
```diff
@@ -1698,7 +1698,9 @@ def __init__(
                 logger.warning(
                     "storage argument ignored as storage information from space takes precedence"
                 )
-            storage_locs_for_space = Storage.filter(space=space)
+            storage_locs_for_space = Storage.filter(
+                space=space, instance_uid=setup_settings.instance.uid
+            ).order_by("id")
             n_storage_locs_for_space = len(storage_locs_for_space)
```
Koncopd: @falexwolf not related to this pr, but could make it `storage_locs_for_space.count()` here instead of `len(storage_locs_for_space)`?

lamindb/models/artifact.py
```diff
@@ -1698,7 +1698,9 @@ def __init__(
                 logger.warning(
                     "storage argument ignored as storage information from space takes precedence"
                 )
-            storage_locs_for_space = Storage.filter(space=space)
+            storage_locs_for_space = Storage.filter(
+                space=space, instance_uid=setup_settings.instance.uid
+            ).order_by("id")
```
Koncopd, line 1708: maybe sort by `created_at`?

lamindb/models/artifact.py
```diff
@@ -1709,8 +1711,15 @@ def __init__(
             else:
                 storage = storage_locs_for_space.first()
                 if n_storage_locs_for_space > 1:
+                    other_storage_locs = ",".join(
+                        f"{s.root}" for s in storage_locs_for_space[1:]
+                    )
                     logger.warning(
-                        f"more than one storage location for space {space}, choosing {storage}"
+                        f"more than one storage location is managed by this instance for space {space},\n"
+                        f"choosing root={storage.root}\n"
```
Koncopd, line 1724: maybe also say here that we give the latest storage? and sort by `created_at`?

lamindb/models/artifact.py
```diff
@@ -3093,6 +3102,11 @@ def save(
             logger.warning("you are saving to a non-latest version of the artifact")
 
         access_token = kwargs.pop("access_token", None)
+        if self.storage.instance_uid != ln_setup.settings.instance.uid:
+            root_as_str = self.storage.root
+            raise ValueError(
+                f"Storage {root_as_str} exists in another instance ({self.storage.instance_uid}), cannot write to it from here."
```
Koncopd: this is not really possible to check here, not all artifacts are actually uploaded, this breaks immediately for any artifact in existing storage that is not managed (like when saving a http link or link to a file in s3 or gs).

Koncopd: also breaks on transferring artifacts from one instance to another

lamindb/models/artifact.py
falexwolf: that's mathematically equivalent unless you have microsecond ties, isn't it?

lamindb/models/artifact.py
falexwolf: agreed

lamindb/models/artifact.py
falexwolf: I would keep the current code which gives the first-created storage location because this means that accidentally created new storage locations don't receive any objects

lamindb/models/artifact.py
falexwolf: > this is not really possible to check here, not all artifacts are actually uploaded, this breaks immediately for any artifact in existing storage that is not managed (like when saving a http link or link to a file in s3 or gs).

Right, but the check should be conditional on uploading something, right?

> also breaks on transferring artifacts from one instance to another

With storage copy, too, or just for zero-copy?

lamindb/models/artifact.py
Koncopd: yep, you are right

lamindb/models/artifact.py
Koncopd: > Right, but the check should be conditional on uploading something, right?

Yes, this should work.

>With storage copy, too, or just for zero-copy?

with zero-copy. Do we even have storage copy on transfer between instances? I don't think so.

lamindb/models/artifact.py
```diff
@@ -3215,6 +3234,15 @@ def save(
...
 
         flag_complete = has_local_filepath and getattr(self, "_to_store", False)
+        if (
+            flag_complete
+            and self.storage.instance_uid is not None
+            and self.storage.instance_uid != ln_setup.settings.instance.uid
+        ):
+            root_as_str = self.storage.root
+            raise ValueError(
+                f"Storage {root_as_str} exists in another instance ({self.storage.instance_uid}), cannot write to it from here."
```
Koncopd: i think this should be raised on init, not in save

lamindb/models/artifact.py
```diff
@@ -3137,6 +3151,11 @@ def save(
                 raise InvalidArgument(
                     "Cannot update the suffix of an artifact in a storage location that is not managed by an instance."
                 )
+            if self.storage.instance_uid != ln_setup.settings.instance.uid:
+                root_as_str = self.storage.root
+                raise ValueError(
+                    f"Storage {root_as_str} exists in another instance ({self.storage.instance_uid}), cannot write to it from here."
```
Koncopd: i think this error and the above is not correct to do like this. lets discuss tomorrow

lamindb/models/artifact.py
falexwolf: > Do we even have storage copy on transfer between instances? I don't think so.

I thought that's what you added: I mean it's just a cross-storage transfer. It doesn't matter whether it's across instances or not, does it?

lamindb/models/artifact.py
falexwolf: The problem is it could be mutated after init.  So, we'd need to raise both on init and on save. That's just more verbose/ more code.

lamindb/models/artifact.py
falexwolf: Ok!

PR #3680, opened 2026-05-11
✨ Compatibility of `ln.track()` and `ln.finish()` with Marimo notebooks

ishitajain9717:
This PR enables `ln.track()` and `ln.finish()` to work in Marimo notebooks in analogy to how they work in Jupyter notebooks.


lamindb/core/_context.py
```diff
@@ -391,11 +409,6 @@ def serialize_params_to_json(params: dict) -> dict:
             logger.warning(
                 f"skipping param {key} with value {value} and dtype {dtype} not JSON serializable"
             )
-            continue
-        if is_sensitive_param_key(key) or is_sensitive_param_value(
```
falexwolf, line 395: Why did you remove this, @ishitajain9717? It breaks the tests.

lamindb/core/_context.py
```diff
@@ -516,8 +529,7 @@ def _track(
             source_code: Source code.
             kind: Transform kind.
             entrypoint: Optional entrypoint name (e.g. function qualname) for the run.
-            initiated_by_run: Optional parent run (or its `uid`) that triggered this run.
-                If `None`, falls back to the `LAMIN_INITIATED_BY_RUN_UID` environment variable when set.
```
falexwolf, line 520: Please keep this.

falexwolf:
This looks nice, now! Can you also think about how to add a test for this?

falexwolf:
We'll also need the test to test the html export/report functionality.

falexwolf:
So, it's worthwhile to install marimo in one of the test groups (or make a new group, but probably not needed).

lamindb/core/_context.py
```diff
@@ -129,7 +130,22 @@ def get_uid_ext(version: str) -> str:
     return encodebytes(hashlib.md5(version.encode()).digest())[:4]  # noqa: S324
 
 
+def get_marimo_notebook_path() -> str | None:
+    if not _is_running_in_marimo():
+        return None
+    from marimo._runtime.context import safe_get_context
```
falexwolf: This export can break because `_runtime` is a private module in marimo. Can we not get a public API for this?

lamindb/core/_context.py
```diff
@@ -129,7 +130,22 @@ def get_uid_ext(version: str) -> str:
     return encodebytes(hashlib.md5(version.encode()).digest())[:4]  # noqa: S324
 
 
+def get_marimo_notebook_path() -> str | None:
+    if not _is_running_in_marimo():
```
falexwolf, line 134: Here we're calling this function the second time, so Sergei's comment about caching it in lamindb-setup like we do for IS_RUN_FROM_IPYTHON is valid.

falexwolf: But maybe not a priority right now. Can also be in a later refactor. I assume this function has no runtime.

ishitajain9717: Comiting the edits today, with also marimo showing the html report✅

lamindb/core/_context.py
ishitajain9717: There is no public api for  `_runtime`. But I can maybe have a couple of testcases where it might break. What do you think.

falexwolf: You could check `hasattr(marimo, "_runtime")` and if false throw an error: "marimo version is incompatible with lamindb, downgrade to <whatever-current-version-is>"

lamindb/core/_context.py
falexwolf: Great!

ishitajain9717:
Added the test to check if marimo gets the correct file name and also added a try and except block to produce an error message if there is an issue with _runtime import.

pyproject.toml
```diff
@@ -35,6 +35,7 @@ full = [
     "jupytext",
     "nbconvert>=7.2.1",  # bound to avoid lxml[html_clean] dependency
     "nbproject==0.11.1",  # adds orjson
+    "marimo",
```
falexwolf: Please remove. marimo is a very heavy dependency.

falexwolf:
@ishitajain9717, can you please do what I suggested rather then the complicated changes you made in your last commit?

https://github.com/laminlabs/lamindb/pull/3680#discussion_r3246337940

---

Also, please do not add marimo to the dependencies. It's much too heavy for a default dependency.

falexwolf:
And please remove the uv.lock file that you committed here.

ishitajain9717:
Will redo as suggested. Added marimo because the test errored with "no module named marimo" on CI, didn't realize how heavy it is.

falexwolf:
> Added marimo because the test errored with "no module named marimo" on CI

You can add it inside `noxfile.py` to run tests. It should not be a default dep, and it also doesn't merit an extra because it's just a single dep.

lamindb/core/_context.py
```diff
@@ -129,7 +130,32 @@ def get_uid_ext(version: str) -> str:
...
+        return None
+    import marimo
+
+    if not hasattr(marimo, "_runtime"):
+        raise RuntimeError(
+            "marimo version is incompatible with lamindb, downgrade to 0.23.6"
+        )
+
+    try:
+        from marimo._runtime.context import safe_get_context
```
falexwolf: @ishitajain9717, can you please deal with the `safe_get_context` attribute like you dealt with `_runtime`? I also think the `_runtime` guard is no longer needed.

A silent try/except is the wrong solution here.

lamindb/core/_context.py
```diff
@@ -849,9 +876,39 @@ def _track_notebook(
...
         else:
             path = Path(path_str)
+
         if pypackages is None:
             pypackages = True
         description = None
+        if self._notebook_runner == "marimo":
+            source = path.read_text(encoding="utf-8")
+            if (
+                'auto_download=["html"]' not in source
```
falexwolf: This should rather be a regex because also these two cases aren't safe. Spacing could be off, there could be multiple elements in the list, etc.

lamindb/core/_context.py
```diff
@@ -849,9 +876,39 @@ def _track_notebook(
...
+                and "auto_download=['html']" not in source
+            ):
+                if 'app = marimo.App(width="medium")' in source:
```
falexwolf: These cases aren't comprehensive.

lamindb/_finish.py
```diff
@@ -330,6 +330,16 @@ def save_context_core(
             logger.warning(
                 f"no html report found; to attach one, create an .html export for your {filepath.suffix} file and then run: lamin save {filepath}"
             )
+    elif notebook_runner == "marimo":
+        report_path = (
+            filepath.parent / "__marimo__" / filepath.with_suffix(".html").name
+        )
+        if not report_path.exists():
+            logger.warning(
```
falexwolf, line 370: Pretty!

lamindb/_finish.py
```diff
@@ -330,6 +330,16 @@ def save_context_core(
             logger.warning(
                 f"no html report found; to attach one, create an .html export for your {filepath.suffix} file and then run: lamin save {filepath}"
             )
+    elif notebook_runner == "marimo":
+        report_path = (
+            filepath.parent / "__marimo__" / filepath.with_suffix(".html").name
+        )
+        if not report_path.exists():
```
falexwolf: if the report_path exists you could run some post-processing that makes it a truly static html as discussed on Slack.

ishitajain9717:
Here in the first screenshot we see, when we use the plain format from marimo the output tables are snapshot of live session, and cannot be scrolled over. 
Screenshot 2026-05-21 at 11 25 11 AM

And thats why we used rich format along with css styling to allowing seeing full content of cells when hovered over them, also allowing scrolling bars. As you can see in the video. 

https://github.com/user-attachments/assets/036729ec-c954-4218-b379-e4390984539c

PR #3685, opened 2026-05-18
:art: Add safeguards against non-latest `revises` argument on saving a versioned sqlrecord

Koncopd:
Improve versioning robustness and errors for `revises` handling for versioned sqlrecords, especially artifacts.

Replaces brittle latest-version assertions with IntegrityError, refreshes/rechecks stale inferred `revises`, and tightens artifact key-based previous-version lookup behavior with targeted tests.

### DIFF
--- a/lamindb/models/artifact.py
+++ b/lamindb/models/artifact.py
@@ -380,12 +380,17 @@ def get_stat_or_artifact(
         skip_hash_lookup = True
     previous_artifact_version = None
     artifacts_qs = Artifact.objects.using(instance)
-    query_key_revises = (
+    # Only run key-based lineage inference when creating a new artifact version from key.
+    # If explicit `revises` was passed in the constructor, this flag is False so that we
+    # do not override/validate lineage from key lookup during init.
+    # Also skip on replace(): replacement updates content in-place and must not create
+    # a new version lineage from key.
+    should_lookup_key_lineage = (
         key is not None and not is_replace and not skip_key_revises_lookup
     )
     if skip_hash_lookup:
         artifact_with_same_hash_exists = False
-        if query_key_revises:
+        if should_lookup_key_lineage:
             # only search for a previous version of the artifact
             # ignoring hash
             queryset_same_hash_or_same_key = artifacts_qs.filter(
@@ -413,7 +418,9 @@ def get_stat_or_artifact(
             queryset_same_hash_or_same_key = artifacts_qs.filter(
                 ~Q(branch_id=-1),
                 (Q(hash=hash) | Q(key=key, storage=storage))
-                if query_key_revises
+                # Key lookup is conditionally included only for inferred lineage. For
+                # explicit `revises`, we only need hash dedup and should skip key lineage.
+                if should_lookup_key_lineage
                 else Q(hash=hash),
             ).order_by("-created_at", "-id")
             queryset_same_hash = queryset_same_hash_or_same_key.filter(hash=hash)
@@ -421,7 +428,7 @@ def get_stat_or_artifact(
     if key is not None and not is_replace:
         if (
             not artifact_with_same_hash_exists
-            and query_key_revises
+            and should_lookup_key_lineage
             and queryset_same_hash_or_same_key.exists()
         ):
             queryset_same_key = queryset_same_hash_or_same_key.filter(is_latest=True)
--- a/lamindb/models/artifact.py
+++ b/lamindb/models/artifact.py
@@ -443,10 +443,8 @@ def get_stat_or_artifact(
                 f"creating new artifact version for key '{key}' in storage '{storage.root}'"
             )
             previous_artifact_version = queryset_same_key.first()
-            assert previous_artifact_version is not None  # noqa: S101
     if artifact_with_same_hash_exists:
         artifact_with_same_hash = queryset_same_hash.first()
-        assert artifact_with_same_hash is not None  # noqa: S101
         logger.important(
             f"returning artifact with same hash: {artifact_with_same_hash}; to track this artifact as an input, use: ln.Artifact.get()"
         )
--- a/lamindb/models/artifact.py
+++ b/lamindb/models/artifact.py
@@ -1644,6 +1644,7 @@ class Artifact(SQLRecord, IsVersioned, TracksRun, TracksUpdates):
         using_key = kwargs.pop("using_key", None)
         description: str | None = kwargs.pop("description", None)
         revises: Artifact | None = kwargs.pop("revises", None)
+        refresh_revises_if_stale = revises is None
         if revises is not None:
             if not isinstance(revises, Artifact):
                 raise TypeError("`revises` has to be of type `Artifact`")
@@ -1882,6 +1883,7 @@ class Artifact(SQLRecord, IsVersioned, TracksRun, TracksUpdates):
         kwargs["space"] = space
         kwargs["otype"] = otype
         kwargs["revises"] = revises
+        kwargs["_refresh_revises_if_stale"] = refresh_revises_if_stale
         # this check needs to come down here because key might be populated from an
         # existing file path during get_artifact_kwargs_from_data()
         if (
--- a/lamindb/models/artifact.py
+++ b/lamindb/models/artifact.py
@@ -348,6 +348,7 @@ def get_stat_or_artifact(
     is_replace: bool = False,
     instance: str | None = None,
     skip_hash_lookup: bool = False,
+    skip_key_revises_lookup: bool = False,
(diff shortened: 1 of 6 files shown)


lamindb/models/artifact.py
```diff
@@ -408,21 +412,33 @@ def get_stat_or_artifact(
             # see the `previous_artifact_version` variable below
             queryset_same_hash_or_same_key = artifacts_qs.filter(
                 ~Q(branch_id=-1),
-                Q(hash=hash) | Q(key=key, storage=storage),
-            ).order_by("-created_at")
+                (Q(hash=hash) | Q(key=key, storage=storage))
+                if query_key_revises
```
falexwolf: Can you please add a comment here why that check depends on `query_key_revises`? It would maybe also help if `query_key_revises` had an easier-to-understand name. Like this it's quite hard to read and understand. I think I know why that's correct, but the comment would help a lot.

lamindb/models/artifact.py
```diff
@@ -408,21 +412,33 @@ def get_stat_or_artifact(
...
     if key is not None and not is_replace:
         if (
             not artifact_with_same_hash_exists
-            and queryset_same_hash_or_same_key.count() > 0
+            and query_key_revises
+            and queryset_same_hash_or_same_key.exists()
         ):
+            queryset_same_key = queryset_same_hash_or_same_key.filter(is_latest=True)
+            if not queryset_same_key.exists():
+                raise ValidationError(
```
falexwolf: I would make this a ValueError or anything else, but not a ValidationError: this is an error that represents an integrity state problem in the system (we should always have one object with `is_latest=True`), not invalid data from the user.

lamindb/models/artifact.py
```diff
@@ -408,21 +412,33 @@ def get_stat_or_artifact(
...
             logger.important(
                 f"creating new artifact version for key '{key}' in storage '{storage.root}'"
             )
-            previous_artifact_version = queryset_same_hash_or_same_key[0]
+            previous_artifact_version = queryset_same_key.first()
+            assert previous_artifact_version is not None  # noqa: S101
     if artifact_with_same_hash_exists:
-        artifact_with_same_hash = queryset_same_hash[0]
+        artifact_with_same_hash = queryset_same_hash.first()
+        assert artifact_with_same_hash is not None  # noqa: S101
```
falexwolf: You don't need to add `# noqa: S101` anymore. Please add an error message after the tuple (I know we haven't stringently done this, but I learned from Fred over the years that this is super helpful and recently only experienced that again)

lamindb/models/sqlrecord.py
```diff
@@ -1156,11 +1168,39 @@ def save(self: T, *args, **kwargs) -> T:
...
+                        has_branch_id = hasattr(revises, "branch_id") and hasattr(
+                            self, "branch_id"
+                        )
                         should_demote = True
-                        if hasattr(revises, "branch_id") and hasattr(self, "branch_id"):
+                        if has_branch_id:
                             should_demote = revises.branch_id == self.branch_id
                         if should_demote:
-                            assert revises.is_latest  # noqa: S101
+                            revises.refresh_from_db(fields=["is_latest"])
```
falexwolf, line 1197: The very high level of indendation make this very hard to read or review on my screen. Not a must-change, but we don't typically indent so far.

lamindb/models/sqlrecord.py
```diff
@@ -1156,11 +1168,39 @@ def save(self: T, *args, **kwargs) -> T:
...
+                                        branch_id=self.branch_id
+                                    )
+                                latest_revises = latest_revises_qs.order_by(
+                                    "-created_at", "-id"
+                                ).first()
+                                if latest_revises is not None:
+                                    revises = latest_revises
+                                    self._revises = latest_revises
+                            if not revises.is_latest:
+                                raise ValidationError(
```
falexwolf: I don't think this should be a ValidationError for the same reason as above.

tests/core/test_artifact_basics.py
```diff
@@ -905,6 +906,42 @@ def test_revise_recreate_artifact(example_dataframe: pd.DataFrame, ccaplog):
...
+
+        # Key-based revises inference should now fail because no matching latest exists.
+        df.iloc[0, 0] = 202
+        with pytest.raises(
+            ValidationError,
+            match="matching non-trashed artifacts exist",
+        ):
+            ln.Artifact.from_dataframe(df, key=key, description="v3-inferred")
+
+        # Explicit revises should bypass key-lineage inference during init.
```
falexwolf: What is "key-lineage"? Maybe try finding words that are a bit easier to understand in this comment?

tests/core/test_artifact_basics.py
```diff
@@ -905,6 +906,42 @@ def test_revise_recreate_artifact(example_dataframe: pd.DataFrame, ccaplog):
...
+        artifact_v1.refresh_from_db()
+        artifact_v2.refresh_from_db()
+        assert not artifact_v1.is_latest
+        assert not artifact_v2.is_latest
+
+        # Key-based revises inference should now fail because no matching latest exists.
+        df.iloc[0, 0] = 202
+        with pytest.raises(
+            ValidationError,
+            match="matching non-trashed artifacts exist",
```
falexwolf: This `match` argument doesn't work the way you think this works. We've all been fooled by it. Either you use a proper regex or you check outside `raises()`. This here is likely always true independent of what is matched. I can dig out the Slack thread from years ago. ☺️

tests/core/test_is_versioned.py
```diff
@@ -225,6 +226,92 @@ def test_transform_versioning_across_branches_preserves_main_latest():
...
+            key="stale-inferred-revises-requery",
+            source_code="v3",
+            kind="pipeline",
+        )
+        assert transform_pending._revises is not None
+        assert transform_pending._revises.uid == transform_v2.uid
+        assert transform_pending._refresh_revises_if_stale
+
+        # Simulate stale latest flags without creating a new version.
+        ln.Transform.objects.filter(id=transform_v2.id).update(is_latest=False)
```
falexwolf: Do you want to purposefully bypass lamindb because you invoke `.objects` on the Django-level? If not, please don't use the Django-level as it confuses agents (they all end up writing Django instead of lamindb with all its downsides).

falexwolf (changes requested):
This is great! But I think I found small issues. ☺️

lamindb/models/artifact.py
Koncopd: renamed to `should_lookup_key_lineage` and added a comment

lamindb/models/artifact.py
Koncopd: Using `IntegrityError` now.

lamindb/models/artifact.py
Koncopd: I just removed these asserts, they are redundant because we check explicitly the existence before.

lamindb/models/sqlrecord.py
Koncopd: also using `IntegrityError` here.

tests/core/test_artifact_basics.py
Koncopd: done

tests/core/test_artifact_basics.py
Koncopd: replaced with explicit check

tests/core/test_is_versioned.py
Koncopd: fixed

lamindb/models/artifact.py
falexwolf: Great! What do you mean with "lineage"? The term "lineage" essentially always means "data lineage" for us, but you have something else in mind, right? That's why I still think this is confusing.

lamindb/models/sqlrecord.py
Koncopd: refactored

Koncopd:
@falexwolf i believe i have addressed all of them.

falexwolf:
Great!

falexwolf:
This one? https://github.com/laminlabs/lamindb/pull/3685#discussion_r3265991815

lamindb/models/artifact.py
Koncopd: version family or version line ( we don't have trees). So i should not use the term?

Koncopd:
answered there

lamindb/models/artifact.py
falexwolf: I think it's highly confusing in this context, yes, and I hope you find another word. I think "version family" is fine. I don't understand what "version line" is.

Koncopd: ok renamed

PR #3688, opened 2026-05-20
📝 More visibility for branching in the docs

falexwolf:
Previously `lamindb.branch` was the only page where one could see several examples. Now the information is also visible from: https://docs.lamin.ai/manage-changes

Before | After
--- | ---
image | image

This FAQ doc disappeared and is now integrated into this comprehensive doc: https://docs.lamin.ai/faq/trash-archive


sheetalgiri:
I think we should be instead explicit what these primary objects are artifacts,transforms and so on. It might be too technical for some to go into sql records documentation and try to figure this out

> All primary objects – any that inherit from [SQLRecord](https://38b2287c.lamindb.pages.dev/lamindb.models#lamindb.models.SQLRecord) – 

I think this information would make more sense inside the Contribution branches section or differently organized 

> All primary objects – any that inherit from [SQLRecord](https://38b2287c.lamindb.pages.dev/lamindb.models#lamindb.models.SQLRecord) – have a [branch](https://38b2287c.lamindb.pages.dev/lamindb.models#lamindb.models.SQLRecord.branch) field that determines their life cycle.
> There are three built-in branches: main, trash, and archive. By default, objects are created on the main branch and visible in queries and searches.

I asked Gemini about improving the organization, and it suggested having an object life cycle section as the second section, and I'm not saying we should exactly follow it but it does make the information clearer

image

falexwolf:
Thanks! I realize I had not yet intended to merge this. Will make another PR to polish.

falexwolf:
How about this?

- https://github.com/laminlabs/lamindb/pull/3689

PR #3690, opened 2026-05-20
✨ Allow users to delete non-empty storage locations

Koncopd:


falexwolf:
Can you add more context? Reading the PR title and skimming the changes this looks like we might introduce danger for accidental data loss?

Koncopd:
@falexwolf the context is here https://laminlabs.slack.com/archives/C062QDNMN6M/p1778148568628459

falexwolf:
I see, makes sense. Let me check the implementation.

PR #3691, opened 2026-05-20
✨ Pull existing artifacts for `s3/gs` paths in `AUTO_KEY_PREFIX` when creating an artifact

Koncopd:
`Artifact` can now be initialized directly from managed cloud paths inside `.lamindb/` on S3/GCS, instead of those paths being rejected.

This makes it easier to reference artifacts that already exist in managed storage, including artifacts managed by another instance.

## Example

```python
import lamindb as ln

path = "s3://my-bucket/.lamindb/ABCDEF1234567890.parquet"

# New: this resolves the existing artifact metadata
artifact = ln.Artifact(path) # assume exists in the current or another instance

print(artifact.uid)
artifact.save() # transfer if on another instance
```

## When it still errors

Clear `ValueError`s are raised when:

- the storage location for the path is not registered,
- no managing instance can be resolved for that storage,
- or no artifact exists at the specified managed path.

For non-cloud paths containing `.lamindb/`, validation still rejects them.

### DIFF
--- a/lamindb/models/artifact.py
+++ b/lamindb/models/artifact.py
@@ -15,7 +15,10 @@ from django.db.models import CASCADE, PROTECT, Q
 from django.db.models.functions import Length
 from lamin_utils import colors, logger
 from lamindb_setup import settings as setup_settings
-from lamindb_setup.core._hub_core import select_storage_or_parent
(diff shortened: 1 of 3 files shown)


falexwolf (approved):
Perfect! 😄

PR #3700, opened 2026-05-25
:children_crossing: No longer attempt deduplicating existing artifacts in storage 

Koncopd:
Given a file or folder is already ingested into a storage location, `Artifact(f"s3://{storage_loc}/...")` will no longer attempt deduplicating (`skip_hash_lookup = True`).

Newly ingested artifacts will still be deduplicated based on hash equivalence (`skip_hash_lookup = False`).

The default value for the `skip_hash_lookup` in `Artifact()` changes from `False` to `None`.

### DIFF
--- a/docs/faq/idempotency.md
+++ b/docs/faq/idempotency.md
@@ -115,6 +115,10 @@ artifact2 = ln.Artifact(filepath, key="my_fcs_file.fcs")
 
 It gives us the existing object:
 
+This behavior applies when data is written into managed storage; for paths that
+already live in a registered storage location, hash lookup is skipped by default
+unless you pass `skip_hash_lookup=False`.
+
 ```python
 assert artifact.id == artifact2.id
 assert artifact.run == artifact2.run
--- a/docs/faq/idempotency.md
+++ b/docs/faq/idempotency.md
@@ -19,7 +19,7 @@ If you set {attr}`~lamindb.core.subsettings.CreationSettings.search_names` to `F
 If you instantiate {class}`~lamindb.Artifact` from data that is written into a storage location,
 the `Artifact()` constructor returns the existing artifact based on a hash lookup.
 For paths that already live in a registered storage location, hash lookup is skipped by default.
-You can override this via `hash_lookup="check"` (or force skipping via `hash_lookup="skip"`).
+You can override this via `skip_hash_lookup=False` (or force skipping via `skip_hash_lookup=True`).
 
 :::
 
@@ -129,7 +129,7 @@ artifact2.save()
 
 For paths that already live in a registered storage location, `Artifact()`
 skips hash lookup by default during initialization. You can override this with
-`hash_lookup="check"` to force hash lookup or `hash_lookup="skip"` to always skip it.
+`skip_hash_lookup=False` to force hash lookup or `skip_hash_lookup=True` to always skip it.
 
 In the hidden cell below, you'll see how this interplays with data lineage.
 
--- a/docs/faq/idempotency.md
+++ b/docs/faq/idempotency.md
@@ -16,7 +16,10 @@ If you set {attr}`~lamindb.core.subsettings.CreationSettings.search_names` to `F
 
 :::{admonition} Artifacts & collections
 
-If you instantiate {class}`~lamindb.Artifact` from data that already exists as an artifact, the `Artifact()` constructor returns the existing artifact based on a hash lookup.
+If you instantiate {class}`~lamindb.Artifact` from data that is written into a storage location,
+the `Artifact()` constructor returns the existing artifact based on a hash lookup.
+For paths that already live in a registered storage location, hash lookup is skipped by default.
+You can override this via `hash_lookup="check"` (or force skipping via `hash_lookup="skip"`).
 
 :::
 
@@ -122,6 +125,14 @@ If you save it again, nothing will happen (the operation is idempotent):
 
 ```python
 artifact2.save()
+```
+
+For a path that already exists in registered storage, a new artifact is created by default:
+
+```python
+registered_path = artifact.path
+artifact4 = ln.Artifact(registered_path).save()
+assert artifact4.id != artifact.id
 ```
 
 In the hidden cell below, you'll see how this interplays with data lineage.
--- a/docs/faq/idempotency.md
+++ b/docs/faq/idempotency.md
@@ -115,9 +115,10 @@ artifact2 = ln.Artifact(filepath, key="my_fcs_file.fcs")
 
 It gives us the existing object:
 
-This behavior applies when data is written into managed storage; for paths that
-already live in a registered storage location, hash lookup is skipped by default
-unless you pass `skip_hash_lookup=False`.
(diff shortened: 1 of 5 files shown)


docs/faq/idempotency.md
```diff
@@ -16,7 +16,10 @@ If you set {attr}`~lamindb.core.subsettings.CreationSettings.search_names` to `F
 
 :::{admonition} Artifacts & collections
 
-If you instantiate {class}`~lamindb.Artifact` from data that already exists as an artifact, the `Artifact()` constructor returns the existing artifact based on a hash lookup.
+If you instantiate {class}`~lamindb.Artifact` from data that is written into a storage location,
```
falexwolf: Wording here is confusing because it reads like it means the opposite of what it's supposed to mean. 😆

docs/faq/idempotency.md
```diff
@@ -16,7 +16,10 @@ If you set {attr}`~lamindb.core.subsettings.CreationSettings.search_names` to `F
...
+the `Artifact()` constructor returns the existing artifact based on a hash lookup.
+For paths that already live in a registered storage location, hash lookup is skipped by default.
+You can override this via `hash_lookup="check"` (or force skipping via `hash_lookup="skip"`).
```
falexwolf: Why do you rename the arg here? Isn't it called `skip_hash_lookup`?

Koncopd: With the new default behavior `skip_hash_lookup` being bool becomes confusing on `False`.

lamindb/models/artifact.py
```diff
@@ -506,7 +506,7 @@ def get_artifact_kwargs_from_data(
     is_replace: bool = False,
     skip_check_exists: bool = False,
     overwrite_versions: bool | None = None,
-    skip_hash_lookup: bool = False,
+    hash_lookup: Literal["auto", "skip", "check"] = "auto",
```
falexwolf: The `skip_hash_lookup` variable is widely used and changing its name now outside of a major version bump introduces some friction (some devs will get many, many warnings because they set the variable everywhere in their logic).

Also, dealing with it via `Literal` breaks style with most other parts of the code. We essentially exclusively use `None` to indicate a parameter that's intelligently chosen (e.g. `overwrite_versions`).

I'd keep the name `skip_hash_lookup` and default it to `None` from now on, typed like `overwrite_versions`. I personally think that's easier than learning a custom convention for a `Literal`. For instance, why is it called "check" and not something else.

falexwolf: I think a more intuitive parameter name could be `deduplicate`, but that'd only be possible in a major version upgrade.

lamindb/models/artifact.py
```diff
@@ -1174,8 +1186,13 @@ class Artifact(SQLRecord, IsVersioned, TracksRun, TracksUpdates):
         branch: `Branch | None = None` The branch of the artifact. If `None`, uses the current branch.
         space: `Space | None = None` The space of the artifact. If `None`, uses the current space.
         storage: `Storage | None = None` The storage location for the artifact. If `None`, uses the default (:attr:`~lamindb.core.Settings.storage`).
-        skip_hash_lookup: `bool = False` Skip the hash lookup so that a new artifact is created even if an artifact with the same hash already exists.
-            Empty files are always treated as if this were `True` because empty content hashes are not used for deduplication.
+        hash_lookup: `Literal["auto", "skip", "check"] = "auto"` Controls hash-based deduplication.
+            `"auto"` checks hashes for upload flows and skips hash lookup for paths already in registered storage.
```
falexwolf: This way of documenting things uses a lot of vertical space if it's properly done, and gives a complicated appearance. I think our CSS isn't able to deal with it, so in the rare places we have complicated optionality we're actually writing it in a paragraph.

Either way, I think it'd be much easier to document if this was just `skip_hash_lookup: bool | None` like all the other ones.

lamindb/models/artifact.py
```diff
@@ -1294,7 +1311,7 @@ class Artifact(SQLRecord, IsVersioned, TracksRun, TracksUpdates):
     .. dropdown:: Will artifacts get duplicated?
 
         If an artifact with the exact same hash already exists, `Artifact()` returns the existing artifact.
-        Exception: empty files are not deduplicated by hash and create a new artifact.
+        Exception: paths that already live in a registered storage location and empty files skip hash deduplication by default.
```
falexwolf, line 1308: How about "managed storage location"?

lamindb/models/artifact.py
```diff
@@ -1668,7 +1686,25 @@ def __init__(
...
         version_tag: str | None = kwargs.pop("version_tag", kwargs.pop("version", None))
         features: dict[str, Any] | None = kwargs.pop("features", None)
-        skip_hash_lookup: bool = kwargs.pop("skip_hash_lookup", False)
+        hash_lookup: Literal["auto", "skip", "check"] = kwargs.pop(
+            "hash_lookup", "auto"
+        )
+        if "skip_hash_lookup" in kwargs:
+            deprecated_skip_hash_lookup = kwargs.pop("skip_hash_lookup")
+            logger.warning(
+                "`skip_hash_lookup` is deprecated and will be removed in a future release; "
```
falexwolf: If we really want to go through with it, it'd have to be a `DeprecationWarning` and not a logger message. But I think I'm pretty strongly against causing this disruption in favor of a name that's not clearer than the current name.

lamindb/models/artifact.py
Koncopd: ok, i will revert

docs/faq/idempotency.md
Koncopd: restored

lamindb/models/artifact.py
Koncopd: restored

lamindb/models/artifact.py
Koncopd: It is not only for managed, for any registered storage actually.

lamindb/models/artifact.py
Koncopd: removed

docs/faq/idempotency.md
Koncopd: is it better now?

lamindb/models/artifact.py
Koncopd: removed that

lamindb/models/artifact.py
falexwolf: Right!

falexwolf:
Let's think through idempotency.

There is one case that we need to pay attention to: repeatedly registering the same path in a managed storage location. This should be covered due to key-lookup, but I think we should also document and double check the test for it.

We have both hash- and key-lookup, both of which helps making the LaminDB API idempotent.

Koncopd:
@falexwolf i think this is covered by this block https://github.com/laminlabs/lamindb/blob/ef474e794ac4ff4b4787c6f41f9a22e87c6dd38c/lamindb/models/sqlrecord.py#L1224

Also see my `test_skip_hash_lookup_true_on_upload_same_key_resolves_collision_on_save`.

PR #3708, opened 2026-06-02
:children_crossing: Improve typing of `Feature()` constructor

ishitajain9717:
Added an overload for the feature
 https://github.com/pfizer-collab/laminlabs/issues/960

### DIFF
--- a/lamindb/models/feature.py
+++ b/lamindb/models/feature.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import importlib
 import warnings
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, Any, cast, get_args, overload
+from typing import TYPE_CHECKING, Any, Literal, cast, get_args, overload
 
 import numpy as np
 import pgtrigger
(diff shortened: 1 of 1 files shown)


falexwolf:
Thanks, @ishitajain!

Can you cross-link the issue? 

This way of fixing it isn't entirely correct I believe. The issue was that `is_type` didn't show in the overload, but if we look at it clearly does. So, something else is wrong in the original issue. I think we should clear this up rather than adding another overload. WDYT?

ishitajain9717:
Actually, I tested this locally with the original version, and the issue doesn’t seem to be with is_type in the overload. If in original version I add the `dtype=` then the IDE does not show an error.  The problem is actually that dtype is treated as required in the default signature. When I add a separate overload where is_type is Literal[True] and dtype is removed, the IDE error disappears. So it looks more like a  mismatch around dtype, not the presence of is_type, thats why I added an another overload.  How do you think about it?

falexwolf:
Great! That makes perfect sense, and you're right that this merits an overload because `dtype` is not required when creating a type.

For consistency and comprehensiveness, I think, this would ideally be defined on the `HasType` child class. Then it'll not only hold for feature types but also for all others. On the other I think that no other class has a situation where another field is required. That would be an argument to do what you do.

I still don't understand why `Literal[True]` is correct, however. Is this how one indicates a required `True` value? In that case also this makes sense.

Thanks for explaining!

ishitajain9717:
Without Literal[True], There is ambiguity because Python matches overloads in order  (I was also thinking when I discussed it with gemini that why do we need it to be True) and cannot distinguish the is_type=True constructor path. So it may fall back to the first compatible overload, which leads to incorrect type inference. I can see the hastype child class, and can update on that

falexwolf:
Got it! Makes sense! Let's keep it this way. It's good to merge from my end. ☺️

PR #3713, opened 2026-06-04
🚸 Inherit space from `storage.space` in Artifact constructor if `storage` is passed and `space` is not passed

falexwolf:
Addresses:

- https://github.com/pfizer-collab/laminlabs/issues/963

### DIFF
--- a/lamindb/models/artifact.py
+++ b/lamindb/models/artifact.py
@@ -1797,12 +1797,17 @@ class Artifact(SQLRecord, IsVersioned, TracksRun, TracksUpdates):
         else:
             storage = setup_settings.instance.storage.record
         if space is None:
-            from lamindb import context as run_context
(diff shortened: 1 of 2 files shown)


lamindb/models/artifact.py
```diff
@@ -1797,12 +1797,17 @@ def __init__(
         else:
             storage = setup_settings.instance.storage.record
         if space is None:
-            from lamindb import context as run_context
+            if storage_was_passed and storage.space_id is not None:
```
Koncopd: why would storage.space_id be None? Is it even possible?

Koncopd (approved):

lamindb/models/artifact.py
falexwolf: You're right. No, it's not possible.

PR #3714, opened 2026-06-05
✨ Constrain the content of dynamic registries to be in the same space as the registry

falexwolf:
Given a `SQLRecord[HasType]` with `is_type=True` should be thought of as a dynamic registry, the mental model of most users is that its content should have the same access permissions as the registry itself.

You can now enforce this by changing the `single_space` setting via:

```python
import lamindb as ln

experiments_registry = ln.Record.get(name="Experiments", is_type=True)
experiments_registry.settings.single_space = True
experiments_registry.save()
```

## Record docs

image

## Settings docs

image


sunnyosun (commented):
What I didn't get from reading the code is that this doesn't seem recursive, right? Meaning if I have nested types under the single space type, do all records under the nested types also require single space?

falexwolf:
Yeah, it is recursive for free. Checking the immediate parent type is indeed enough.

PR #3715, opened 2026-06-05
♻️ Homogenize branch and space treatment in constructors

falexwolf:
There were some inconsistencies before; which isn't astonishing given it wasn't centralized and comprehensively tested.


falexwolf:
Please take a look, @Koncopd!

falexwolf:
There were inconsistencies before; which isn't astonishing given it wasn't centralized and comprehensively tested.

PR #3716, opened 2026-06-05
:children_crossing: Store `index`-like features in the `name` field of `Record`

sunnyosun:
## Summary

When a record sheet schema defines `Schema.index`, the index feature is stored on `Record.name` instead of in feature link tables (`RecordJson`, etc.). This mirrors `Schema.index` for `DataFrame` / `AnnData` curation: the index is the row key, not a regular feature column.

On record sheets, the index feature must have `dtype=str` because `Record.name` is a string field. Non-string index dtypes remain valid for artifact curation schemas (`DataFrame`, `AnnData`, etc.).

## Write path

- `features.add_values()`, `set_values()`, `Record(features=...)`, and `Record.from_dataframe()` route the index feature to `record.name`
- Link-table writes for the index feature are skipped
- Saving a record type with a non-`str` index feature raises `ValueError`

## Read path

- `features.get_values()` and `features["<index>"]` inject the index from `record.name`

## Export

`Record.to_dataframe()` puts the index on `df.index` (named after the index feature) instead of as a column.

| Schema has `index` | Export shape |
| --- | --- |
| **Yes** | Clean export: index on `df.index`; encoded metadata columns omitted (`__lamindb_record_id__`, `__lamindb_record_uid__`, `__lamindb_record_name__`, etc.) |
| **No** | Previous behavior unchanged (metadata columns included) |

This removes the duplicate `name` + `__lamindb_record_name__` columns when the index feature is called `name`.

## Import

- `Record.from_dataframe()` accepts a dataframe whose index matches the schema index feature (or the index feature as a column)

## CSV artifacts

- `to_artifact()` writes with `index=True` when the sheet schema has an index

## Documentation

- Record sheet index behavior documented on `Record` (class Notes, `to_dataframe`, `from_dataframe`, `to_artifact`)
- `Schema` / `Schema.index` cross-reference record sheets briefly; operational details live on `Record`

## Test plan

- [x] `test_record_schema_index_stored_on_name` — write/read/export/import round-trip, clean export, CSV artifact, non-`str` index rejected
- [x] `test_record_sheet_examples` — compound-treatment fixture with indexed export
- [x] `test_record_docstring_examples` — indexed sheet example

## Out of scope

- No migration or cleanup of legacy index values already stored in link tables


falexwolf:
This is awesome! 😄 

> Out of scope: no migration or cleanup of legacy index values already stored in link tables.

That shouldn't be a big deal, because to our knowledge `index=True` wasn't used in the context of defining sheet schemas, right?

lamindb/models/record.py
```diff
@@ -60,6 +61,209 @@
...
+
+
+def apply_index_feature_to_record(
+    record: Record,
+    feature: Feature,
+    value: Any,
+    *,
+    persist: bool = True,
+) -> None:
+    """Set ``record.name`` from an index feature value."""
```
falexwolf: Please no double backticks anywhere in docstrings.

lamindb/models/record.py
```diff
@@ -60,6 +61,209 @@
...
+        feature for feature in feature_objects if feature.uid != index_feature.uid
+    ]
+    return dictionary, feature_objects
+
+
+def save_record_json_values(feature_json_values: list) -> None:
+    """Persist ``RecordJson`` rows, upserting on ``(record, feature)``."""
+    from .record import RecordJson
+
+    for record_json in feature_json_values:
```
falexwolf: These seems prohibitively inefficient compared to the bulk operation we had before. An LLM artifact?

tests/core/test_record_basics.py
```diff
@@ -143,6 +143,42 @@ def test_record_from_dataframe_bulk_save_paths():
...
+        features=[score],
+        index=sample_id,
+        name="index-on-name-schema",
+    ).save()
+    sheet = ln.Record(name="index-sheet", is_type=True, schema=schema).save()
+
+    record = ln.Record(type=sheet, features={"sample_id": "S-001", "score": 1.5}).save()
+    assert record.name == "S-001"
+    assert ln.models.RecordJson.filter(record=record, feature=sample_id).count() == 0
+    assert record.features.get_values() == {"sample_id": "S-001", "score": 1.5}
```
falexwolf, line 200: Great!

tests/core/test_record_basics.py
```diff
@@ -143,6 +143,42 @@ def test_record_from_dataframe_bulk_save_paths():
...
+
+    df = sheet.to_dataframe()
+    assert df.index.name == "sample_id"
+    assert "sample_id" not in df.columns
+    assert "__lamindb_record_name__" not in df.columns
+    assert df.index.tolist() == ["S-001"]
```
falexwolf, line 208: This sequence of asserts is also great!

tests/core/test_record_basics.py
```diff
@@ -143,6 +143,42 @@ def test_record_from_dataframe_bulk_save_paths():
...
+
+    batch = ln.Record.from_dataframe(
+        pd.DataFrame({"score": [2.5]}, index=pd.Index(["S-002"], name="sample_id")),
```
falexwolf, line 219: I wasn't aware that `DataFrame` accepts a `name` argument that would be scoped to its `index` arg. Interesting.

Are we using that?

tests/core/test_record_basics.py
```diff
@@ -143,6 +143,42 @@ def test_record_from_dataframe_bulk_save_paths():
...
+        type=sheet,
+    )
+    batch.save()
+    record2 = ln.Record.get(name="S-002")
+    assert record2.features.get_values()["score"] == 2.5
```
falexwolf: You could use `record2.features["score"]` for more concise code.

falexwolf:
Great! 

The one thing I wonder is whether we now still have a case where we test that `__lamindb_record_name__` still works in case `record.type` schema doesn't have an index? Did that survive the refactor?

---

I also wonder how this can be extended to map other features on record fields, e.g., if I have a feature `created_by`, I'd want to map this on `record.created_by` rather than a link table. Similar for `uid`, `description`, etc.

If we want to be consistent with that we need to indicate that a feature should be used as a "special column" when stored as a record.

And then we run into the question whether this should be stored on the schema-level or on the feature-level with its pros and cons.

---

Should it be a property of a feature to act as an index/created_by/description/uid/etc. or should that be a property of using a feature in the context of a schema?

I'm almost leaning to the former even though both can be argued. It could also be defined on the feature level (e.g. useful if records aren't used with schemas) and then superseded by a schema-level specification.

tests/core/test_record_basics.py
sunnyosun: Not yet in lamindb I think, but I use that sometimes.

sunnyosun:
> This is awesome! 😄
> 
> > Out of scope: no migration or cleanup of legacy index values already stored in link tables.
> 
> That shouldn't be a big deal, because to our knowledge `index=True` wasn't used in the context of defining sheet schemas, right?

Right, exactly.

sunnyosun:
> The one thing I wonder is whether we now still have a case where we test that __lamindb_record_name__ still works in case record.type schema doesn't have an index? Did that survive the refactor?

Correct, this case still exists. I want to have another small PR after this that introduces the hide/show __lamindb__record__xx columns.

Update, we agreed on when a schema has index feature, we don't export __lamindb_record_xxx columns.

falexwolf:
Ok!

sunnyosun:
> Should it be a property of a feature to act as an index/created_by/description/uid/etc. or should that be a property of using a feature in the context of a schema?
> 
> I'm almost leaning to the former even though both can be argued. It could also be defined on the feature level (e.g. useful if records aren't used with schemas) and then superseded by a schema-level specification.

I think for `index` mapping to `Record.name` it should be at the schema level as is. It's more efficient to just look for type.schema instead of look through all features in the schema to figure out which one is the index. Also if it's defined at the feature level, user might make mistake to add two index features to the schema which needs additional logics to check.

> I also wonder how this can be extended to map other features on record fields, e.g., if I have a feature created_by, I'd want to map this on record.created_by rather than a link table. Similar for uid, description, etc.
> 
> If we want to be consistent with that we need to indicate that a feature should be used as a "special column" when stored as a record.
> 
> And then we run into the question whether this should be stored on the schema-level or on the feature-level with its pros and cons.

For `description`, `created_by` and other fields, probably setting them at the feature level makes more sense.

falexwolf:
I just took another look and see that you're not documenting anything!

PR #3731, opened 2026-06-11
fix(feature): avoid IndexError when parsing a dtype with a trailing dot

devteamaegis:
## What's broken

`parse_nested_brackets` crashes with `IndexError: string index out of range` on any dtype segment ending in a dot, e.g. `"bionty."`. It's reachable from the public `parse_dtype` via a malformed categorical dtype:

```python
parse_dtype("cat[bionty.]")
```
```
  File ".../lamindb/models/feature.py", line 300, in parse_nested_brackets
    if len(parts) == 2 and parts[1][0].isupper():
IndexError: string index out of range
```

## Why it happens

After `dtype_str.split(".")`, the code checks `len(parts) == 2` but then reads `parts[1][0]` without checking that `parts[1]` is non-empty — a trailing dot makes it `""`.

## Fix

Guard `parts[1] != ""` before indexing. The malformed input now falls through to the bare-registry branch, so `parse_dtype("cat[bionty.]")` raises a clear `ValidationError` ("invalid dtype") instead of an opaque `IndexError`.

## Test

`test_parse_nested_brackets_trailing_dot` checks `"bionty."` no longer crashes and `parse_dtype("cat[bionty.]")` raises `ValidationError`.

### DIFF
--- a/lamindb/models/feature.py
+++ b/lamindb/models/feature.py
@@ -297,7 +297,7 @@ def parse_nested_brackets(dtype_str: str) -> dict[str, Any]:
         # No brackets - handle simple cases like "A" or "A.field"
         if "." in dtype_str:
             parts = dtype_str.split(".")
-            if len(parts) == 2 and parts[1][0].isupper():
+            if len(parts) == 2 and parts[1] != "" and parts[1][0].isupper():
                 # bionty.CellType
                 return {"registry": dtype_str, "filter_str": "", "field": ""}
             elif len(parts) == 3:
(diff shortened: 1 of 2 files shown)


falexwolf:
Thank you very much for the contribution! Will merge a version of this.

Given you ran into this: Have you considered passing Python objects instead of strings?

falexwolf:
I think what we should probably do is reason through a few more cases in which strings that a user passes could raise opaque errors.

This fix is good but it also feels a bit ad hoc; also testing strategy.

Thank you anyway, we'll mull over this for a bit more.

falexwolf:
Pinged you via email! Can re-open if it's important.

PR #3733, opened 2026-06-11
🐛 Fix edge cases in version handling

Koncopd:
### Bugs & Issues

1. **Version chains couldn't grow past `…000Z`.** (Reported in https://github.com/pfizer-collab/laminlabs/issues/988) When building up an artifact/transform version family, once the version suffix crossed from `…000Z` to `…000a`, LaminDB kept selecting `…000Z` as the latest version, regenerating `…000a` over and over instead of advancing the chain. Root cause: the "latest version in family" lookup relied on the database's string ordering of `uid`, which matches base62 order only under binary collation (e.g. SQLite). Under locale-aware collations (e.g. Postgres `en_US.UTF-8`), uppercase `Z` sorts after lowercase `a`, breaking the chain at the `Z → a` boundary. This is the user-reported bug.

2. **Stale-revision recovery could drop data on concurrent writes.** When a new version was inferred from `key` and another version was created concurrently before save, the record's id wasn't recomputed against the new family state. The save then collided on the unique id and silently returned the concurrently-created record, discarding the user's content.

3. **Initial inferred `revises` can select a head from the wrong branch.** When creating a new version by `key` without explicitly passing `revises`, LaminDB inferred the previous version by querying `is_latest=True` ordered by most-recently-created - but without scoping to the branch the new record was being created on. Because `is_latest` is tracked per branch, a single version family can have several heads (one per branch). The lookup therefore returned the globally most-recently-created head, which could belong to a *different* branch than the target one. Since demotion is branch-aware (a head is only demoted when it's on the same branch as the new record), revising from a foreign-branch head left the target branch's own head undemoted — resulting in two `is_latest=True` records on the target branch.

4. **Deleting a version head didn't promote a successor per branch.** When soft- or permanently deleting the latest version of a family, LaminDB promoted a new `is_latest` by picking the globally most-recently-created remaining version, ignoring the branch. Because `is_latest` is tracked per branch (a family can have one head per branch), deleting a branch's head could "promote" a record that lives on another branch — which was already `is_latest=True` there, so a no-op — and leave the deleted record's own branch with **no** `is_latest=True` version (a headless branch), or flip the wrong branch's record to latest. On top of that, promoting the successor and trashing/deleting the old head were not done in a single transaction, so a failure in between could leave both the successor and the old head marked `is_latest=True` (or neither).

### Fixes

1. Latest-version selection is now computed in a collation-independent way, so chains grow correctly across all boundaries (`9 → A`, `Z → a`) on any database backend.

2. On inferred revisions, the new version is re-derived from the current family state only when the previously chosen revision has gone stale, so concurrent writes produce a proper next version instead of a collision/silent-return.

3. The inferred lookup now prefers the latest version on the branch the record will actually be created on, and only falls back to the most recent head across branches when the target branch has no version of the key yet.

4. Promotion on delete is now branch-aware: for each deleted head, the successor is the most-recently-created remaining version of the same family **on the same branch**, so heads on other branches are left intact and the deleted record's branch always gets a new head (when one exists). The successor promotion and the delete/trash now run inside a single transaction on both the single-record and queryset delete paths, so they either both apply or neither does — no transient or persisted double/zero-head states.

### DIFF
--- a/tests/core/test_is_versioned.py
+++ b/tests/core/test_is_versioned.py
@@ -522,6 +522,74 @@ def test_inferred_revises_prefers_target_branch_head():
         branch.delete(permanent=True)
 
 
+def test_inferred_revises_prefers_target_branch_head_artifact():
+    main_branch = ln.Branch.get(name="main")
+    ln.setup.switch(main_branch.name)
+    branch = ln.Branch(name="test_inferred_revises_branch_artifact").save()
+    key = "inferred-revises-branch-aware-artifact.parquet"
+    artifact_main_v1 = ln.Artifact.from_dataframe(
+        pd.DataFrame({"inferred_feat": [20, 21]}),
+        key=key,
+        description="main-v1",
+    ).save()
+    # sanity check: `switch` must actually steer the creation branch, otherwise the
+    # rest of the test (which relies on heads living on distinct branches) is moot.
+    assert artifact_main_v1.branch_id == main_branch.id
+    try:
+        # create a *more recently* created head on another branch for the same key;
+        # with no explicit `revises` it infers the family head (here only main's) and
+        # joins the family on the contribution branch, leaving main's head intact.
+        ln.setup.switch(branch.name)
+        artifact_branch = ln.Artifact.from_dataframe(
+            pd.DataFrame({"inferred_feat": [22, 23]}),
+            key=key,
+            description="branch-v1",
+        ).save()
+        assert artifact_branch.branch_id == branch.id
+        assert artifact_branch.is_latest
+        assert artifact_branch.stem_uid == artifact_main_v1.stem_uid
+        artifact_main_v1.refresh_from_db()
+        # main head is preserved (cross-branch revision does not demote it)
+        assert artifact_main_v1.is_latest
+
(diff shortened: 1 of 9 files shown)


tests/core/test_is_versioned.py
```diff
@@ -309,6 +310,107 @@ def test_inferred_revises_refreshes_and_requeries_latest():
...
+        source_code="v1",
+        kind="pipeline",
+    ).save()
+    transform_v2 = ln.Transform(
+        key="stale-inferred-revises-concurrent",
+        revises=transform_v1,
+        source_code="v2",
+        kind="pipeline",
+    ).save()
+    try:
```
falexwolf, line 372: We don't use this kind of try/except to clean up the database because we delete the database.

I think it messes with readability of the tests. We could introduce this across all tests with a helper but that would be a refactor.

So, I'd rather stay consistent and unindent.

falexwolf (approved):
Looks great! Just formatting the tests consistently would be good.

tests/core/test_is_versioned.py
Koncopd: discussed, agreed that this is fine to avoid clashing with other tests in case of any tets failures.

PR #3734, opened 2026-06-11
🐛 Fix single-key dict collapse in `get_values()`

sunnyosun:
## Summary

- Fix `get_values()` returning a bare key string instead of a dict when a `dict`-dtype feature has exactly one entry (e.g. `{"HANCESTRO:0005": 1.0}` → `'HANCESTRO:0005'`).
- Root cause: `get_features_data()` unwrapped single-element collections with `next(iter(values))`. On SQLite this iterates dict keys; on PostgreSQL it had been accidentally unwrapping `ArrayAgg` singleton lists instead.
- Skip unwrapping for `list` and `dict` dtypes in `get_features_data()`, and unwrap singleton `ArrayAgg` results for those dtypes in `get_non_categoricals()` so PostgreSQL still returns bare dicts/lists.
- Apply the same dtype-aware skip in `query_set.reshape_annotate_result()` so `Record.to_dataframe()` does not collapse single-key dicts.
- Extend existing comprehensive artifact/record feature tests with a single-key `feature_dict` value instead of adding a standalone test.


lamindb/models/_feature_manager.py
```diff
@@ -552,6 +552,8 @@ def get_features_data(
             # Handle dictionary conversion
             if feature_dtype.startswith("list[cat"):
                 converted_values = values  # is already a list
+            elif feature_dtype.startswith(("list", "dict")):
```
falexwolf: Can you integrate this with line 553? Is there any reason for why `list[cat` was restricted to just `cat` before? If not, we should remove that path entirely and could remove L553 and L554.

tests/core/test_feature.py
```diff
@@ -262,3 +262,22 @@ def test_serialize_pandas_datetime_dtypes():
 
     assert serialize_pandas_dtype(datetime_series.dtype) == "datetime"
     assert serialize_pandas_dtype(datetime_tz_series.dtype) == "datetime64[ns, UTC]"
+
+
+def test_get_values_single_key_dict():
```
falexwolf: Rather than adding an extra test here, can you extend the comprehensive feature tests, please?

Here: https://github.com/laminlabs/lamindb/blob/6fc7d63af756b54e47c8b26fa5f09ac92bc60729/tests/core/test_artifact_features_annotations.py#L31

And here: https://github.com/laminlabs/lamindb/blob/6fc7d63af756b54e47c8b26fa5f09ac92bc60729/tests/core/test_record_basics.py#L629

The individual tests are impossible to keep track off.

falexwolf (changes requested):
Thank you! Please make the small changes I requested.

PR #3736, opened 2026-06-12
:sparkles: Recognize `schema.index` in the `Record` API

sunnyosun:
## Summary

- On `Schema.save()` for record-sheet schemas, when the schema hash changes because `schema.index` was set or unset, migrate row keys in bulk between `Record.name` and the `RecordJson` link table.
  - **Set index:** copy link-table values to `Record.name` and delete the link rows.
  - **Unset index:** copy `Record.name` to the link table and clear `Record.name`.
- Migration runs only when the hash changes. The previous index uid is read from persisted `_aux["af"]["3"]` inside that block.
- `schema.index` assignment is symmetric: it only sets or clears the index marker in `_aux` and does not add or remove schema members. Add features via `schema.features.add()` or `Schema(..., index=feature)` at construction. The setter requires a saved schema (`AssertionError` otherwise).
- If `Record.name` and the link-table index value both exist and differ, `schema.save()` prompts to keep `Record.name` or use feature values. Pass `index_name_conflict="keep_name"` or `"keep_feature"` to skip the prompt in scripts (`LAMIN_TESTING=true` defaults to `keep_name`).

## UX

```python
>>> schema.save()
! you updated the schema hash and might invalidate datasets that were previously validated with this schema:
                       uid                                                key                         description  ...
3264  b3BN3ADHGYtPhWkj0000  sheet_exports/input - My samples schema 2025-0...  Export of sheet O9oZfVX88ZaiEsn5:   ...
3233  GJ8FnqFCCLKMT2CM0000  sheet_exports/Uploads/Book2__2026-04-26T19-09-...  Export of sheet OgoulsHOA0mKma42:   ...

[8 rows x 22 columns]
1 sheet row(s) have both Record.name and 'name' values that differ.
  sheet My samples 2025-05 (Kzpu3Xo8g7xy), record 21977: Record.name='schmidt22_perturbseq', name='sample-001'
Keep Record.name (y) or use feature values (n)? y
```

The prompt only appears when both values exist and differ; matching values migrate silently.

## Test plan

- [x] `test_record_schema_index_stored_on_name` — constructor index; symmetric setter; `features.add` then index; migration on save; conflict resolution
- [x] `test_record_schema_index_name_conflict_resolution` — `index_name_conflict="keep_name"` / `"keep_feature"`
- [x] `test_schema_update` — mark existing member as index; add feature then index; unset keeps feature in members
- [x] `test_hash_index_feature` — schema hash unchanged for indexed schemas

### DIFF
--- a/lamindb/models/schema.py
+++ b/lamindb/models/schema.py
@@ -941,6 +941,17 @@ class Schema(SQLRecord, HasType, CanCurate, TracksRun, TracksUpdates):
 
         features_to_delete = []
         print_hash_mutation_warning = kwargs.pop("print_hash_mutation_warning", True)
+        using = kwargs.get("using") or self._state.db
+        old_index_uid = None
+        if self.pk is not None:
+            aux = (
+                Schema.objects.using(using)
+                .filter(pk=self.pk)
+                .values_list("_aux", flat=True)
+                .first()
+            )
+            if aux and isinstance(aux, dict) and "af" in aux and "3" in aux["af"]:
+                old_index_uid = aux["af"]["3"]
 
         if self.pk is not None:
             existing_features = self.members.to_list() if self.members.exists() else []
@@ -992,6 +1003,14 @@ class Schema(SQLRecord, HasType, CanCurate, TracksRun, TracksUpdates):
                         )
                 self.hash = validated_kwargs["hash"]
                 self.n_members = validated_kwargs["n_members"]
+            from .record import migrate_record_sheet_index_on_schema_save
+
+            migrate_record_sheet_index_on_schema_save(
+                self,
+                old_index_uid=old_index_uid,
+                new_index_uid=self._index_feature_uid,
+                using=using,
+            )
         super().save(*args, **kwargs)
         if hasattr(self, "_slots"):
             # analogous to save_schema_links in core._data.py
@@ -1008,7 +1027,6 @@ class Schema(SQLRecord, HasType, CanCurate, TracksRun, TracksUpdates):
             delattr(self, "_slots")
         if hasattr(self, "_features"):
             assert self.n_members > 0  # noqa: S101
-            using: bool | None = kwargs.pop("using", None)
             related_name, records = self._features
 
             # self.related_name.set(features) does **not** preserve the order
@@ -1145,8 +1163,6 @@ class Schema(SQLRecord, HasType, CanCurate, TracksRun, TracksUpdates):
     @index.setter
     def index(self, value: None | Feature) -> None:
         if value is None:
-            current_index = self.index
-            self.features.remove(current_index)
             self._index_feature_uid = value
         else:
             self.features.add(value)
--- a/lamindb/models/schema.py
+++ b/lamindb/models/schema.py
@@ -206,7 +206,7 @@ class Schema(SQLRecord, HasType, CanCurate, TracksRun, TracksUpdates):
         index: `Feature | None = None` Index feature for row keys. For `DataFrame` /
             `AnnData` curation, validates `df.index` or `obs` / `var` indices. On record
             sheets, stored on :attr:`~lamindb.Record.name` and must have `dtype=str`;
-            see :class:`~lamindb.Record`. Adds the feature to ``schema.features``.
+            see :class:`~lamindb.Record`. The index feature does not have to be a schema member.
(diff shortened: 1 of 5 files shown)


lamindb/models/schema.py
```diff
@@ -1145,8 +1163,6 @@ def index(self) -> None | Feature:
     @index.setter
     def index(self, value: None | Feature) -> None:
         if value is None:
-            current_index = self.index
-            self.features.remove(current_index)
```
falexwolf, line 1149: I wasn't aware that we'd remove the index feature from features.

This would now be a breaking change. Requires more thought.

lamindb/models/schema.py
```diff
@@ -1145,8 +1163,6 @@ def index(self) -> None | Feature:
...
             self._index_feature_uid = value
         else:
             self.features.add(value)
```
falexwolf: If we do not remove upon "unset", we can not add upon "set", right? Otherwise this is not symmetric and predictable.

falexwolf:
We have to properly document the user stories. This will also provide clarity about how the API should behave.

**Story 1: Pass a feature to `index` in the constructor.**

Should perform two actions: 

1. Add the feature object to the M2M field `schema.features`
2. Mark the index feature as the index via `schema._index_feature_uid`

**Story 2: Set or unset one of the existing features as the `index`.**

Two actions: 

1. This should set `schema._index_feature_uid`, but not touch `schema.features`. 
2. Clean up storage format of records (but not of datasets).

**Story 3: Add a new feature and mark it as an index.**

These are two actions again.

**Story 4: Remove the index feature.**

lamindb/models/schema.py
sunnyosun: yeah, that was what I mentioned yesterday.

sunnyosun:
@falexwolf can you check again?

lamindb/models/schema.py
```diff
@@ -941,6 +944,17 @@ def save(self, *args, **kwargs) -> Schema:
 
         features_to_delete = []
         print_hash_mutation_warning = kwargs.pop("print_hash_mutation_warning", True)
+        using = kwargs.get("using") or self._state.db
+        old_index_uid = None
+        if self.pk is not None:
+            aux = (
+                Schema.objects.using(using)
```
falexwolf: Can you only make this database request in case the hash changes; not pro-actively for every update operation?

lamindb/models/schema.py
```diff
@@ -1125,11 +1146,13 @@ def n(self, value: int | None) -> None:
 
     @property
     def index(self) -> None | Feature:
-        """The feature configured to act as index.
+        """The index feature, if configured.
 
-        For `DataFrame` / `AnnData` schemas, validates row indices during curation.
-        For record sheet schemas, the index feature must have `dtype=str`; see
-        :class:`~lamindb.Record`. To unset, set `schema.index` to `None`.
+        Set ``schema.index = feature`` to mark a schema member as the row key, or
```
falexwolf: Please no double back ticks.

falexwolf:
Looks great! Thanks, @sunnyosun!

One more small request: In `migrate_record_sheet_index_on_schema_save()`, can you put the INSERT and DELETE operations into one `with transaction` block?

falexwolf:
This looks good up to tests failing!

lamindb/models/schema.py
```diff
@@ -1123,13 +1161,60 @@ def n(self) -> int | None:
     def n(self, value: int | None) -> None:
         self.n_members = value
 
+    def _revalidate_unsaved_hash(self) -> None:
```
falexwolf: We already have a private function that does almost exactly this so that one could call validate whenever needed: `_validate_kwargs_calculate_hash`

Why do we have to essentially duplicate its call signature? 🤔

lamindb/models/schema.py
```diff
@@ -1145,12 +1230,23 @@ def index(self) -> None | Feature:
     @index.setter
     def index(self, value: None | Feature) -> None:
         if value is None:
-            current_index = self.index
-            self.features.remove(current_index)
             self._index_feature_uid = value
-        else:
+            if self._state.adding:
```
falexwolf: I would not support setting `schema.index` for an unsaved schema because this isn't needed and it makes the code a lot simpler.

You can just hard error:
```python
assert self._state.adding is False, "Cannot set index on unsaved schema, pass index to constructor instead."
```

PR #3740, opened 2026-06-17
🐛 Reorder schema features on update from `features` input

sunnyosun:
## Summary

Fixes schema feature ordering when updating an existing schema via the `features` argument (e.g. by name). New or reordered features were previously appended to the end because `Schema.save()` used `bulk_create(..., ignore_conflicts=True)`, which only inserts missing links and leaves existing through-table rows in place.

Now, when the desired member order differs from what is stored, all feature links for the schema are replaced in the order given by `features`.

## Test plan

- [x] `test_schema_update_reorders_features` — insert a feature in the middle and fully reorder members
- [x] `test_schema_update_implicit_through_name_equality` — existing name-based schema update flow still passes


falexwolf (approved):
Great!

PR #3741, opened 2026-06-18
:memo: Adding space and branch to signatures

ishitajain9717:
Adding Space and Branch to the Lamin registries.  

Fixes pfizer-collab/laminlabs#998

### DIFF
--- a/lamindb/models/artifact.py
+++ b/lamindb/models/artifact.py
@@ -1191,8 +1191,6 @@ class Artifact(SQLRecord, IsVersioned, TracksRun, TracksUpdates):
         overwrite_versions: `bool | None = None` Whether to overwrite versions. Defaults to `True` for folders and `False` for files.
         run: `Run | bool | None = None` The run that creates the artifact. If `False`, suppress tracking the run.
             If `None`, infer the run from the global run context.
-        branch: `Branch | None = None` The branch of the artifact. If `None`, uses the current branch.
-        space: `Space | None = None` The space of the artifact. If `None`, uses the passed `storage.space` if `storage` is passed; otherwise uses the default space (:attr:`~lamindb.setup.core.SetupSettings.space`).
         storage: `Storage | None = None` The storage location for the artifact. If `None`, uses a storage location of the `space` if `space` is passed; otherwise uses the default storage location (:attr:`~lamindb.core.Settings.storage`).
         skip_hash_lookup: `bool | None = None` Controls hash-based deduplication.
             If `None`, checks hashes for upload flows and skips hash lookup for paths already in registered storage.
@@ -1205,6 +1203,8 @@ class Artifact(SQLRecord, IsVersioned, TracksRun, TracksUpdates):
             and change it globally, e.g., `ln.settings.creation._artifact_use_virtual_keys = False`.
             If `True`, `key` is treated as metadata for versioning/querying and the on-storage path is auto-generated from the artifact `uid`.
             If `False`, `key` is treated as the concrete relative storage path for writes in managed storage.
+        branch: `Branch | None = None` A branch. If `None`, uses the current branch.
+        space: `Space | None = None` A space. If `None`, uses the current space or, if `storage` is passed, the passed `storage.space`.
 
     See Also:
         :class:`~lamindb.Storage`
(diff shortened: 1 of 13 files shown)


lamindb/models/record.py
```diff
@@ -527,9 +529,11 @@ class Record(SQLRecord, HasType, HasParents, CanCurate, TracksRun, TracksUpdates
             classifies other records).
         features: `dict[str | Feature, Any] | None = None` Lazy feature values
             to persist on `.save()` or `ln.save([...])`.
-        schema: `Schema | None = None` A schema defining allowed features for records of this type. Only applicable when `is_type=True`.
```
falexwolf, line 530: This shouldn't be removed, right? ☺️

falexwolf:
Thank you, Ishita! 🙏 😄 

One suggestion to polish.

I think it's great that branch/space come at the bottom!

image

But I think we can reduce the amount of tokens & space without losing clarity.

Instead of: "The branch of the transform. If None, uses the current branch.", I'd say: "A branch. If `None`, uses the current branch."

And same for space, I'd just say: "A space. If `None`, uses the current space."

Then things will fit onto a single line and the Args block will be more compact. And it's just a few words less to read.

falexwolf:
And can you please cross-link the original issue and the related bionty PR at the top in the description?

lamindb/models/record.py
ishitajain9717: Yes, I by mistake removed it while editing Args section, added it back in the latest commit ☺️

lamindb/models/block.py
```diff
@@ -215,8 +215,8 @@ class Block(BaseBlock, SQLRecord):
         version: `str | None = None` A version string.
         revises: `Block | None = None` An old version of the block.
         anchor: `Block | None = None` The anchor block this block attaches to.
-        branch: `Branch | None = None` The branch of the block. If `None`, uses the current branch.
-        space: `Space | None = None` The space of the block. If `None`, uses the default space (:attr:`~lamindb.setup.core.SetupSettings.space`).
+        branch: `Branch | None = None` A branch. If None, uses the current branch.
```
falexwolf: @ishitajain9717, you dropped the backticks around `None`. Please add them back.

Otherwise this is good to merge! ☺️

falexwolf:
@Koncopd, there are some strange completely unrelated JWT failures.

image

Have you ever seen them?

lamindb/models/transform.py
```diff
@@ -62,8 +64,10 @@ class Transform(SQLRecord, IsVersioned, TracksRun):
         reference: `str | None = None` A reference, e.g., a URL.
         reference_type: `str | None = None` A reference type, e.g., 'url'.
         source_code: `str | None = None` Source code of the transform.
-        revises: `Transform | None = None` An old version of the transform.
+                revises: `Transform | None = None` An old version of the transform.
```
falexwolf: @ishitajain9717, please watch out for such formatting problems and fix this one! This will mess with the docs. After that I hope the docs build and we can take a final look whether everything looks perfect!

falexwolf:
I found another two registries: `Run` & `Reference` where it was still missing. 

Now it should be homogeneous everywhere! ☺️

lamindb/models/artifact.py
```diff
@@ -1680,10 +1675,10 @@ def __init__(
         overwrite_versions: bool | None = None,
         run: Run | False | None = None,
         storage: Storage | None = None,
+        skip_hash_lookup: bool | None = None,
+        key_is_virtual: bool = True,
```
Koncopd, line 1679: this is incorrect in my opinion, we have a default option on None, now it is a dead path.

falexwolf: How do you mean? This here should never have been exposed to the user: `ln.settings.creation._artifact_use_virtual_keys`.

falexwolf: That's why I removed it. Do you still think we have a default option?

Koncopd: i mean here this becomes meaningless https://github.com/laminlabs/lamindb/blob/4e7f382872318d798d10a1935b975ceef7c37d4e/lamindb/models/artifact.py#L630

falexwolf: Now I see what you mean. Ok, I'll look back at the code. The intent is that `None` cannot be passed and `ln.settings.creation._artifact_use_virtual_keys` disappears completely: I think this would be the correct refactor.

I need to think through whether there is any need for this private setting; it changed from public to private already 3 years ago. Its use dates back so long ago that it should be safe to remove.

The only other option would be to make this a public setting, and add back `None` as possibility. 🤔

falexwolf: >i mean here this becomes meaningless

Right. GitHub is only displaying this now for me.

PR #3743, opened 2026-06-19
📝 Adding validation check and improving docs for Feature

ishitajain9717:
Fixing pfizer-collab/laminlabs#1022

### DIFF
--- a/lamindb/models/feature.py
+++ b/lamindb/models/feature.py
@@ -454,6 +454,20 @@ def serialize_dtype(
                 dtype = [dtype]
             elif not isinstance(dtype, list):
                 raise ValueError(error_message.format(dtype))
+            # A union (more than one element) is only supported for static registry
+            # types or fields - not for `Record`/`ULabel` subtype instances, which would
+            # otherwise be serialized into a malformed dtype without `|` separators
+            # (e.g. `Record[uid1]Record[uid2]`) and fail validation with a confusing error.
+            if len(dtype) > 1 and any(
+                isinstance(one_dtype, (ULabel, Record)) for one_dtype in dtype
+            ):
+                raise InvalidArgument(
+                    "Cannot create a union dtype that includes a `Record` or `ULabel` "
+                    "subtype, e.g. `dtype=[record_type_1, record_type_2]`. Unions are only "
+                    "supported for static registry types or fields, e.g. "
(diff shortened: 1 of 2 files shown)


lamindb/models/feature.py
```diff
@@ -454,6 +454,20 @@ def serialize_dtype(
...
             elif not isinstance(dtype, list):
                 raise ValueError(error_message.format(dtype))
+            # A union (more than one element) is only supported for static registry
+            # types or fields - not for `Record`/`ULabel` subtype instances, which would
+            # otherwise be serialized into a malformed dtype without `|` separators
+            # (e.g. `Record[uid1]Record[uid2]`) and fail validation with a confusing error.
+            if len(dtype) > 1 and any(
+                isinstance(one_dtype, (ULabel, Record)) for one_dtype in dtype
+            ):
+                raise InvalidArgument(
```
falexwolf, line 466: Can you please add a test for that this is raised?

lamindb/models/feature.py
```diff
@@ -979,6 +993,11 @@ class Feature(SQLRecord, HasType, CanCurate, TracksRun, TracksUpdates):
 
     **Union data types.**
 
+    Unions are only supported for static registry types or fields. You cannot
```
falexwolf: Let's make this briefer.

**Union data types.** Unions are currently only supported for static registries.

falexwolf:
Thank you! It looks great! Small comments.

tests/core/test_feature_dtype.py
```diff
@@ -111,6 +111,29 @@ def test_serialize_union_of_registries():
...
+    sample1 = ln.Record(name="sample1", is_type=True).save()
+    sample2 = ln.Record(name="sample2", is_type=True).save()
+    # raised when constructing a Feature
+    with pytest.raises(ln.errors.InvalidArgument) as error:
+        ln.Feature(name="contact", dtype=[sample1, sample2], nullable=False)
+    assert (
+        "Cannot create a union dtype that includes a `Record` or `ULabel` subtype"
+        in error.exconly()
+    )
+    # raised directly by serialize_dtype
```
falexwolf: I think that second test here isn't necessary. It's easier to read and maintain if we just have the top-level test here.

falexwolf: No, you're right. The whole file tests the underlying layer. It's better to test serialize_dtype(). I'll remove the top-level call for consistency.

PR #3750, opened 2026-06-20
🎨 Do not re-save feature set in curator

sunnyosun:
Fixes issues when the user doesn't have write access to the schema space.


falexwolf:
@sunnyosun, this would have deserved a test!

PR #3752, opened 2026-06-22
:zap: Speed up `Record.from_dataframe()` by a factor ~2 and fix casting of bool/int features with partially null entries

ishitajain9717:
Fixes a ValidationError in `Record.from_dataframe(...).save()` when a `bool/int` feature has `None` on some rows but not others. 

Improves performance of the  `Record.from_dataframe()` by a factor of 2 for 10k records: 

Before | After
--- | ---
https://lamin.ai/laminlabs/lamindata/run/hnGNmdrq8PwS244w | https://lamin.ai/laminlabs/lamindata/run/ZCnnyAOejjnbIKVa
image | Screenshot 2026-06-24 at 10 09 23 AM

## Materials

Previous PR:

- https://github.com/laminlabs/lamindb/pull/3649

Fixes:

- https://github.com/pfizer-collab/laminlabs/issues/1009

## Background

Internally the save path rebuilt the frame with `pd.DataFrame(prepared_rows)`, which let pandas re-infer and degrade those columns `(bool → object, int → float64)`, failing the `dtype` check even for correctly-typed nullable input. 

The frame is now built column-by-column using each feature's declared `dtype` (convert_to_pandas_dtype), so dtypes never degrade. Adds tests covering both the DataFrameCurator (accepts nullable boolean/Int64, rejects degraded object/float64) and the `from_dataframe()` save round-trip.

### DIFF
--- a/tests/core/test_curator_basics.py
+++ b/tests/core/test_curator_basics.py
@@ -505,8 +505,14 @@ def test_curator_partial_null_bool_int():
             "cur-count": [1, None, 3],  # float64
         }
     )
-    with pytest.raises(ln.errors.ValidationError):
+    with pytest.raises(ln.errors.ValidationError) as excinfo:
         ln.curators.DataFrameCurator(df_degraded, schema).validate()
+    assert "Column 'cur-flag' failed dtype check for 'bool': got object" in str(
+        excinfo.value
+    )
+    assert "Column 'cur-count' failed dtype check for 'int': got float64" in str(
+        excinfo.value
+    )
 
     schema.delete(permanent=True)
     flag.delete(permanent=True)
--- a/tests/core/test_curator_basics.py
+++ b/tests/core/test_curator_basics.py
@@ -480,6 +480,39 @@ def test_nullable():
     disease.delete(permanent=True)
 
 
+def test_curator_partial_null_bool_int():
+    flag = ln.Feature(name="cur-flag", dtype=bool, nullable=True).save()
+    count = ln.Feature(name="cur-count", dtype=int, nullable=True).save()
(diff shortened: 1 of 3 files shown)


tests/core/test_curator_basics.py
```diff
@@ -480,6 +480,39 @@ def test_nullable():
...
+    # bool + null -> object, int + null -> float64.
+    # Should FAIL — this is correct curator behaviour, not a bug.
+    df_degraded = pd.DataFrame(
+        {
+            "cur-flag": [True, None, False],  # object
+            "cur-count": [1, None, 3],  # float64
+        }
+    )
+    with pytest.raises(ln.errors.ValidationError):
+        ln.curators.DataFrameCurator(df_degraded, schema).validate()
```
falexwolf, line 509: Can you please add a test that makes evident which error message is raised here?

tests/core/test_curator_basics.py
```diff
@@ -480,6 +480,39 @@ def test_nullable():
...
+    flag = ln.Feature(name="cur-flag", dtype=bool, nullable=True).save()
+    count = ln.Feature(name="cur-count", dtype=int, nullable=True).save()
+    schema = ln.Schema(features=[flag, count]).save()
+
+    # Case 1: properly-typed nullable columns -> should PASS
+    # This is what a well-behaved user supplies, and what our _feature_manager fix
+    # produces after the cast. We are asserting the curator accepts these dtypes.
+    df_good = pd.DataFrame(
+        {
+            "cur-flag": pd.array([True, None, False], dtype="boolean"),
```
falexwolf, line 493: Perfect!

falexwolf (approved):
Looks great!

PR #3753, opened 2026-06-22
🐛 Fix ambiguous Record match in `sheet.to_artifact()` for duplicate names

sunnyosun:
## Summary

- Fix `ValidationError: Ambiguous match for Record '...'` when calling `Record.to_artifact()` on a sheet whose Record-type features (e.g. `sample`) reference multiple records with the same display name.
- During curator validation, if the dataframe includes `__lamindb_record_uid__`, resolve linked Record categoricals from existing `RecordRecord` rows instead of re-matching by name.
- No change to exported CSV content — only the annotation/validation path is updated.

## Test plan

- [x] `pytest tests/core/test_record_sheet_examples.py::test_to_artifact_with_duplicate_linked_record_names`

### DIFF
--- a/lamindb/curators/core.py
+++ b/lamindb/curators/core.py
@@ -177,6 +177,74 @@ Returns:
 LAMINDB_COLUMN_PREFIX_REGEX = r"^__lamindb_.*$"
 
 
+def _resolve_record_categorical_from_sheet_export(
+    cat_vector: CatVector,
+    str_values: list[str],
+) -> tuple[list[str], SQLRecordList, list[str]] | None:
+    """Resolve Record categoricals from existing DB links when sheet row uids are present."""
+    if cat_vector.feature is None or cat_vector._cat_manager is None:
+        return None
+    if cat_vector._registry.__name__ != "Record":
+        return None
+    if cat_vector.feature.pk is None:
+        return None
+
+    dataset = cat_vector._cat_manager._dataset
+    if not isinstance(dataset, pd.DataFrame):
+        return None
+
+    from lamindb.models.query_set import (
+        encode_lamindb_fields_as_columns,
+        get_default_branch_ids,
+    )
+    from lamindb.models.record import Record as RecordModel
+    from lamindb.models.record import RecordRecord
+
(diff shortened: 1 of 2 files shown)


tests/core/test_record_sheet_examples.py
```diff
@@ -549,3 +549,65 @@ def test_record_export_populates_initiated_by_run(
     finally:
         ln.context._run = None
         artifact.delete(permanent=True)
+
+
+def test_to_artifact_with_duplicate_linked_record_names():
```
falexwolf, line 554: Rather than adding a new test, can you just edit one of the existing tests? It makes it much easier to keep an overview and establish completeness of edge cases.

falexwolf:
Otherwise looks great!

lamindb/curators/core.py
```diff
@@ -1522,14 +1590,22 @@ def _add_validated(self) -> tuple[list, list]:
             validated_values = str_values  # type: ignore
             return validated_values, []
 
+        sheet_export_result = _resolve_record_categorical_from_sheet_export(
```
falexwolf, line 1593: Given this is in the generic Curator code base, shouldn't this be called something else? Actually I'm still trying to understand everything in `_resolve_record_categorical_from_sheet_export`.

falexwolf:
Actually upon pondering this, I'm not sure whether this should be done at all.

This is a lot of code that we have to maintain and that has a runtime penalty when the underlying problem is more fundamental and not addressed.

The underlying problem is that the record contains duplicated names which make it impossible to disambiguate the exported `.csv` or `.parquet` file. That kind of problem is a nightmare for any downstream use case of that sheet, isn't it? It produces the situation where you load a dataframe and then you don't know what the entities in that dataframe actually are.

So, I think that an export of a registry to a dataframe should at least print a big warning: "your dataframe won't be identifiable because of ambiguous names". Or it might should error and ask the user to make the names unambiguous. Or the export should identify entities based on their `uid` and not the names, which is the behavior if as `dtype` one chooses `Entity.uid` instead of `Entity.name/ Entity`.

What do you think, @sunnyosun?

PR #3754, opened 2026-06-22
🐛 Fix ambiguous Record match in `sheet.to_artifact()` for duplicate names

sunnyosun:
## Summary

Fixes `ValidationError: Ambiguous match for Record '...'` when calling `Record.to_artifact()` on sheets whose Record-type features (e.g. `sample`, `treatment`) link to multiple records with the same display name.

### Problem

`to_artifact()` exports a dataframe and re-validates it. Record-type feature columns only contain the linked record's **display name** (e.g. `poolsample1`), not its uid. The default curator path resolves those values by **global name lookup**, which fails when two different linked records share a name.

### Solution

Add `_resolve_record_categorical_from_sheet_export()` in `lamindb/curators/core.py`. When the dataframe still identifies **which sheet row** each value came from, resolve linked records via existing `RecordRecord` DB links instead of global name matching.

Two export shapes are supported:

| Sheet type | Row key in dataframe | DB lookup |
|---|---|---|
| No `Schema.index` (e.g. nf-core samplesheet) | `__lamindb_record_uid__` column | `record__uid__in=...` |
| With `Schema.index` (e.g. compound-treatment sheet) | `df.index` (schema index feature) | `record__name__in=...`, scoped to sheet type |

- **No change to exported CSV content** — only the annotation/validation path is updated.
- External dataframes without row keys fall back to the existing global name lookup unchanged.
- Includes an extensive docstring on the helper explaining background, both export shapes, and scope.

### Performance

Negligible for typical sheets: a targeted `RecordRecord` query per Record-type feature column on `to_artifact()` round-trips only (replaces, not stacks on, the old global name lookup). No extra DB work for non-Record categoricals or dataframes without sheet row keys.

## Test plan

- [x] `pytest tests/core/test_record_sheet_examples.py`
  - `test_record_example_compound_treatment` — indexed sheet: two Treatments named `shared_treatment`, `to_artifact()` resolves via `df.index`
  - `test_nextflow_sheet_with_samples` — no-index sheet: two BioSamples named `poolsample1`, `to_artifact()` resolves via `__lamindb_record_uid__`

Replaces https://github.com/laminlabs/lamindb/pull/3753

### DIFF
--- a/lamindb/curators/core.py
+++ b/lamindb/curators/core.py
@@ -181,7 +181,66 @@ def _resolve_record_categorical_from_sheet_export(
     cat_vector: CatVector,
     str_values: list[str],
 ) -> tuple[list[str], SQLRecordList, list[str]] | None:
-    """Resolve Record categoricals from existing DB links when sheet row uids are present."""
+    """Resolve linked ``Record`` categoricals during sheet export round-trips.
+
+    Background
+    ----------
+    Sheet rows can link to other :class:`~lamindb.Record` objects through features
+    with dtypes like ``cat[Record[BioSample].name]``. On export, only the linked
+    record's *display field* (usually ``name``) is written into the dataframe column
+    — not the linked record's uid.
+
+    When validating that export dataframe again (e.g. ``sheet.to_artifact()``), the
+    default categorical resolver looks up registry values **globally by name**. If two
+    different linked records share the same display name (e.g. two BioSamples both
+    named ``poolsample1``), that lookup is ambiguous and raises
+    :class:`~lamindb.errors.ValidationError`.
+
+    This helper avoids global name matching when the dataframe still carries enough
+    information to identify **which sheet row** each value came from. It resolves
+    linked records through existing :class:`~lamindb.models.record.RecordRecord` rows
+    in the database instead.
+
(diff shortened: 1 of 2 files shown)


sunnyosun:
From @falexwolf: https://github.com/laminlabs/lamindb/pull/3753#issuecomment-4768460114

> Actually upon pondering this, I'm not sure whether this should be done at all.
> 
> This is a lot of code that we have to maintain and that has a runtime penalty when the underlying problem is more fundamental and not addressed.
> 
> The underlying problem is that the record contains duplicated names which make it impossible to disambiguate the exported .csv or .parquet file. That kind of problem is a nightmare for any downstream use case of that sheet, isn't it? It produces the situation where you load a dataframe and then you don't know what the entities in that dataframe actually are.
> 
> So, I think that an export of a registry to a dataframe should at least print a big warning: "your dataframe won't be identifiable because of ambiguous names". Or it might should error and ask the user to make the names unambiguous. Or the export should identify entities based on their uid and not the names, which is the behavior if as dtype one chooses Entity.uid instead of Entity.name/ Entity.
> 
> What do you think, @sunnyosun?

PR #3757, opened 2026-06-22
:children_crossing: Throw an error if categorical values would be read from `RecordJson`

ishitajain9717:
Adding validation-on read in get_non_categoricals: when a value read from the JSON link table belongs to a categorical feature, raise a clear ValidationError. The error names the feature, dtype, offending value, and how to fix it via the validated API.

Fixes pfizer-collab/laminlabs#1034


falexwolf (approved):
This looks great! I'm also pretty sure that this has not performance implication. Agreed?

Good to merge if so! ☺️

PR #3761, opened 2026-06-24
🚸 Catch `NoWriteAccess` in Lamin CLI 

falexwolf:
Before | After
--- | ---
image | image

Requires:

- https://github.com/laminlabs/lamin-cli/pull/230
- https://github.com/laminlabs/lamindb-setup/pull/1354

### DIFF
--- a/lamindb/errors.py
+++ b/lamindb/errors.py
@@ -15,7 +15,6 @@ LaminDB.
 .. autoexception:: UpdateContext
 .. autoexception:: IntegrityError
 .. autoexception:: FieldValidationError
-.. autoexception:: NoWriteAccess
 .. autoexception:: BlobHashNotFound
 .. autoexception:: FileNotInDevDir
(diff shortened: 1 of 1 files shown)


lamindb/errors.py
```diff
@@ -31,6 +30,9 @@
     ObjectDoesNotExist,  # noqa: F401
 )
 
+# backward compatibility, will be removed in the future
+from lamindb_setup.errors import NoWriteAccess  # noqa: F401
```
falexwolf, line 34: FYI @Koncopd, I think that's a concern of `lamindb_setup`. Beyond semantics the reason why I'm moving is because lamin-cli still can't import lamindb as the root for latency reasons.

`NoWriteAccess` is a very basic error that needs to be caught and cast to ClickException for several CLI commands.

Koncopd: ok!

PR #3762, opened 2026-06-25
🚸 Early fallback to SQLite clone

falexwolf:
Needs:

- https://github.com/laminlabs/lamindb-setup/pull/1355


Koncopd:
Can we have a clear indication when the clone is used? Like a info log or a warning or something like this.

PR #3767, opened 2026-06-27
:zap: Speed up `Record.from_dataframe()` in presence of indexes

ishitajain9717:
In this PR we address the following issues:

1. `schema.index` called per row (4 separate call sites) `inject_index_into_feature_dict`, `strip_index_for_record_persistence`, `pop_index_from_feature_dictionary`, and `_collect_record_feature_writes` each call schema.index independently, which is a live DB query. With 10k rows × 4 calls × ~68ms network round-trip on Postgres approximately 45 minutes just for this.
2. `suggest_records_with_similar_names` called per row inside `_build_records`.  Every Record.__init__ triggers a similarity search in the DB (3 queries: _search, .first(), .exists()). For 10k rows that's ~30k extra DB queries.

## Materials

Is a follow-up optimization on:

Has been further optimized here:

- https://github.com/laminlabs/lamindb/pull/3769

Addresses:

- https://github.com/pfizer-collab/laminlabs/issues/1064

### DIFF
--- a/lamindb/models/record.py
+++ b/lamindb/models/record.py
@@ -144,10 +144,13 @@ def inject_index_into_feature_dict(record: Record, dictionary: dict[str, Any]) -
 
 
 def pop_index_from_feature_dictionary(
-    dictionary: dict[str, Any], schema: Schema
+    dictionary: dict[str, Any],
+    schema: Schema,
+    index_feature: Feature | None = None,
 ) -> tuple[str | None, dict[str, Any]]:
     """Extract index value for `record.name` and remove it from feature payload."""
-    index_feature = schema.index
+    if index_feature is None:
+        index_feature = schema.index
     if index_feature is None:
         return None, dictionary
     index_name = index_feature.name
@@ -499,7 +502,7 @@ class RecordBatch:
 
             if index_feature is not None and self._resolved_type.schema is not None:
                 name_from_features, features = pop_index_from_feature_dictionary(
-                    features, self._resolved_type.schema
+                    features, self._resolved_type.schema, index_feature=index_feature
                 )
                 if name is None:
                     name = name_from_features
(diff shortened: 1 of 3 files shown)


lamindb/models/record.py
```diff
@@ -464,10 +469,14 @@ def type(self) -> Record:
...
         import pandas as pd
 
+        from lamindb import settings
+
         index_feature = get_type_schema_index(self._resolved_type)
         records: list[Record] = []
         work_df = dataframe_for_record_batch(self._df, index_feature)
         row_dicts = work_df.to_dict(orient="records")
+        _search_names = settings.creation.search_names
+        settings.creation.search_names = False
```
falexwolf, line 479: Rather than doing this I think we should expose `search_names` as an argument and pass that to the constructor as `False` in the context of this batched function.

Do you know where he problem is caused between L480 and L514? I don't see where the constructor with `search_names` would be called. Maybe in: `pop_index_from_feature_dictionary`?

falexwolf:
This looks great, @ishitajain9717! Thank you!

My only question is about `search_names` in the comments up there.

And about adjusting our benchmarking script to also include an index so that we trigger these bad paths (https://laminlabs.slack.com/archives/D09M77R7AL8/p1782711795179939?thread_ts=1782561047.150959&cid=D09M77R7AL8).

You can run the example both on main and on this branch and report back the results in the PR description.

lamindb/models/record.py
ishitajain9717: The constructor is called at L513 I guess `records.append(self._cls(name=name, **record_kwargs))` Thanks for pointing out. Rather than mutating `settings.creation.search_names` globally (which affects the entire process for the duration of the loop and requires save/restore), we can pass `_search_names=False` directly as a kwarg to the constructor. The constructor would pop it from kwargs and use it as the local override, falling back to `settings.creation.search_names` if not provided.

falexwolf: Yes, that sounds like the right thing. Can you make that small refactor in another PR once I merged my big PR?

https://github.com/laminlabs/lamindb/pull/3768

ishitajain9717: Yes  I can do that

PR #3769, opened 2026-06-29
⚡️ Additional performance improvements to `Record.from_dataframe()` if `schema.index` is present

ishitajain9717:
In this PR, we make the following changes:

1. Preventing the generation of SQL for all columns, when schema.index is present, which brings down the processing time from 64s to 36s for 10k records. 
2. Passing `pass _search_names=False` directly as a kwarg to the constructor. The constructor would pop it from kwargs and use it as the local override, falling back to settings.creation.search_names if not provided.


falexwolf:
Thanks! This looks great!

PR #3770, opened 2026-06-29
:children_crossing: Introduce `Feature(..., related_name="...")`

falexwolf:


falexwolf:
Decided against this to keep the API simpler. Will instead document how backward queries without introducing this concept.

PR #3773, opened 2026-06-29
Added file called tables.md

Raaghav-Pillai:


falexwolf:
@Raaghav-Pillai, you have write access to this repo. You don't need to make a fork. ☺️ 

If you make a PR on the source repo, the CI will run, your guide will be executed, and the docs will build and you can look at the result.

Can you please move the tables/dataframe/parquet sections from `arrays.md` into `tables.md` and make sure that it still flows well?

You can add `tables` to the TOC here: https://github.com/laminlabs/lamindb/blob/7990b64bed607752a7167249f525d2f10b12ad68/docs/query-search.md?plain=1#L11

The logical flow should be registries -> tables -> arrays. Does that make sense?

PR #3776, opened 2026-06-30
📝 A dedicated `tables` query guide

Raaghav-Pillai:
Related PRs:

- https://github.com/laminlabs/lamin-blog/pull/55
- https://github.com/laminlabs/lamindb/pull/3807


docs/arrays.md
```diff
@@ -6,120 +6,62 @@ execute_via: python
...
-with collection.open(engine="polars", use_fsspec=True) as lazy_df:
-    display(lazy_df.collect().to_pandas())
-```
+For tabular data — DataFrame, Parquet, querying with PyArrow, Polars, DuckDB, Iceberg, or LanceDB — see [Query tables](tables.md). This guide covers array-like data: AnnData, SpatialData, and generic HDF5.
 
 ## AnnData
 
 We'll need some test data:
 
-```python
```
falexwolf, line 89: Can you add these `python` markers back please?

docs/arrays.md
```diff
@@ -150,91 +92,87 @@ artifact = ln.Artifact(
...
 You can subset it and read into memory as an actual `AnnData`:
 
-```python
+```
 table_subset = table[table.obs["clone"] == "diploid"]
 
 table_subset
 ```
 
-<!-- #region -->
```
falexwolf, line 205: This is important to keep.

docs/tables.md
```diff
@@ -0,0 +1,194 @@
+---
+execute_via: python
+---
+
+# Query tables from storage
+
+This guide walks through querying tabular data stored as Parquet — streaming directly from disk or cloud storage with PyArrow, Polars, DuckDB, Iceberg, or LanceDB.
+
+```
+# replace with your username and S3 bucket
```
falexwolf: Ideally you just remove this. Since this is a pure read-only query we don't need a database and we also don't need to authenticate!

docs/tables.md
```diff
@@ -0,0 +1,194 @@
...
+!lamin login testuser1
+!lamin init --storage s3://lamindb-ci/test-tables
+```
+
+Import lamindb and track this notebook.
+
+```
```
falexwolf: You're lacking the `python` marker.

docs/tables.md
```diff
@@ -0,0 +1,194 @@
...
+import lamindb as ln
+
+ln.track()
```
falexwolf: Can you add a comment that `ln.track()` is optional?

docs/tables.md
```diff
@@ -0,0 +1,194 @@
...
+db = ln.DB("laminlabs/lamindata")  # we'll pull datasets from there
+```
+
+## Streaming from a single artifact
+
+A dataframe stored as sharded `parquet`.
+
+```
+artifact = db.Artifact.get(key="sharded_parquet")
+```
```
falexwolf: Why a new paragraph here? 🤔

docs/tables.md
```diff
@@ -0,0 +1,194 @@
...
+Import lamindb and track this notebook.
+
+```
+import lamindb as ln
+
+ln.track()
+db = ln.DB("laminlabs/lamindata")  # we'll pull datasets from there
+```
+
+## Streaming from a single artifact
```
falexwolf: Can we use an example that's not a folder of sharded files? That's already such an advanced concept. I think it makes more sense to just take a single parquet file (feel free to use one from your blog post) and then query that directly. And then continue with showing that this also works across collections of parquet files.

docs/tables.md
```diff
@@ -0,0 +1,194 @@
...
+
+```
+dataset
+```
+
+```
+dataset.head(5).to_pandas()
+```
+
+## Streaming from a set of artifacts
```
falexwolf: I think this should be integrated with the previous section or be under one subsection because it's so similar and people should go away with the feeling that "this is really easy".

docs/tables.md
```diff
@@ -0,0 +1,194 @@
...
+dataset
+```
+
+Once you have a storage-backed dataset, you can query it like this:
+
+```
+dataset.to_table().to_pandas()
+```
+
+## Filtering with predicate pushdown
```
falexwolf: This subsection should somehow group in a "queries" section. I really like the ones from your blog post. It'd be cool to show a slight variation? Starting with something as simple as we're seeing here but ending in something more complicated and linking to benchmarks?

docs/tables.md
```diff
@@ -0,0 +1,194 @@
...
+## Choosing a query engine
+
+By default `Artifact.open()` and `Collection.open()` use `pyarrow` to lazily open dataframes. `polars` can also be used by passing `engine="polars"`. Note also that `.open(engine="polars")` returns a context manager with [LazyFrame](https://docs.pola.rs/api/python/stable/reference/lazyframe/index.html).
+
+```
+with collection.open(engine="polars", use_fsspec=True) as lazy_df:
+    display(lazy_df.collect().to_pandas())
+```
+
+LaminDB artifacts are plain Parquet files on S3, so any engine that reads Parquet works. DuckDB, Iceberg, and LanceDB are also supported — you pick based on your query pattern, not on how the data was stored.
```
falexwolf, line 84: Nicely put! 😄

docs/tables.md
```diff
@@ -0,0 +1,194 @@
...
+    "local", uri="sqlite:///iceberg_catalog.db", warehouse="s3://your-bucket/iceberg_warehouse"
+)
+catalog.create_namespace("demo")
+table = catalog.create_table("demo.data", schema=arrow.schema)
+table.append(arrow)
+
+result = table.scan(row_filter=NotNull("disease")).to_arrow().to_pandas()
+```
+
+Note: a SQLite catalog is used here for portability. Production deployments would use a Glue or REST catalog.
```
falexwolf: I think we can remove that note in the context of the Lamin docs.

falexwolf:
Thanks, @Raaghav-Pillai! This already looks very promising! 😄 

The CI didn't complete because HF is a bit unreliable:

image

In such cases please simply re-run the job.

falexwolf:
I left some comments, which would be great if you addressed them. And then we'll look at this together again with the blog post.

falexwolf:
I think we should call this guide `parquet.md` and the title "Query tabular data in parquet files".

docs/arrays.md
```diff
@@ -1,10 +1,6 @@
----
-execute_via: python
```
falexwolf, line 2: @Raaghav-Pillai, you shouldn't remove this but keep executing the guide because otherwise examples will break!

docs/arrays.md
```diff
@@ -1,10 +1,6 @@
...
----
-
 # Stream datasets from storage
```
falexwolf: This heading should now read "Stream arrays from storage".

docs/arrays.md
```diff
@@ -18,68 +14,8 @@ Import lamindb and track this notebook.
...
-```python
-dataset.to_table().to_pandas()
-```
-
-By default `Artifact.open()` and `Collection.open()` use `pyarrow` to lazily open dataframes. `polars` can be also used by passing `engine="polars"`. Note also that `.open(engine="polars")` returns a context manager with [LazyFrame](https://docs.pola.rs/api/python/stable/reference/lazyframe/index.html).
-
-```python
-with collection.open(engine="polars", use_fsspec=True) as lazy_df:
-    display(lazy_df.collect().to_pandas())
+ln.track()  # this step is optional
```
falexwolf: I'm tempted to say we remove this given the guide will work just fine without `ln.track()`.

docs/parquet.md
```diff
@@ -0,0 +1,178 @@
+# Query tables from storage
```
falexwolf: I'd say: "Query tables in storage"

docs/parquet.md
```diff
@@ -0,0 +1,184 @@
...
+```
+
+These filters run the same way on every engine below. For heavier patterns — schema evolution, time travel, appends, and concurrent writers — the [LaminDB lakehouse benchmark](https://lamin.ai/blog/lakehouse-benchmark) walks through all six operations end to end across the five engines on a shared genomics dataset.
+
+## Choosing a query engine
+
+By default `Artifact.open()` and `Collection.open()` use `pyarrow` to lazily open dataframes. `polars` can also be used by passing `engine="polars"`. Note also that `.open(engine="polars")` returns a context manager with [LazyFrame](https://docs.pola.rs/api/python/stable/reference/lazyframe/index.html).
+
+```python
+with collection.open(engine="polars", use_fsspec=True) as lazy_df:
```
Koncopd: @Raaghav-Pillai could you try without use_fsspec=True here?

docs/tables.md
```diff
@@ -0,0 +1,184 @@
...
+
+s3_paths = [str(a.path) for a in collection.ordered_artifacts.all()]
+con.execute(f"CREATE OR REPLACE VIEW data AS SELECT * FROM read_parquet({s3_paths})")
+
+result = con.execute("SELECT * FROM data WHERE disease IS NOT NULL").df()
+```
+
+:::::
+
+:::::{tab-item} Iceberg
```
Koncopd, line 129: @falexwolf do we really want this here?

falexwolf: I think that a small example in a tab re Iceberg provides interesting context. I've had these conversations with different teams that ask about it. Where else in the docs would you put it?

PR #3782, opened 2026-07-02
:children_crossing: Allow only file and folder suffixes from the pre-defined lists

Koncopd:
This PR makes validating suffixes for the `Artifact.suffix` field more stringent. For docs & background, see:

- https://github.com/laminlabs/lamindb/pull/3784

Needs: 

- https://github.com/laminlabs/lamindb-setup/pull/1357

Addresses: 

- https://github.com/pfizer-collab/laminlabs/issues/1074


lamindb/core/storage/__init__.py
```diff
@@ -15,7 +15,6 @@
 
 from lamindb_setup.core.upath import LocalPathClasses, UPath, infer_filesystem
 
-from ._valid_suffixes import VALID_SUFFIXES
```
falexwolf, line 18: This breaks a public documented API, it has to remain here as a re-export from `lamindb_setup` at least for backward compat until lamindb v3. See the docs above.

lamindb/core/storage/_valid_suffixes.py
```diff
@@ -1,21 +0,0 @@
-from __future__ import annotations
```
falexwolf, line 1: Great that this file disappeared!

lamindb/models/artifact.py
```diff
@@ -1964,11 +1969,7 @@ def set_private_attributes():
         if is_automanaged_path and _is_internal_call:
             kwargs["_key_is_virtual"] = True
             assert _s().AUTO_KEY_PREFIX in kwargs["key"]  # noqa: S101
-            uid = (
-                kwargs["key"]
-                .replace(_s().AUTO_KEY_PREFIX, "")
-                .replace(kwargs["suffix"], "")
-            )
+            uid = PurePosixPath(kwargs["key"]).stem
```
falexwolf, line 1975: That is a LOT prettier ☺️

falexwolf:
Can you cross-ref the original issue and document the user story? Also, we should document this more prominently in the docs along the lines of what we discussed in the call so that everybody understands that the lamindb `suffix` is different from the `pathlib` suffixes in that it goes through an additional validation step.

lamindb/core/storage/__init__.py
Koncopd: fixed

lamindb/models/artifact.py
Koncopd: thx

PR #3785, opened 2026-07-04
🐛 Retain empty columns in `Record.from_dataframe()`

ishitajain9717:
When calling `ln.Record.from_dataframe(df, type=st).save()`, any DataFrame column where every cell is `NaN` was silently dropped. Records were saved with no trace that the feature existed, even though it was declared in the schema. This made it impossible to distinguish "feature has no data yet" from "feature was never part of this batch".

Addresses:

- https://github.com/pfizer-collab/laminlabs/issues/1082


lamindb/models/_feature_manager.py
```diff
@@ -2299,7 +2299,10 @@ def bulk_set_features_in_records(records: Iterable[Record]) -> None:
 
     data: dict[str, pd.Series] = {}
     for column in ordered_columns:
-        values = [row.get(column, pd.NA) for row in prepared_rows]
+        values = [
+            pd.NA if v is None else v
```
falexwolf, line 2305: Could you add a comment here that explains the rationale behind each case?

The code here looks not very performant (also before your edit) but I guess it's alright given database connections are what drives latency.

lamindb/models/record.py
```diff
@@ -495,6 +496,8 @@ def _build_records(self) -> list[Record]:
                 if pd.api.types.is_scalar(value) and pd.isna(value):
                     continue
                 features[key] = value
+            for col in all_null_cols:
+                features[col] = None
```
falexwolf: What exactly is this going to do? We're not storing `None` values in `RecordJson`, we're just leaving it empty. Is this still the case?

tests/core/test_record_basics.py
```diff
@@ -189,6 +189,50 @@ def test_record_from_dataframe_partial_null_bool_int():
     label.delete(permanent=True)
 
 
+def test_record_from_dataframe_all_null_column_preserved_in_features():
```
falexwolf: Rather than adding a new test, is it possible to extend an existing test and just use a dataframe that has all null values in a column in the existing test?

falexwolf:
Thanks, @ishitajain9717! This looks very promising given how minimal the changes are. I still have a few comments that you can hopefully address.

lamindb/models/record.py
ishitajain9717: Yes, that is still the case. Here if a column has no values i.e. its empty the feature itself gets dropped. Computed `all_null_cols` once before the row loop (not inside it, to avoid recomputing per row). After the per-cell loop, explicitly added each entirely-null column with None. Also used `None` instead of `pd.NA` as it is `JSON-serializable`

lamindb/models/_feature_manager.py
ishitajain9717: We are doing this because here `pd.NA` is accepted for the data types `StringDtype`, `BooleanDtype`, `Int64Dtype`

tests/core/test_record_basics.py
ishitajain9717: Yes, working on it

lamindb/models/_feature_manager.py
ishitajain9717: Two changes:

`Rationale behind each change` -> also added to PR description
Change 1 — `record.py _build_records`: compute `all_null_cols` and append `None`

Problem avoided: When a DataFrame column is entirely `NaN`, the per-cell loop skips every value with continue, so that feature never enters the features dict. The record is built with no knowledge that the column existed,  it is silently dropped.

What we did: Computed `all_null_cols` once before the row loop (not inside it, to avoid recomputing per row). After the per-cell loop, explicitly added each entirely-null column with None. 
Why None and not pd.NA: None is JSON-serializable. pd.NA is not,  it would break any code path that serialises the features dict.

Change 2 — `_feature_manager.py` Series construction: `None → pd.NA`

Problem avoided: None from Change 1 flows through into prepared_rows and then into the Series built for the validation DataFrame. Python None is not the correct missing-value sentinel for pandas extension dtypes (StringDtype, BooleanDtype, Int64Dtype), those expect pd.NA. It triggers pandera dtype-check failures at validation time.

What we did: This happens only inside the validation DataFrame build, the features dict itself still holds None (keeping it JSON-serializable).

falexwolf: Thanks! Can you also add a much shorter version as a comment into the code, please?

lamindb/models/record.py
falexwolf: > Also used None instead of pd.NA as it is JSON-serializable

What I'm trying so say above is that we're not serializing `None` into the link table. 

> We're not storing None values in RecordJson, we're just leaving it empty.

Can you please assert in your test that this is still not happening?

tests/core/test_record_basics.py
```diff
@@ -179,6 +184,13 @@ def test_record_from_dataframe_partial_null_bool_int():
     assert exported["from-df-flag"].dtype.name == "boolean"
     assert exported["from-df-count"].dtype.name == "Int64"
 
+    # entirely-null column: present in export but nothing stored in RecordJson
+    from lamindb.models.record import RecordJson
+
+    assert "from-df-empty" in exported.columns
+    assert exported["from-df-empty"].isna().all()
+    assert RecordJson.objects.filter(feature=empty).count() == 0
```
falexwolf, line 192: Great! 😄

tests/core/test_record_basics.py
```diff
@@ -179,6 +184,13 @@ def test_record_from_dataframe_partial_null_bool_int():
     assert exported["from-df-flag"].dtype.name == "boolean"
     assert exported["from-df-count"].dtype.name == "Int64"
 
+    # entirely-null column: present in export but nothing stored in RecordJson
+    from lamindb.models.record import RecordJson
+
+    assert "from-df-empty" in exported.columns
```
falexwolf, line 190: Perfect!

falexwolf:
Thanks for addressing everything! This looks great now 😄 

Just this PR should never have touched `tests/core/test_transfer.py` since the breakage there was unrelated to the concern of this PR and was fixed upstream.

I'll fix this and merge! ☺️

falexwolf:
For the documentation of this PR, it would be great to rerun the export:

This should also have been documented in the context of this PR:

- https://github.com/laminlabs/lamindb/pull/3767

The benchmarking script should be updated so that the test data it generates encompasses these new cases that your two PRs hit. I think you did this in v9: https://lamin.ai/laminlabs/lamindata/transform/JuJZZEsit1KV0009

image

Please just beware that you keep the original layout of the Imports registry.

That is, this type here: https://lamin.ai/laminlabs/lamindata/transform/JuJZZEsit1KV0009#L23

Should nest under `Imports`: https://lamin.ai/laminlabs/lamindata/record/W3WdiFRZTvTJajNp

image

It's important that the records of `lamindata` are cleanly organized. We don't want sheets at the root level ☺️

ishitajain9717:
I actually edited as the earlier test had an assertion error, where `lamindata` instance had some extra features, so I just added those

falexwolf:
> > Just this PR should never have touched tests/core/test_transfer.py since the breakage there was unrelated to the concern of this PR and was fixed upstream.
>
> I actually edited as the earlier test had an assertion error, where lamindata instance had some extra features, so I just added those

I know why you did it. Just what you did was the wrong thing because you mixed concerns. This was the right thing to do: https://github.com/laminlabs/lamindb/commit/5daf6d5eab377539e3a79fdece30d0ed05ab49f4 ☺️

PR #3792, opened 2026-07-07
🚸 Print a warning if `overwrite_versions` is true

falexwolf:
It might be surprising in particular for new users to see that.


falexwolf:
The annoying thing is if you ingest 100 folders and then you see this 100 times.

This is what Python's warning module can help avoid. But I'm also not sure whether we need to raise an official warning; this is just a logging message to the user IMO.

falexwolf:
We might actually have had this warning before and removed it for this this reason. 🤔

sheetalgiri:
I only see the warning if I explicitly set the parameter which is good but for folders where it is true by default (probably this is set somewhere later in the code) the warning doesn't show up
Screenshot 2026-07-07 at 16 44 12

PR #3797, opened 2026-07-08
🚸 Do not treat the bionty default space differently from other entities

falexwolf:
There was a special case in which the configured default space would be ignored for `bionty` registries and the `Storage` registry.

### DIFF
--- a/lamindb/models/sqlrecord.py
+++ b/lamindb/models/sqlrecord.py
@@ -1167,8 +1167,6 @@ class BaseSQLRecord(models.Model, metaclass=Registry):
                 if (
                     issubclass(self.__class__, SQLRecord)
                     and self.__class__.__name__ != "Storage"
-                    # do not save bionty entities in restricted spaces by default
-                    and self.__class__.__module__ != "bionty.models"
                 ):
                     from lamindb import context as run_context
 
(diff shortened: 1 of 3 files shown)


lamindb/models/sqlrecord.py
```diff
@@ -1167,8 +1167,6 @@ def resolve_fk_or_id(field_name: str) -> bool:
                 if (
                     issubclass(self.__class__, SQLRecord)
                     and self.__class__.__name__ != "Storage"
```
falexwolf: Now I'm confused that we're excluding `Storage`, here, @Koncopd.

I think that should also just be treated like any other registry. I remember that early on (probably more than a year ago) there were issues with initializing storage locations but I figure that back then we didn't have the clarity we achieved around mid 2025, when we realized that a space should govern the permissions to a storage location.

falexwolf: I think also this line should probably be removed. No special implicit rules for anything.

falexwolf: I went ahead with this.

Koncopd: i don't think i ever added this, so i can't really answer.

Koncopd: Are there any unintended implications? I am even a bit surprised we had this for stroage actually.

PR #3800, opened 2026-07-11
✨ Adding artifact.get(uid) in error

ishitamjain:
Fixing: pfizer-collab/laminlabs#95

In this PR the error message was updated to show `did you mean ln.Artifact.get(uid)` if `ln.Artifact(path)` is a uid along with `FileNotFoundError`

### DIFF
--- a/lamindb/models/artifact.py
+++ b/lamindb/models/artifact.py
@@ -223,7 +223,10 @@ def process_pathlike(
     if not skip_existence_check:
         try:  # check if file exists
             if not filepath.exists():
-                raise FileNotFoundError(filepath)
+                hint = ""
+                if Artifact.filter(uid=filepath.name).first() is not None:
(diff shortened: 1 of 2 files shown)


lamindb/models/artifact.py
```diff
@@ -233,20 +222,13 @@ def process_pathlike(
...
-            ):
-                raise
-            logger.warning(f"failed to check existence of {filepath}, proceeding: {e}")
-        else:
-            if not exists:
-                raise FileNotFoundError(filepath)
+            if not filepath.exists():
+                hint = ""
+                if Artifact.filter(uid=filepath.name).first() is not None:
+                    hint = f', did you mean Artifact.get("{filepath.name}")?'
```
Koncopd, line 228: Should probably check startswith then

falexwolf:
Let's not do this.

Issue #3744, opened 2026-06-19
✨ Enable an arg that allows skipping tracking in `@ln.flow()`


falexwolf:
Sometimes despite decorating a function we want to skip tracking.

falexwolf:
I forgot why I made the note. 🤔 

We can re-open this if somebody can come up with a good reason that would justify the increased complexity in decorator signature.

PR #3806, opened 2026-07-14
♻️ Add skip_track parameter to tracked decorators and flow function

Raaghav-Pillai:
Fix for Issue #3744


falexwolf:
Hmmmm, I would not want to merge this because it makes the API complicated. I know I made this issue at the time but I forgot to document it thoroughly. Let's first have a discussion whether the issue is legit. It was more a note at the time.

PR #3808, opened 2026-07-15
🐛 Schema members should take precedence over flexible features

ishitajain9717:
Here we make minor fixes to make sure, if a `schema` with `schema.flexible is True` has feature names identical to concrete features of the schema (those in `schema.members`), the schema members are given precedence. 

Fixes:

- https://github.com/pfizer-collab/laminlabs/issues/1112

### DIFF
--- a/lamindb/curators/core.py
+++ b/lamindb/curators/core.py
@@ -681,20 +681,12 @@ class ComponentCurator(Curator):
                 # Schema features take precedence over flexible-fetched features
                 # of the same name. Remove any flexible-fetched feature whose name
                 # is explicitly defined by the schema, then append the schema version.
-                schema_feature_names = {
-                    f.name for f in schema_features if isinstance(f, Feature)
-                }
-                features_filtered: list[Feature] = [
+                schema_feature_names = {f.name for f in schema_features}  # type: ignore
+                features = [
                     f
                     for f in features
-                    if isinstance(f, Feature) and f.name not in schema_feature_names
-                ]
-                # schema_features is expected to be a list[Feature] (members of Schema)
-                # but we still guard via isinstance for type-checking.
-                features_filtered.extend(
-                    f for f in schema_features if isinstance(f, Feature)
-                )
-                features = features_filtered
+                    if f.name not in schema_feature_names  # type: ignore
+                ] + schema_features  # type: ignore
             else:
                 features.extend(schema_features)
         else:
@@ -1798,21 +1790,17 @@ class CatVector:
                     and self._schema
                     and self._schema.n_members
                 ):
-                    schema_members = [
-                        f
-                        for f in self._schema.members.to_list()  # type: ignore[attr-defined]
-                        if isinstance(f, Feature)
-                    ]
+                    schema_members = cast(
+                        list[Feature],
+                        self._schema.members.to_list(),  # type: ignore
+                    )
                     schema_feature_by_name = {f.name: f.id for f in schema_members}
-                    filtered_records: list[Any] = []
-                    for r in records:
-                        if isinstance(r, Feature):
-                            if (
-                                r.name in schema_feature_by_name
-                                and r.id != schema_feature_by_name[r.name]
-                            ):
-                                continue
-                        filtered_records.append(r)
+                    filtered_records: list[Any] = [
+                        r
+                        for r in cast(list[Feature], records)
+                        if r.name not in schema_feature_by_name
+                        or r.id == schema_feature_by_name[r.name]
+                    ]
                     records = cast(
                         SQLRecordList[Any],
                         SQLRecordList(filtered_records),
--- a/lamindb/curators/core.py
+++ b/lamindb/curators/core.py
@@ -660,6 +660,7 @@ class ComponentCurator(Curator):
             dataset=dataset, schema=schema, require_saved_schema=require_saved_schema
         )
 
+        self._features = None
         categoricals = []
         features = []
         feature_ids: set[int] = set()
@@ -691,6 +692,8 @@ class ComponentCurator(Curator):
                 features.extend(schema_features)
         else:
             assert schema.itype is not None  # noqa: S101
+
+        self._features = features
 
         pandera_columns = {}
         self._pandera_schema = None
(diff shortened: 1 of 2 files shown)


lamindb/curators/core.py
```diff
@@ -678,11 +678,23 @@ def __init__(
...
+                }
+                features_filtered: list[Feature] = [
+                    f
+                    for f in features
+                    if isinstance(f, Feature) and f.name not in schema_feature_names
+                ]
+                # schema_features is expected to be a list[Feature] (members of Schema)
+                # but we still guard via isinstance for type-checking.
+                features_filtered.extend(
+                    f for f in schema_features if isinstance(f, Feature)
```
falexwolf: No need to add `isinstance(f, Feature)`, see the previous version of the code.

lamindb/curators/core.py
```diff
@@ -678,11 +678,23 @@ def __init__(
...
+                # Schema features take precedence over flexible-fetched features
+                # of the same name. Remove any flexible-fetched feature whose name
+                # is explicitly defined by the schema, then append the schema version.
+                schema_feature_names = {
+                    f.name for f in schema_features if isinstance(f, Feature)
+                }
+                features_filtered: list[Feature] = [
+                    f
+                    for f in features
+                    if isinstance(f, Feature) and f.name not in schema_feature_names
```
falexwolf: No need to add `isinstance(f, Feature)`, see the previous version of the code.

lamindb/curators/core.py
```diff
@@ -678,11 +678,23 @@ def __init__(
...
             if feature_ids:
-                features.extend(
-                    feature
-                    for feature in schema_features
-                    if feature.id not in feature_ids  # type: ignore
+                # Schema features take precedence over flexible-fetched features
+                # of the same name. Remove any flexible-fetched feature whose name
+                # is explicitly defined by the schema, then append the schema version.
+                schema_feature_names = {
+                    f.name for f in schema_features if isinstance(f, Feature)
```
falexwolf: No need to add `isinstance(f, Feature)`, see the previous version of the code.

falexwolf:
Thanks, @ishitajain9717! 🙏 

I think what you're doing here 💯 achieves the goal.

So, just a thought on code organization.

First, the `ComponentCurator.__init__()` isn't as parsimonious and elegant as it could be; we'll have to accept this for the purpose of this PR. It would likely make sense to properly refactor it to make it as minimal and performant as possible. But let's accept the state of the code.

What you're adding to `__init__()` is indeed pretty minimal and amounts to a simple filtering step, filtering out erroneously fetched flexible features from the top of the implementation. You're just making it more complicated than needed by adding `isinstance(f, Feature)`, everywhere. I left comments for this.

Overall one might consider to switch the order in the feature: First consider concrete features (those in `schema.members`) and only if those don't cover the dataset attempt fetching additional ones. This would be more logical and performant but we shouldn't attempt this here in the PR to not make things too complicated.

---

Now what I don't understand at first review is why we need to do more than the changes in `__init__()`? The validation is conditioned on what we set up there, isn't it? It seems that all of this here shouldn't be needed.

```python
                # If the schema explicitly defines a Feature with this name, prefer that
                # specific Feature instance over any duplicates that may exist in the DB.
                # This prevents ambiguity errors for schema-defined column/features in
                # flexible schemas when duplicate root-level Features exist.
                if (
                    registry.__name__ == "Feature"
                    and self._schema
                    and self._schema.n_members
                ):
                    schema_members = [
                        f
                        for f in self._schema.members.to_list()  # type: ignore[attr-defined]
                        if isinstance(f, Feature)
                    ]
                    schema_feature_by_name = {f.name: f.id for f in schema_members}
                    filtered_records: list[Any] = []
                    for r in records:
                        if isinstance(r, Feature):
                            if (
                                r.name in schema_feature_by_name
                                and r.id != schema_feature_by_name[r.name]
                            ):
                                continue
                        filtered_records.append(r)
                    records = cast(
                        SQLRecordList[Any],
                        SQLRecordList(filtered_records),
                    )
```

It might be needed because there may be an architectural flaw in the class. Let me check.

falexwolf:
I see. Because we have `flexible=True`, we might want to _parse_ information about features that aren't define in the schema.

We need to ensure that during that parsing step we don't overwrite feature objects that we fetched during the scoped fetch.

We already have similar logic a bit further up, above where you added your diff:

```python
                    if self._schema and self._schema.n_members:
                        type_ids = {
                            m.type_id
                            for m in self._schema.members
                            if m.type_id is not None
                        }
                        if type_ids:
                            qs = registry.filter(
                                Q(type_id__in=type_ids) | Q(type_id__isnull=True)
                            )
```

Now I wonder whether we can do something similar, i.e., construct the `qs` in a way that doesn't overwrite the schema members.

falexwolf:
After reviewing `_add_validated()`, it appears that it doesn't account for `schema.members` other than for feature type resolution (what I pasted up there).

Given that `__init__()` already resolves all relevant features correctly, I think we should pass that information to `_add_validated()`. In the best case we don't even call `_add_validated()` for feature parsing anymore because we already did that in `__init__()`! 

In plain English: While there is a overlap between validating column names of a dataframe and validating the categories in a categorical column, we're doing a lot of work for the former already in `__init__()` and we might neither have to do validation nor parsing again later on in the flow.

Can you check whether we can skip `_add_validated()` altogether by remembering the feature objects? And then populating the following outcome of `_add_validated()` right in the `CatVector.__init__()`?

```python
        self.records = all_records
        # validated values, non-validated values
        return all_validated, remaining_values
```

ishitajain9717:
I think we can only skip in some cases, when values are already records/features (i.e. SQLRecord inputs) or empty.
But not when values are strings, then we can't because strings require value→record resolution. Does it makes sense ☺️

falexwolf:
> I think we can only skip in some cases, when values are already records/features (i.e. SQLRecord inputs) or empty.
But not when values are strings, then we can't because strings require value→record resolution. Does it makes sense ☺️

Values are always strings. Features are always `Feature` objects. I'll briefly call you. ☺️

falexwolf:
As discussed on the call, it's not easy to do a point refactor. I reverted my attempt.

Conclusion: Let's merge your fix @ishitajain9717 and then take time to overhaul the entire curator code base which became unruly over the past 2 years.

PR #3816, opened 2026-07-21
🦺 Check key on init of `sqlrecord` and saving `artifact`

Koncopd:
Keys on `sqlrecords` (`artifacts`, `transforms`, etc.) are now validated when set, so unsafe path-like keys are rejected early.

**What’s allowed:** relative, `/`-separated keys such as `myfolder/my file.parquet` or `Introduction v2`.

**What’s rejected:** empty keys, backslashes, leading/trailing/`//` slashes, and `.` / `..` segments (`./`, `../`, `/../`, etc.).

Validation runs on `sqlrecord` creation and when changing an `artifact` `key` on `save`.

### DIFF
--- a/lamindb/models/artifact.py
+++ b/lamindb/models/artifact.py
@@ -76,6 +76,7 @@ from .sqlrecord import (
     Space,
     SQLRecord,
     _get_record_kwargs,
+    check_key,
 )
 from .storage import Storage
 from .ulabel import ULabel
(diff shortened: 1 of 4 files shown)


lamindb/models/artifact.py
```diff
@@ -3326,6 +3327,7 @@ def save(
             new_key = self.key
             if new_key is None:
                 raise InvalidArgument("Cannot update an artifact key to None.")
+            check_key(new_key)
```
falexwolf, line 3330: Can we call it `validate_key()`? This would be standard name for it. `check_key()` is less of a convention, I think.

falexwolf (approved):
Great! Just a small naming convention remark.

lamindb/models/artifact.py
Koncopd: I called it so because we already have `validate_fields` but this is different.

falexwolf: Ok, keep it this way!

Koncopd:
@falexwolf btw do you know what is with the docs?

PR #3817, opened 2026-07-22
✨ Introduce `lamin integrations notion sync` and add Notion to `lamindb.integrations`

Raaghav-Pillai:
Also see:

- https://github.com/laminlabs/lamin-cli/pull/276

Also see follow-up:

- https://github.com/laminlabs/lamindb/pull/3913


lamindb/integrations/_notion.py
```diff
@@ -0,0 +1,296 @@
...
+import logging
+import time
+from typing import TYPE_CHECKING, Any
+
+if TYPE_CHECKING:
+    import pandas as pd
+    import requests
+
+# Pinned to 2022-06-28. Notion's 2025-09-03 version replaces the
+# /v1/databases query endpoints with /v1/data_sources, since a database can
```
falexwolf: Why target the legacy API that will break? Can't we call the new data_sources endpoint instead?

lamindb/integrations/_notion.py
```diff
@@ -0,0 +1,296 @@
...
+_MAX_BACKOFF = 60
+_RETRY_STATUSES = frozenset({429, 502, 503, 504})
+
+_log = logging.getLogger(__name__)
+
+
+def _flatten_property(prop: dict) -> Any:
+    """Flatten a single Notion page property value to a plain Python object.
+
+    Dispatches on ``prop["type"]`` and returns scalars, lists, or ``None``
```
falexwolf: Always single-back-ticks in doc strings, please!

lamindb/integrations/_notion.py
```diff
@@ -0,0 +1,296 @@
...
+# /v1/databases endpoints fail with a validation error. Migration path:
+# GET /v1/databases/{id} -> data_sources[0].id -> POST /v1/data_sources/{id}/query.
+NOTION_API_VERSION = "2022-06-28"
+_NOTION_BASE_URL = "https://api.notion.com/v1"
+_MAX_RETRIES = 3
+_TIMEOUT = 30
+_MAX_BACKOFF = 60
+_RETRY_STATUSES = frozenset({429, 502, 503, 504})
+
+_log = logging.getLogger(__name__)
```
falexwolf: Please use the `from lamin_utils import logger`.

lamindb/integrations/__init__.py
```diff
@@ -18,10 +23,12 @@
 """
 
 from ._croissant import curate_from_croissant
+from ._notion import Notion
```
falexwolf: Can we avoid an eager import?

falexwolf: Instead of eagerly importing, rename `_notion.py` to `notion.py` and the class `Notion` to `Reader`.

falexwolf:
The question is also how we get relationships to work, e.g. between People and Organizations. This will be hard to do with just a dataframe 🤔 

But maybe it's still better to use a dataframe as an intermediate format and then map the Notion's uuids 🤔

PR #3833, opened 2026-08-01
♻️ Raising error when loading project with same names

ishitamjain:
Raise an error when trying to save a project with name that already exists, that points users to ln.Project.get(name=...)
Fixing:pfizer-collab/laminlabs#1139


falexwolf:
Can we close this and delete the branch?

PR #3844, opened 2026-08-05
:children_crossing: Make the default record writing space consistent with the UI

sunnyosun:
## Summary
- When creating a typed record without an explicit `space`, default to the type's `_aux.ss` single-space policy (`1`/`"1"` → type owning space, or an exact space uid).
- Keep existing precedence otherwise: explicit `space`/`space_id` > single-space policy > `context`/`settings.space` > `all`.
- Aligns SDK create behavior with the Hub write-space default so callers no longer need to pass the locked space manually (DB triggers still reject explicit mismatches).

## Test plan
- [x] Extend `test_single_space_enforces_type_space_or_specific_space` for auto-default on `ss=1` and exact uid
- [x] Keep coverage that wrong explicit `space=` still fails the Postgres trigger

## Materials

Resolves:

- https://github.com/pfizer-collab/laminlabs/issues/1137

### DIFF
--- a/lamindb/models/sqlrecord.py
+++ b/lamindb/models/sqlrecord.py
@@ -257,8 +257,9 @@ class SQLRecordSettings:
 
         Can only be set if the `SQLRecord` class inherits from `HasType` and `.is_type` is `True`.
 
-        This only affects updates to objects that will start to throw an error upon `INSERT` or `UPDATE`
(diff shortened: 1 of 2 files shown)


falexwolf (approved):
Looks great to me but requires careful communication and cross-linking of the pfizer-collab issue!

PR #3845, opened 2026-08-06
🐛 Fix the filtering for the existing records check

ishitamjain:
Here, in this PR we are trying to fix the criteria for checking of existing projects to avoid deduplication. 

Fixes: pfizer-collab/laminlabs#1139

### DIFF
--- a/lamindb/models/sqlrecord.py
+++ b/lamindb/models/sqlrecord.py
@@ -621,7 +621,11 @@ def suggest_records_with_similar_names(
     # name field in case the record is versioned (e.g. for Transform key)
     if isinstance(record, HasType):
         if kwargs.get("type", None) is None:
-            subset = record.__class__.filter(type__isnull=True)
+            # Do not restrict to type__isnull=True: a record can have is_type=True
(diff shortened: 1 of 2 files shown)


lamindb/models/sqlrecord.py
```diff
@@ -627,6 +627,18 @@ def suggest_records_with_similar_names(
     else:
         subset = record.__class__
     exact_match = subset.filter(**{name_field: kwargs[name_field]}).first()
+    # Fallback: if no match was found and no type= was passed, also search for
+    # is_type=True records that have a non-null parent type. Such records (e.g. a
```
falexwolf, line 631: I don't understand why `is_type=True` is missed above. Can you explain?

lamindb/models/sqlrecord.py
```diff
@@ -627,6 +627,18 @@ def suggest_records_with_similar_names(
...
+    # Project subtype) are missed by the filter(type__isnull=True) above and would
```
falexwolf, line 632: Are you 100% sure about this? I'm not seeing why it would be missed.

falexwolf:
Can we close this and delete the branch?

PR #3846, opened 2026-08-06
🚸 Enable linting scripts to check whether they produce complete data lineage

sheetalgiri:
- Added verify_lineage() function that parses a Python script with `ast` and checks:
  
  - lineage tracking calls (ln.track(...) / ln.finish(...) or imported equivalents)
  - Finds lines in the code that contain a file/folder path/URL and track them, and check if these these are eventually used in a LaminDB call to verify lineage tracking
 
- Updated skills.md so that for every python script that the agent creates/plans to run, it verifies the lineage by calling the function and doing assertions on the verification results, make sure everything passes. If there is a failure, it goes back and makes modifications to the script and runs the function again.  

- Update skills.md to create artifacts from in-memory objects when possible using from_dataframe(), from_anndata(),etc instead of saving it to disk and doing save()

Why:
- Agents can be unreliable when it comes to doing the lineage tracking properly 

Limitations:
- Only supports python scripts

### DIFF
--- a/lamindb/core/_verify_lineage.py
+++ b/lamindb/core/_verify_lineage.py
@@ -64,8 +64,8 @@ class LaminLineageChecker(ast.NodeVisitor):
         self.untracked_paths: dict[str, list[int]] = {}
         # Best-effort binding of variable names to path-like strings as code is traversed.
         self.var_map: dict[str, set[str]] = {}  # var_name -> set of path strings
-        # Variable names that currently refer to an ln.Artifact(...) instance.
-        self.artifact_vars: set[str] = set()  # vars assigned from ln.Artifact(...)
+        # Variable names that currently refer to a lamindb Artifact(...) instance.
+        self.artifact_vars: set[str] = set()  # vars assigned from lamindb.Artifact(...)
         # Session-level lifecycle flags required for a "fully tracked" script.
         self.has_track_call: bool = False
         self.has_finish_call: bool = False
@@ -100,7 +100,7 @@ class LaminLineageChecker(ast.NodeVisitor):
                 else self._get_func_name(decorator)
             )
             if self._is_lamindb_flow_name(decorator_name):
-                # `@ln.flow` provides run lifecycle management.
+                # `@lamindb.flow` provides run lifecycle management.
                 self.has_track_call = True
                 self.has_finish_call = True
         # Save function bodies for later path-flow tracing at call sites.
@@ -192,7 +192,7 @@ class LaminLineageChecker(ast.NodeVisitor):
         self.generic_visit(node)
 
     def _get_func_name(self, node: ast.AST) -> str:
-        """Recursively resolves AST attribute names (e.g. ln.Artifact.get)."""
+        """Recursively resolves AST attribute names (e.g. lamindb.Artifact.get)."""
         if isinstance(node, ast.Name):
             return node.id
         elif isinstance(node, ast.Attribute):
@@ -219,8 +219,9 @@ class LaminLineageChecker(ast.NodeVisitor):
         )
 
     def _is_lamindb_artifact_save_call(self, node: ast.Call) -> bool:
-        # Treat both `ln.Artifact(...).save()` and `artifact.save()` (where artifact
-        # was created from ln.Artifact(...)) as LaminDB lineage-tracked operations.
+        # Treat both `lamindb.Artifact(...).save()` and `artifact.save()`
+        # (where artifact was created from `lamindb.Artifact(...)`)
+        # as LaminDB lineage-tracked operations.
         if not (isinstance(node.func, ast.Attribute) and node.func.attr == "save"):
             return False
         if isinstance(node.func.value, ast.Call):
@@ -230,7 +231,7 @@ class LaminLineageChecker(ast.NodeVisitor):
         return False
 
     def _is_lamindb_call(self, node: ast.Call, func_name: str) -> bool:
-        # All ln./lamindb.* calls are considered lineage-aware.
+        # All calls rooted at `lamindb` (or any imported alias like `ln`) are lineage-aware.
         # We also include artifact save calls recognized above.
         root_name = func_name.split(".", 1)[0]
         if root_name in self.lamindb_module_aliases:
@@ -357,12 +358,14 @@ def verify_lineage(script_path: str) -> VerifyLineageResult:
     """Statically analyze one script and report missing LaminDB lineage hooks.
 
     Coverage (static AST checks):
-    - Confirms the script contains explicit session lifecycle calls:
-      `ln.track()` and `ln.finish()`.
+    - Confirms the script contains explicit session lifecycle calls from LaminDB:
+      `track()` and `finish()` (or `flow()`).
     - Extracts path-like strings from literals, variables, call arguments,
       and simple path joins (string `+` or path `/` operations).
     - Classifies path usage as tracked when it appears in LaminDB calls
-      (`ln.*`, `lamindb.*`, including `ln.Artifact(...).save()` patterns),
+      (`ln.*`, `lamindb.*`, or any alias like `import lamindb as alias`,
+      including `x=lamindb.Artifact(...)->x.save()` patterns and
+      `from lamindb import Artifact`),
       and flags paths that only appear in non-LaminDB calls.
     - Propagates path-like argument bindings into user-defined helper
       functions to catch indirect path usage.
@@ -410,7 +413,8 @@ def verify_lineage(script_path: str) -> VerifyLineageResult:
 
     missing_lineage: list[str] = []
 
-    # A script is fully tracked only if it explicitly opens and closes tracking.
+    # A script is fully tracked only if it opens and closes tracking
+    # (or uses `flow()` lifecycle management).
     if not checker.has_track_call:
         missing_lineage.append("Missing ln.track() call in script.")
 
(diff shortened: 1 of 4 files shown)


ebadukhan (approved):

lamindb/core/_verify_lineage.py
```diff
@@ -0,0 +1,379 @@
...
+
+def verify_lineage(script_path: str) -> VerifyLineageResult:
+    """Statically analyze one script and report missing LaminDB lineage hooks.
+
+    Coverage (static AST checks):
+    - Confirms the script contains explicit session lifecycle calls:
+      `ln.track()` and `ln.finish()`.
+    - Extracts path-like strings from literals, variables, call arguments,
+      and simple path joins (string `+` or path `/` operations).
+    - Classifies path usage as tracked when it appears in LaminDB calls
```
falexwolf, line 365: I see quite a lot of logic related to this. Why not just flag paths that are **not** going through lamindb? Wouldn't this enable us to eliminate a lot of the code?

lamindb/core/_verify_lineage.py
```diff
@@ -0,0 +1,379 @@
...
+      (`ln.*`, `lamindb.*`, including `ln.Artifact(...).save()` patterns),
+      and flags paths that only appear in non-LaminDB calls.
```
falexwolf, line 369: This is the important thing.

tests/no_instance/test_verify_lineage.py
```diff
@@ -0,0 +1,486 @@
...
+""".strip(),
+    )
+
+    result = verify_lineage(script_path)
+
+    assert result.is_fully_tracked is True
+    assert result.missing_lineage == ()
+
+
+def test_verify_lineage_positive_output_from_dataframe(tmp_path: Path):
```
falexwolf: I would combine this test and the previous one through parametrization. Both tests are very simple baseline tests.

tests/no_instance/test_verify_lineage.py
```diff
@@ -0,0 +1,486 @@
...
+""".strip(),
+    )
+
+    result = verify_lineage(script_path)
+
+    assert result.is_fully_tracked is True
+    assert result.missing_lineage == ()
+
+
+def test_verify_lineage_imported_symbols_match_current_behavior(tmp_path: Path):
```
falexwolf: What is the reason for testing this false positive behavior? 🤔

tests/no_instance/test_verify_lineage.py
```diff
@@ -0,0 +1,486 @@
...
+    )
+
+    result = verify_lineage(script_path)
+
+    assert result.is_fully_tracked is False
+    assert "Missing ln.track() call in script." in result.missing_lineage
+    assert "Missing ln.finish() call in script." in result.missing_lineage
+
+
+def test_verify_lineage_positive_zero_io_script(tmp_path: Path):
```
falexwolf: I would merge this into tests 1 & 2 through parametrization as the structure is the same

tests/no_instance/test_verify_lineage.py
```diff
@@ -0,0 +1,486 @@
...
+""".strip(),
+    )
+
+    result = verify_lineage(script_path)
+
+    assert result.is_fully_tracked is True
+    assert result.missing_lineage == ()
+
+
+def test_verify_lineage_negative_missing_lineage_tracking_calls(tmp_path: Path):
```
falexwolf: What about `@ln.flow()`? If I had a script that used `@ln.flow()` instead of `ln.track()` and `ln.finish()`, would that be also a false negative (meaning negative is "not tracked").

tests/no_instance/test_verify_lineage.py
```diff
@@ -0,0 +1,486 @@
...
+    )
+
+    result = verify_lineage(script_path)
+
+    assert result.is_fully_tracked is False
+    assert "Missing ln.track() call in script." in result.missing_lineage
+    assert "Missing ln.finish() call in script." in result.missing_lineage
+
+
+def test_verify_lineage_negative_external_input_even_when_script_has_lamindb_io(tmp_path: Path):
```
falexwolf: Great test! I think creating a catalog of these negative examples and then running a parametrized test against them is the most important

What matters is to detect scripts that don't produce lineage with high confidence

tests/no_instance/test_verify_lineage.py
```diff
@@ -0,0 +1,486 @@
...
+""".strip(),
+    )
+
+    result = verify_lineage(script_path)
+
+    assert result.is_fully_tracked is False
+    assert any("./local_input.csv" in item for item in result.missing_lineage)
+
+
+def test_verify_lineage_negative_external_output_write(tmp_path: Path):
```
falexwolf: Great!

tests/no_instance/test_verify_lineage.py
```diff
@@ -0,0 +1,486 @@
...
+""".strip(),
+    )
+
+    result = verify_lineage(script_path)
+
+    assert result.is_fully_tracked is False
+    assert any("./local_output.txt" in item for item in result.missing_lineage)
+
+
+def test_verify_lineage_negative_open_read_untracked(tmp_path: Path):
```
falexwolf: Great!

tests/no_instance/test_verify_lineage.py
```diff
@@ -0,0 +1,486 @@
...
+""".strip(),
+    )
+
+    result = verify_lineage(script_path)
+
+    assert result.is_fully_tracked is False
+    assert any("./local_input.txt" in item for item in result.missing_lineage)
+
+
+def test_verify_lineage_positive_local_write_then_lamindb_save(tmp_path: Path):
```
falexwolf: That's cool that this is properly classified. Impressive!

falexwolf:
Cool! 😄 

(1) What about scripts that track lineage via `@ln.flow()`?
(2) I think I understand now why we care about classifying positive usage of `ln.Artifact()` etc.: because one needs to detect whether an invalid call writes a temp file that's later ingested, right?
(3) I would use parametrized tests a bit more heavily since this hinges so much on being a catalog of scripts with their error types
(4) What about docs and public API? Is this supposed to be a public API or do we want to first use this internally only? If it's a public API we should add a screenshot of the docs and a nice example. ☺️

tests/no_instance/test_verify_lineage.py
sheetalgiri: this is a mistake, it should be tracked

lamindb/core/_verify_lineage.py
sheetalgiri: This was my initial idea but in practice I saw agents using non-lamindb write functions in combination with lamindb save. Its also inevitable in general because we would have to do this for file formats other than those where we have `ln.Artifact.from_*` methods like `from_dataframe` or `from_anndata`

sheetalgiri:
@falexwolf thanks for the thorough feedback ! 

> (1) What about scripts that track lineage via `@ln.flow()`? 

This is something I missed, will add it

> (2) I think I understand now why we care about classifying positive usage of `ln.Artifact()` etc.: because one needs to detect whether an invalid call writes a temp file that's later ingested, right?

Exactly, when the path is written to a temp file it adds it to the 'untracked' list. Once it sees that the same path is saved on Lamindb, it adds it to the 'tracked' list and then its considered resolved.

> (3) I would use parametrized tests a bit more heavily since this hinges so much on being a catalog of scripts with their error types 

Makes sense 👍 

> (4) What about docs and public API? Is this supposed to be a public API or do we want to first use this internally only? If it's a public API we should add a screenshot of the docs and a nice example. ☺️

I would say we use it internally first and get more feedback/experience using it with agents, what do you think?

falexwolf:
> I would say we use it internally first and get more feedback/experience using it with agents, what do you think?

Agreed!

sheetalgiri:
@falexwolf I made the changes, could you please have another look?

sheetalgiri:
While testing this, I noticed that our skills are kind of bad at using `ln.flow()`. There is no mention of using it in the SKILLS.md and then it also uses it incorrectly. It generated:

``` python
import lamindb as ln

# 1. Start global script tracking
ln.track()

# 2. Use @ln.flow() for functions nested inside an ln.track() session
@ln.flow()
def process_sample(sample_id: str):
    df = ln.examples.datasets.mini_immuno.get_dataset1()
    return ln.Artifact.from_dataframe(df, key=f"processed_{sample_id}.parquet").save()

# 3. Call the step function
artifact = process_sample(sample_id="batch_01")

# 4. Finish the global tracking session
ln.finish()
```
which caused the error
``` python
RuntimeError: Please use @ln.step() or clear the global run context before using @ln.flow(): no `ln.track()` or `@ln.flow(global_run='clear')`
```

So, ln.flow() should never be used inside ln.track(). Is that correct @falexwolf ? 

I could add the skills for this but I would need to understand better when exactly we want agents to use ln.flow(), is it something meant to be used for notebooks and not for normal scripts?

FYI @Ebad371

falexwolf:
> So, ln.flow() should never be used inside ln.track(). Is that correct @falexwolf ?

Yes, that's correct.

While `ln.flow()` is ultimately the more powerful and better pattern, we can also be pragmatic for now and just focus on `ln.track()` and `ln.finish()`. We can push out dealing with `@ln.flow()` in the upcoming weeks, when the need arises. I use it in most of my work, so I'll run into this sooner or later.

Issue #3847, opened 2026-08-07
2.9.1 release not compatible with pandas > 3


seohyonkim:
### Add a description

Hello! I noticed that the version 2.9.1 doesn't work with pandas > 3.
```
conda run -n lamin2.9-pd3-test pip install "lamindb==2.9.1" "pandas==3.0.5"
Defaulting to user installation because normal site-packages is not writeable
Collecting lamindb==2.9.1
  Using cached lamindb-2.9.1-py3-none-any.whl.metadata (19 kB)
Collecting pandas==3.0.5
  Downloading pandas-3.0.5-cp313-cp313-manylinux_2_24_aarch64.manylinux_2_28_aarch64.whl.metadata (79 kB)
Collecting lamindb-core==2.9.1 (from lamindb-core[full]==2.9.1->lamindb==2.9.1)
  Using cached lamindb_core-2.9.1-py3-none-any.whl.metadata (21 kB)
Requirement already satisfied: numpy>=1.26.0 in /usr/lib64/python3.13/site-packages (from pandas==3.0.5) (2.2.6)
Requirement already satisfied: python-dateutil>=2.8.2 in /usr/lib/python3.13/site-packages (from pandas==3.0.5) (2.8.2)
Collecting lamin_utils==0.16.4 (from lamindb-core==2.9.1->lamindb-core[full]==2.9.1->lamindb==2.9.1)
  Using cached lamin_utils-0.16.4-py2.py3-none-any.whl.metadata (962 bytes)
Collecting lamin_cli==1.19.2 (from lamindb-core==2.9.1->lamindb-core[full]==2.9.1->lamindb==2.9.1)
  Using cached lamin_cli-1.19.2-py2.py3-none-any.whl.metadata (803 bytes)
Collecting lamindb_setup==1.25.6 (from lamindb_setup[aws]==1.25.6->lamindb-core==2.9.1->lamindb-core[full]==2.9.1->lamindb==2.9.1)
  Using cached lamindb_setup-1.25.6-py3-none-any.whl.metadata (1.9 kB)
Collecting psycopg2-binary (from lamindb-core==2.9.1->lamindb-core[full]==2.9.1->lamindb==2.9.1)
  Using cached psycopg2_binary-2.9.12-cp313-cp313-manylinux_2_27_aarch64.manylinux_2_28_aarch64.whl.metadata (4.9 kB)
[...]

PR #3848, opened 2026-08-07
⬆️ Add compatibility with `pandas` >= 3

Koncopd:
Address https://github.com/laminlabs/lamindb/issues/3847

### DIFF
--- a/lamindb/base/dtypes.py
+++ b/lamindb/base/dtypes.py
@@ -17,6 +17,19 @@ def is_list_of_type(value: Any, expected_type: Any) -> bool:
     return False
 
 
+def check_str_index(series) -> bool:
+    """Validate a DataFrame index as pandera ``str``, allowing empty indexes.
+
+    Empty exports keep a default ``RangeIndex`` (``int64``); non-empty indexes must
+    match pandera's ``str`` dtype (e.g. ``string[pyarrow]`` on pandas 3).
+    """
+    from pandera.engines import pandas_engine
+
+    if len(series) == 0:
+        return True
+    return bool(pandas_engine.Engine.dtype("str").check(series.dtype))
+
+
 def check_dtype(expected_type: Any, nullable: bool) -> Callable:
     """Creates a check function for Pandera that validates a column's dtype.
 
--- a/lamindb/base/dtypes.py
+++ b/lamindb/base/dtypes.py
@@ -44,7 +44,14 @@ def check_dtype(expected_type: Any, nullable: bool) -> Callable:
             return True
         elif expected_type == "num" and pd.api.types.is_numeric_dtype(series.dtype):
             return True
-        elif expected_type == "str" and pd.api.types.is_string_dtype(series.dtype):
+        elif expected_type == "str" and (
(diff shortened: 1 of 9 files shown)


Koncopd:
I see, we need to support `pandas` nullable dtypes like `anndata` does.

Koncopd:
Also https://github.com/scverse/anndata/issues/2595

lamindb/base/dtypes.py
```diff
@@ -17,6 +17,48 @@ def is_list_of_type(value: Any, expected_type: Any) -> bool:
     return False
 
 
+def check_pandera_str(series) -> bool:
```
Koncopd, line 20: @falexwolf this unifies validation for `str` between `pandas` 2 and 3, taking 2 as the correct behavior.

lamindb/curators/core.py
```diff
@@ -631,6 +631,15 @@ def convert_dict_to_dataframe_for_validation(d: dict, schema: Schema) -> pd.Data
                             f"{value.__class__.__name__} {getattr(value, getattr(value, 'name_field', 'name'), value.uid)} is not saved."
                         )
                     df[feature.name] = pd.Categorical(df[feature.name])
+        # pandas 3 defaults tz-aware datetimes to [us]; lamin/pandera expect [ns]
```
Koncopd, line 634: @falexwolf also see this.

lamindb/curators/core.py
```diff
@@ -787,11 +801,23 @@ def __init__(
...
-                index = pandera.Index(
-                    schema.index._dtype_str
-                    if not schema.index._dtype_str.startswith("cat")
-                    else str
+                index_dtype = (
+                    "str"
+                    if schema.index._dtype_str.startswith("cat")
+                    else schema.index._dtype_str
                 )
+                if index_dtype == "str":
```
Koncopd, line 809: and this for indices.

lamindb/base/dtypes.py
```diff
@@ -17,6 +17,48 @@ def is_list_of_type(value: Any, expected_type: Any) -> bool:
...
+    """Validate a series/index as lamin ``str``, matching pandas 2 pandera results.
```
falexwolf: Please never use double backticks! We use single-backticks throughout all our repos.

lamindb/base/dtypes.py
falexwolf: Why not take pandas 3 as the correct behavior? Wouldn't that be more future proof? I don't think the nuances will matter. But I'm also fine merging this way and changing this in lamindb 3. We'll just forget about it if we don't make a bullet for a "lamindb 3" issue (which is easy to do).

lamindb/curators/core.py
falexwolf: Ok!

lamindb/curators/core.py
falexwolf: Ok!

falexwolf (approved):

lamindb/base/dtypes.py
Koncopd: fixed

PR #3850, opened 2026-08-10
🚸 Broaden name deduplication in `HasType` to search the hierarchy if no `type` is passed

ishitamjain:
When creating a `HasType` object:

- If no `type` is passed, search across the full type hierarchy.
- If `type` is passed, restrict to the corresponding hierarchy (including `type=None` meaning root).

Also adds a warning when `type=None` is explicitly passed for non-type objects, since root-level creation should be intentional and rare.

```python
>>> import lamindb as ln
>>> ln.ULabel(name="my-test-label")
! tip: pass `type` to map ulabel into a type hierarchy
```

Addresses:

- https://github.com/pfizer-collab/laminlabs/issues/1139

### DIFF
--- a/lamindb/base/types.py
+++ b/lamindb/base/types.py
@@ -156,7 +156,7 @@ DtypeObject = SimpleDvalue  # backward compat
 class Unset:
     """Sentinel type to distinguish 'not passed' from explicit ``None``.
 
-    The singleton :data:`~lamindb.base.types.UNSET` is the only instance.
+    The singleton ``UNSET`` is the only instance.
     Use ``is`` checks to compare against it::
 
         if value is UNSET:
--- a/lamindb/base/types.py
+++ b/lamindb/base/types.py
@@ -26,8 +26,8 @@ Basic types
 .. autoclass:: ListLike
 .. autoclass:: FieldAttr
 
-Sentinel types
---------------
+Auxiliary types
+---------------
 
 .. autoclass:: Unset
 """
@@ -154,10 +154,10 @@ Dtype = DtypeStr  # backward compat
 DtypeObject = SimpleDvalue  # backward compat
 
 class Unset:
-    """Sentinel type to distinguish 'not passed' from explicit ``None``.
+    """Sentinel type to distinguish 'not passed' from explicit `None`.
 
-    The singleton ``UNSET`` is the only instance.
-    Use ``is`` checks to compare against it::
+    The `UNSET` constant is the only instance.
+    Use `is` checks to compare against it::
 
         if value is UNSET:
             ...  # user did not pass the argument
@@ -167,21 +167,13 @@ class Unset:
             ...  # user passed a real value
     """
 
-    _instance: Unset | None = None
-
-    def __new__(cls) -> Unset:
-        if cls._instance is None:
-            cls._instance = super().__new__(cls)
-        return cls._instance
-
-    def __repr__(self) -> str:
-        return "UNSET"
+    pass
 
 
 UNSET: Unset = Unset()
-"""Sentinel value to distinguish 'argument not passed' from explicit ``None``.
+"""Sentinel value to distinguish 'argument not passed' from explicit `None`.
 
-See :class:`~lamindb.base.types.Unset`.
+See `Unset`.
 """
 
 RegistryId = Literal[
--- a/lamindb/base/types.py
+++ b/lamindb/base/types.py
@@ -25,6 +25,11 @@ Basic types
 .. autoclass:: StrField
 .. autoclass:: ListLike
 .. autoclass:: FieldAttr
+
+Sentinel types
+--------------
+
+.. autoclass:: Unset
 """
 
 from __future__ import annotations
@@ -147,6 +152,37 @@ SimpleDtypeStr = Literal[
 DtypeStr = SimpleDtypeStr  # backward compat
 Dtype = DtypeStr  # backward compat
 DtypeObject = SimpleDvalue  # backward compat
+
+class Unset:
+    """Sentinel type to distinguish 'not passed' from explicit ``None``.
+
+    The singleton :data:`~lamindb.base.types.UNSET` is the only instance.
+    Use ``is`` checks to compare against it::
+
+        if value is UNSET:
+            ...  # user did not pass the argument
+        elif value is None:
+            ...  # user explicitly passed None
+        else:
+            ...  # user passed a real value
+    """
+
+    _instance: Unset | None = None
+
+    def __new__(cls) -> Unset:
+        if cls._instance is None:
+            cls._instance = super().__new__(cls)
+        return cls._instance
+
+    def __repr__(self) -> str:
+        return "UNSET"
+
+
+UNSET: Unset = Unset()
+"""Sentinel value to distinguish 'argument not passed' from explicit ``None``.
+
+See :class:`~lamindb.base.types.Unset`.
+"""
 
 RegistryId = Literal[
     "__lamindb_artifact__",
--- a/lamindb/models/feature.py
+++ b/lamindb/models/feature.py
@@ -43,6 +43,8 @@ from .run import (
     TracksRun,
     TracksUpdates,
 )
+from lamindb.base.types import Unset
+
 from .sqlrecord import (
     UNSET,
     BaseSQLRecord,
--- a/lamindb/models/feature.py
+++ b/lamindb/models/feature.py
@@ -51,6 +51,7 @@ from .sqlrecord import (
     Space,
     SQLRecord,
     UNSET,
+    Unset,
     _get_record_kwargs,
     pop_space_branch_kwargs,
 )
@@ -671,7 +672,7 @@ def process_init_feature_param(args, kwargs):
     name: str = kwargs.pop("name", None)
     dtype: SimpleDtype | SimpleDtypeStr | str | None = kwargs.pop("dtype", None)
     is_type: bool = kwargs.pop("is_type", False)
-    type_: Feature | str | None = kwargs.pop("type", UNSET)
+    type_: Feature | str | None | Unset = kwargs.pop("type", UNSET)
     description: str | None = kwargs.pop("description", None)
     space_branch_kwargs = pop_space_branch_kwargs(kwargs)
     _skip_validation = kwargs.pop("_skip_validation", False)
@@ -1221,7 +1222,7 @@ class Feature(SQLRecord, HasType, CanCurate, HasSynonyms, TracksRun, TracksUpdat
         | Registry
         | list[Registry]
         | FieldAttr,
-        type: Feature | None = UNSET,
+        type: Feature | None | Unset = UNSET,
         is_type: bool = False,
         unit: str | None = None,
         description: str | None = None,
--- a/lamindb/models/feature.py
+++ b/lamindb/models/feature.py
@@ -1221,7 +1221,7 @@ class Feature(SQLRecord, HasType, CanCurate, HasSynonyms, TracksRun, TracksUpdat
         | Registry
         | list[Registry]
         | FieldAttr,
-        type: Feature | None = None,
+        type: Feature | None = UNSET,
         is_type: bool = False,
         unit: str | None = None,
         description: str | None = None,
--- a/lamindb/models/project.py
+++ b/lamindb/models/project.py
@@ -252,8 +252,7 @@ class Reference(
         if len(args) == len(self._meta.concrete_fields):
             super().__init__(*args, **kwargs)
             return None
-        if len(args) > 0:
-            raise ValueError("Only keyword args allowed")
+        assert not args, "Only keyword args allowed."
         name: str = kwargs.pop("name", None)
         type: Reference | None | Unset = kwargs.pop("type", UNSET)
         is_type: bool = kwargs.pop("is_type", False)
@@ -518,8 +517,7 @@ class Project(
         if len(args) == len(self._meta.concrete_fields):
             super().__init__(*args, **kwargs)
             return None
-        if len(args) > 0:
-            raise ValueError("Only keyword args allowed")
+        assert not args, "Only keyword args allowed."
         name: str = kwargs.pop("name", None)
         type: Project | None | Unset = kwargs.pop("type", UNSET)
         is_type: bool = kwargs.pop("is_type", False)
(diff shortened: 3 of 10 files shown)


falexwolf:
Great!

> Also adds a warning when type=None is explicitly passed for non-type objects, since root-level creation should be intentional and rare.

Can you condition this on `is_type=False`? If `is_type=True`, then this is likely intentional.

ishitajain9717:
Hi, 
It is already conditioned on `is_type=False`.

lamindb/models/sqlrecord.py
```diff
@@ -610,7 +610,7 @@ def get_branch_id_for_create(
 
 
 def suggest_records_with_similar_names(
-    record: SQLRecord, name_field: str, kwargs
+    record: SQLRecord, name_field: str, kwargs, type_val=UNSET
```
falexwolf: Why do we need to introduce `type_val`? The `kwargs` have `type` in them and you can detect whether it matches `UNSET`. Why is that not possible?

lamindb/models/sqlrecord.py
```diff
@@ -624,10 +624,17 @@ def suggest_records_with_similar_names(
     # the below needs to be .first() because there might be multiple records with the same
     # name field in case the record is versioned (e.g. for Transform key)
     if isinstance(record, HasType):
-        if kwargs.get("type", None) is None:
```
falexwolf, line 650: I would have expected something like this:

```python
type = kwargs.get("type", UNSET)
if type is UNSET:
     subset = record.__class__.filter()
elif type is None:
    subset = record.__class__.filter(type__isnull=True)
else:
    subset = record.__class__.filter(type=kwargs["type"])
```

falexwolf: There is a question whether we should use `kwargs["type"]` instead of `kwargs.get()` because we might now actually have a contract that guarantees the presence of `type` since it would never be popped? 🤔

lamindb/models/sqlrecord.py
```diff
@@ -1214,13 +1218,21 @@ def resolve_fk_or_id(field_name: str) -> bool:
...
+                if isinstance(self, HasType) and kwargs.get("type", UNSET) is UNSET:
+                    kwargs.pop("type", None)
                 super().__init__(**kwargs)
             else:
                 from ..core._settings import settings
                 from .can_curate import CanCurate
                 from .collection import Collection
                 from .transform import Transform
 
+                # ensure "type" is always present in kwargs for HasType models so that
```
falexwolf: Do you understand why this is necessary? My assumption is that every HasType model populates `type=UNSET` or whatever the user passes. How can it be that there are cases where you need this defensive line?

If this is AI generated: Can you remove it and see if the tests pass?

lamindb/models/sqlrecord.py
```diff
@@ -1214,13 +1218,21 @@ def resolve_fk_or_id(field_name: str) -> bool:
                     # the current one), so the record is created on that branch.
                     kwargs["created_on"] = kwargs["branch"]
             if skip_validation:
+                # strip UNSET just before Django sees kwargs — FK descriptors reject non-model values
+                if isinstance(self, HasType) and kwargs.get("type", UNSET) is UNSET:
```
falexwolf: The logic here is incorrect: if `"type" is not in kwargs` the `.get()` statement returns `UNSET` and one enters `kwargs.pop()` even though there is no key `"type"`!

IMO the same rationale that would make https://github.com/laminlabs/lamindb/pull/3850/changes#r3782589607 easy would make this easy: replacing `kwargs.get()` with `kwargs["type"] is UNSET`.

tests/pydata/test_record_basics.py
```diff
@@ -690,14 +690,17 @@ def test_feature_manager_raise_not_validated_values():
 def test_name_lookup():
     my_type = ln.Record(name="MyType", is_type=True).save()
     label1 = ln.Record(name="label 1", type=my_type).save()
+    # same type → returns existing typed record
```
falexwolf, line 693: Great idea to add the comments here! 🙏

tests/pydata/test_record_basics.py
```diff
@@ -690,14 +690,17 @@ def test_feature_manager_raise_not_validated_values():
...
     label2 = ln.Record(name="label 1", type=my_type)
     assert label2 == label1
+    # no type passed, only typed record exists → fallback returns the typed one
```
falexwolf, line 696: Great! This is the behavior change now.

tests/pydata/test_record_basics.py
```diff
@@ -690,14 +690,17 @@ def test_feature_manager_raise_not_validated_values():
 def test_name_lookup():
```
falexwolf, line 690: Aren't we now missing a test here where the user explicitly passes `type=None`?

falexwolf:
Thanks, @ishitajain9717! This is now much better designed than yesterday! 🙏 

Can you mark the comments that you resolved as resolved?

Now only very small issues are left, for which I made new comments. ☺️

lamindb/models/sqlrecord.py
ishitajain9717: I think the injection is needed for `Project` and `Reference`. Unlike `Record`, `ULabel`, `Feature`, and `Schema`, which  explicitly pop and re-inject type (defaulting to UNSET) in their own constructors, `Project` and `Reference` have __init__ methods that pass `args`, `kwargs` straight through to super(). So if a user calls Project(name="foo") without type, the key is simply absent from kwargs when BaseSQLRecord receives it.
Please suggest if it makes sense

lamindb/models/sqlrecord.py
ishitajain9717: I agree, this is logically imprecise. I am correcting it in upcoming changes

lamindb/models/sqlrecord.py
```diff
@@ -626,15 +626,15 @@ def suggest_records_with_similar_names(
     if isinstance(record, HasType):
         # "type" is always present in kwargs at this point (Solution A contract)
         type = kwargs["type"]
-        if type is UNSET:
+        if type is None:
+            # explicit type=None → user wants a new root-level record, skip dedup
```
falexwolf: This is wrong, isn't it? If we pass `type=None` we need to deduplicate on the root level.

```
# explicit type=None → search root-level (type IS NULL)
            subset = record.__class__.filter(type__isnull=True)
```

tests/pydata/test_record_basics.py
```diff
@@ -696,10 +696,18 @@ def test_name_lookup():
     # no type passed, only typed record exists → fallback returns the typed one
     label2 = ln.Record(name="label 1")
     assert label2 == label1
-    # root-level record exists → no-type search finds root first before typed
+    # explicit type=None → always skip dedup, always a new root-level record
+    label_new = ln.Record(name="label 1", type=None)
+    assert label_new != label1
+    assert label_new._state.adding  # not yet saved, truly a new record
+    # explicit type=None, even if a root-level record exists → still skips dedup
```
falexwolf: Why?

lamindb/models/sqlrecord.py
ishitajain9717: I realised it immediately, and I was just correcting it 😅. I was actually reviewing changes

lamindb/models/sqlrecord.py
falexwolf: It's great that you found this!

> Project and Reference have init methods that pass args, kwargs straight through to super().

Hmmm. If that's what's happening then we should see `type: UNSET` in the `kwargs`, dict, right? I think your reasoning might be incorrect here.

ishitajain9717: You're right that we should always see type: UNSET in kwargs — that's the contract. But the problem is that UNSET is an internal sentinel. So "type" only ends up in kwargs if the model's __init__ explicitly injects it. Do you agree?

falexwolf: Maybe I missed something in the previous PR: I understood that we changed `type = None` to `type = UNSET` in the previous PR. This isn't an internal but a user-facing change. If it wasn't user-facing, how would a user learn about the difference of passing `type = None` versus not passing `type`?

I'm convinced that every `HasType` registry needs to expose `type = UNSET` to the user also in the docs; and of course, if the user doesn't pass `type` that needs to get passed into the downstream calling cascade. That should IMO be automatic since `UNSET` is the default value of the constructor.

ishitajain9717: I agree with that, this should be a user facing change, and should also be mentioned in the docs specifically. I would make the change now

lamindb/models/feature.py
```diff
@@ -1221,7 +1221,7 @@ def __init__(
         | Registry
         | list[Registry]
         | FieldAttr,
-        type: Feature | None = None,
+        type: Feature | None = UNSET,
```
falexwolf: Can you please also update the type annotation? The type annotation currently says `Feature | None`, but it has to account for the `UNSET` type now, which is neither, isn't it?

lamindb/models/project.py
```diff
@@ -237,7 +248,44 @@ def __init__(
     ): ...
 
     def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
+        if len(args) == len(self._meta.concrete_fields):
+            super().__init__(*args, **kwargs)
+            return None
+        if len(args) > 0:
+            raise ValueError("Only one non-keyword arg allowed")
```
falexwolf: Which non-keyword argument is allowed, here? None, right? Only keyword arguments are allowed.

lamindb/models/project.py
```diff
@@ -466,7 +514,38 @@ def __init__(
     ): ...
 
     def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
+        if len(args) == len(self._meta.concrete_fields):
+            super().__init__(*args, **kwargs)
+            return None
+        if len(args) > 0:
```
falexwolf: Same comment as above.

falexwolf:
Great! This starts to be super nice!

A comment about your current design of the `Unset` class.

```python
class Unset:
    """Sentinel type to distinguish 'not passed' from explicit ``None``.

    The singleton ``UNSET`` is the only instance.
    Use ``is`` checks to compare against it::

        if value is UNSET:
            ...  # user did not pass the argument
        elif value is None:
            ...  # user explicitly passed None
        else:
            ...  # user passed a real value
    """

    _instance: Unset | None = None

    def __new__(cls) -> Unset:
        if cls._instance is None:
            cls._instance = super().__new__(cls)
        return cls._instance

    def __repr__(self) -> str:
        return "UNSET"
```

Are you sure that this needs to be so complicated? 

I'm pretty sure that the following would also work since before you simply used `object()`:

```python
class Unset:
    """Sentinel type to distinguish 'not passed' from explicit `None`.

    The `UNSET` constant is the only instance.
    Use `is` checks to compare against it::

        if value is UNSET:
            ...  # user did not pass the argument
        elif value is None:
            ...  # user explicitly passed None
        else:
            ...  # user passed a real value
    """
    pass
```

 Please also note that we use single backticks throughout our docs, not double back ticks. ☺️

lamindb/base/types.py
```diff
@@ -148,6 +153,37 @@
...
+    The singleton ``UNSET`` is the only instance.
+    Use ``is`` checks to compare against it::
+
+        if value is UNSET:
+            ...  # user did not pass the argument
+        elif value is None:
+            ...  # user explicitly passed None
+        else:
+            ...  # user passed a real value
+    """
```
falexwolf, line 168: See my comment below: https://github.com/laminlabs/lamindb/pull/3850#issuecomment-5325170137

lamindb/base/types.py
```diff
@@ -148,6 +153,37 @@
...
+        if cls._instance is None:
+            cls._instance = super().__new__(cls)
+        return cls._instance
+
+    def __repr__(self) -> str:
+        return "UNSET"
+
+
+UNSET: Unset = Unset()
+"""Sentinel value to distinguish 'argument not passed' from explicit ``None``.
```
falexwolf: Please use single back ticks ☺️

lamindb/models/sqlrecord.py
```diff
@@ -108,9 +108,8 @@
 
 T = TypeVar("T", bound="SQLRecord")
 
-# Sentinel to distinguish "user didn't pass type=" from "user explicitly passed type=None".
-# Uses object() so identity checks (is) never call __eq__ on model instances.
-UNSET = object()
+from lamindb.base.types import UNSET, Unset  # noqa: E402
```
falexwolf: You don't need to import the class here, I believe, only the value. ☺️

lamindb/models/sqlrecord.py
```diff
@@ -624,10 +623,18 @@ def suggest_records_with_similar_names(
     # the below needs to be .first() because there might be multiple records with the same
     # name field in case the record is versioned (e.g. for Transform key)
     if isinstance(record, HasType):
-        if kwargs.get("type", None) is None:
+        # "type" is always present in kwargs at this point (Solution A contract)
```
falexwolf: This part of the comment "(Solution A contract)"  is probably left over from the early days of this PR -- the code design today provides the contract; you could call this out but in this case I think it would also be apparent to most developers that `type` should be contained in `kwargs`

I'd remove the whole comment because already L627 implies that `"type"` is always present as it'd error otherwise

lamindb/base/types.py
```diff
@@ -25,6 +25,11 @@
 .. autoclass:: StrField
 .. autoclass:: ListLike
 .. autoclass:: FieldAttr
+
+Sentinel types
```
falexwolf: Can you call this Auxiliary types?

falexwolf:
One more thing:

```python
        if len(args) > 0:
            raise ValueError("Only keyword args allowed")
```

This introduces an uncovered line unless we write a new test. I suggest you turn this into an assert

```python
    assert not args, "Only keyword args allowed."
```

This way we don't deteriorate coverage and still don't need to write a test. ☺️

falexwolf:
Finally, re-reading the PR description, I believe this might have gone missing in all the iterations on this PR:

> Also adds a warning when type=None is explicitly passed for non-type objects, since root-level creation should be intentional and rare.

The warning should be printed if `type is UNSET`. I think we could add it here in L628:

```python
if type is UNSET:
    # user passed nothing → search all type contexts
    # catches typed records with same name, fixes silent dup bug
    logger.warning("tip: pass `type` to map {record.__class__.__name__.lower()} into a type hierarchy")
    subset = record.__class__.filter()
```

falexwolf:
Great work, @ishitajain9717! 🙏

PR #3851, opened 2026-08-11
:recycle: Refactor handling of unset types in `HasType`

ishitamjain:
In this PR, we make a minor change to make "Auto" as default rather than "None", so as to deal with duplication issues separately

### DIFF
--- a/lamindb/models/sqlrecord.py
+++ b/lamindb/models/sqlrecord.py
@@ -109,8 +109,8 @@ if TYPE_CHECKING:
 T = TypeVar("T", bound="SQLRecord")
 
 # Sentinel to distinguish "user didn't pass type=" from "user explicitly passed type=None".
-# Using a unique string so it is serialisation-safe and easy to read in tracebacks.
-_UNSET = "Auto"
+# Uses object() so identity checks (is) never call __eq__ on model instances.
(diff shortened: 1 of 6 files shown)


testbranch_id.txt
```diff
@@ -0,0 +1 @@
+branch_id
```
falexwolf: Please remove!

lamindb/models/sqlrecord.py
```diff
@@ -107,6 +107,10 @@
 
 
 T = TypeVar("T", bound="SQLRecord")
+
+# Sentinel to distinguish "user didn't pass type=" from "user explicitly passed type=None".
+# Uses object() so identity checks (is) never call __eq__ on model instances.
+_UNSET = object()
```
falexwolf: Can we not prefix this with an underscore for consistency with all these other variables? ☺️

falexwolf:
Awesome, @ishitajain9717! This is a great solution.

I only have very minor comments. Feel free to merge once they're addressed.

falexwolf (approved):

testbranch_id.txt
ishitajain9717: Done

PR #3852, opened 2026-08-12
:children_crossing: Consistent API for saving to non-default databases via the `using` argument in `ln.save()` & `BaseSQLRecord.save()`

falexwolf:
**With:** @Raaghav-Pillai

Sometimes one wants to save objects to non-default databases. The `using` argument is now officially supported in `BaseSQLRecord.save()` and also introduced in `ln.save()`.

Needs:

- https://github.com/laminlabs/lamindb-setup/pull/1382

### DIFF
--- a/lamindb/models/artifact.py
+++ b/lamindb/models/artifact.py
@@ -2110,14 +2110,12 @@ class Artifact(SQLRecord, IsVersioned, TracksRun, TracksUpdates):
             artifact.path
             #> PosixPath('/home/runner/work/lamindb/lamindb/docs/guide/mydata/myfile.csv')
         """
-        filepath, _ = _s().filepath_from_artifact(self, using_key=settings._using_key)
+        filepath, _ = _s().filepath_from_artifact(self)
         return filepath
 
     @property
     def _cache_path(self) -> UPath:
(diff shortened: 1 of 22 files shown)


Raaghav-Pillai:
Need to add a similar fix that was for write for read in which when it is doing` .using(LOAD).to_dataframe(include="features")` it should use connected database as default

Raaghav-Pillai:
`bulk_set_features_in_records` at line 2326 needs to handle multi-valued features

Koncopd:
i will need to review this also.

I actually really hoped we would remove `using_key` intead of extending it.

falexwolf:
> I actually really hoped we would remove using_key intead of extending it.

This is what's happening in this PR.

lamindb/models/artifact.py
```diff
@@ -255,7 +255,7 @@ def process_pathlike(
         # already-registered storage locations
         result = None
         # within the hub, we don't want to perform check_path_in_existing_storage
-        if using_key is None:
+        if using is None:
```
Koncopd, line 258: why is this kept?

Koncopd (approved):

PR #3853, opened 2026-08-12
✨ Add a function to create a download link for artifacts that can be used by agents when referencing the artifact in markdown/html

sheetalgiri:
Why
- Agents trying to create markdown with images that are artifacts but this doesn't work well because it refers to local paths of the image https://lamin.ai/laminlabs/lamindata/artifact/6bWK22v0wFFOc5gF0000

Solution
- Add functionality to create such a link using `artifact.create_download_link()`. Closest thing that existed before was `ln.Artifact.path` but this does not create a proper link. The implementation closely follows the implementation in LaminHub for the same :https://github.com/laminlabs/laminhub/blob/6f213282d58707ecbd86ba896b4ebab78851c03b/frontend/src/lib/utils/utils.ts#L44-L69, except that the /storage/s3 endpoint can only be used if the instance is on LaminHub, so this is being checked for.
- Also instructed the skills.md to use this function 
Working example: https://lamin.ai/laminlabs/lamindata/artifact/pwSp3LRvGTVBoUzp0000?branch=verify_lineage_test

### DIFF
--- a/lamindb/core/storage/paths.py
+++ b/lamindb/core/storage/paths.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import shutil
 from typing import TYPE_CHECKING
+from urllib.parse import urlencode
 
 import fsspec
 from lamindb_setup.core import StorageSettings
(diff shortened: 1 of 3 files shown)


falexwolf:
Nice idea! But rather than adding this here, I think we should extend `UPath.to_url()`: https://github.com/laminlabs/lamindb-setup/blob/307c46e1e64b73def27c9fd3d137a850ab1eaa87/lamindb_setup/core/upath.py#L851-L871

`to_url()` currently only works for public S3 buckets, since it misses the full authorization layer. It'd be nice if this was overloaded with a Lamin URL from our own storage service if the storage location is not public (the path object is authed).

sheetalgiri:
@falexwolf created a PR for `UPath.to_url()` here https://github.com/laminlabs/lamindb-setup/pull/1379, please have a look there. I'm closing this one.

PR #3854, opened 2026-08-17
🚸 Allow `Artifact` registration also if cloud stats cannot be read

Koncopd:
`Artifact` creation from a cloud path no longer fails when storage is reachable but not fully readable.

If `lamindb` cannot fetch size or hash (network error, access denied, VPC Service Controls, and similar), it logs a warning and still creates the `artifact`, leaving those fields empty.

The same applies to the existence check: `PermissionError` (S3) and `Forbidden` (GCS) warn and continue. A path that is actually missing still raises `FileNotFoundError`.

### DIFF
--- a/lamindb/models/artifact.py
+++ b/lamindb/models/artifact.py
@@ -214,11 +214,6 @@ OUTDATED_ARTIFACT_FILES_OVERWRITTEN_MSG = (
 )
 
 
-def _is_cloud_access_denied(error: OSError) -> bool:
-    # s3fs raises PermissionError; gcsfs raises OSError("Forbidden: ...")
-    return isinstance(error, PermissionError) or "forbidden" in str(error).lower()
-
-
 def process_pathlike(
     filepath: UPath,
     storage: Storage,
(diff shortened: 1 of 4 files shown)


lamindb/models/artifact.py
```diff
@@ -223,10 +228,15 @@ def process_pathlike(
     """Determines the appropriate storage for a given path and whether to use an existing storage key."""
     if not skip_existence_check:
         try:  # check if file exists
-            if not filepath.exists():
+            exists = filepath.exists()
+        except OSError as e:
+            # PermissionError is an OSError; s3fs raises it, gcsfs uses OSError("Forbidden: ...")
+            if not _is_cloud_access_denied(e):
```
falexwolf: I think this would be easier to read if there was no helper function. Since it's only used in this single place and just a single line, I'd just move L219 here?

Koncopd: agreed

Koncopd: removed that

falexwolf (approved):
Adding a GCP example would probably be good to check whether the "forbidden" thing ever changes.

Koncopd:
it is hard to trigger this naturally and patching is useless for this. I have checked gcsfs code and actual errors for this on sentry.

falexwolf:
Why not trigger it by using a private path on GCP? This is the test case that we're interested in, isn't it?

PR #3855, opened 2026-08-17
:children_crossing: If a path inside a folder-like artifact is passed to `Artifact()`, return the existing folder-like artifact

Koncopd:
If a path in `.lamindb/` is provided to `Artifact()` and is already also a child of a folder registered as an `Artifact`, then that folder-like `Artifact` is returned, similar to how an artifact is returned upon matching the full path or a `hash`.

Before this change, a file inside a folder-like artifact under `.lamindb/` did not resolve to that folder. Now `Artifact()` retries the `.lamindb/<child>` parent:

```python
folder = ln.Artifact("./my_folder", key="my_folder").save()
# stored at s3://bucket/.lamindb/{uid}/...

nested = f"{folder.path}/subdir/file.txt"
artifact = ln.Artifact(nested)
assert artifact.uid == folder.uid
```

The same lookup runs in the current database and, if needed, in the storage’s managing database.

Addresses:

-  https://github.com/pfizer-collab/laminlabs/issues/1157

### DIFF
--- a/lamindb/core/loaders.py
+++ b/lamindb/core/loaders.py
@@ -29,6 +29,7 @@ from lamindb_setup.core.upath import (
 
 if TYPE_CHECKING:
     from anndata import AnnData
+    from lamindb_setup.core.upath import UPath  # noqa
     from lamindb_setup.types import AnyPathStr
     from mudata import MuData
     from pandas import DataFrame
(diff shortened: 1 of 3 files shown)


falexwolf:
@Koncopd, this looks interesting but a bit different from what I imagined. I imagined instrumenting `.cache()` on `UPath()` such that it would register the parent folder as an input:

- https://github.com/pfizer-collab/laminlabs/issues/1151

But I think what you're doing addresses the topic at a lower level so that nf-lamin can also handle it.

Either way it'd be great to have more of a PR description and documentation so that a user (& I) can understand what happens. ☺️

Koncopd:
I have added the description.

PR #3859, opened 2026-08-18
🐛 Fix slow `BasicQuerySet.first()` implementation

Koncopd:
It evaluated the whole query set before returning with `limit 1`, this can be very slow on large databases.


falexwolf:
Nice!

PR #3861, opened 2026-08-20
Minimal fix to find objects in merge commits

ishitamjain:
Add `-m` (--diff-merges=on) to the git log invocation in `get_git_commit_hash`. This makes git diff merge commits against each parent individually instead of using the combined diff, so blobs introduced by clean merges are correctly detected by `--find-object`.

Fixing: pfizer-collab/laminlabs#1152

### DIFF
--- a/lamindb/core/_sync_git.py
+++ b/lamindb/core/_sync_git.py
@@ -64,11 +64,17 @@ def get_git_commit_hash(blob_hash: str, repo_dir: Path | None = None) -> str | N
     fetch_command = ["git", "fetch", "origin", "+refs/heads/*:refs/remotes/origin/*"]
     subprocess.run(fetch_command, cwd=repo_dir, check=True)
 
-    # Find the commit containing the blob hash in all branches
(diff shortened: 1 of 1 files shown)


Koncopd:
Looks good. 
Is a test possible for this? Or could you maybe verify manually if this indeed solves the problem?

ishitajain9717:
I tested it locally and the fix seems to works. A  pytest test can be written that either builds a temp repo or clones a known public repo (in this case after cloning we would also need to find a merge commit case) to reproduce the scenario and calls `get_git_commit_hash` directly. The tests works locally.

Koncopd (approved):

Koncopd:
ok, building a repo with certain commits sounds complicated for CI runs.

Koncopd:
@ishitajain9717 could you get back to the issue please and write it is resolved and will land in a new release.

ishitajain9717:
Yes the tests are complicated to test this kind of behaviour (when building repo from scratch) . Comparatively, testing the behaviour with a public repository is easier.

PR #3864, opened 2026-08-20
🚸 `lamin init` creates a database with `dev-dir` in the current directory, analogous to `git`; if not passing `--storage`, the default storage location is `./storage/`

falexwolf:
When you run `lamin init`, LaminDB will now auto-connect in the parent directory and its children, in the same way in which git recognizes a repo in a directory hierarchy.

For example:

```bash
mkdir mydata && cd mydata
lamin init
```

This will create a LaminDB instance in the `mydata` directory, with `mydata` acting as the development directory  (`dev-dir`) that holds light-weight scripts, notebooks, and notes that are modeled as transforms and records.

If you do not pass the `--storage` flag (as in the example), the storage location will be the: `./storage/`.

See:

- https://github.com/laminlabs/lamindb-setup/pull/1383
- https://github.com/laminlabs/lamin-cli/pull/262


Koncopd:
this https://github.com/laminlabs/lamindb/pull/3864/commits/0cddbf8441d0f26dea7dd369edbf7d5b8e9466dc bothers me a bit. Why did it fix the failing test?

falexwolf:
> this https://github.com/laminlabs/lamindb/commit/0cddbf8441d0f26dea7dd369edbf7d5b8e9466dc bothers me a bit. Why did it fix the failing test?

I know. Sorry. I should have made another attempt. I'll clean this up.

It's because we're asserting this in `acid.md`:

```
Cell In[5], line 4
      1 with pytest.raises(PermissionError) as error:
      2     artifact.save()
      3 print(error.exconly())
----> 4 assert len(ln.Artifact.filter()) == 0
```

We should rather `assert not ln.Artifact.filter(key="acid.fasta").exists()`. This would be the cleaner way anyway. I didn't see it this clearly at first.

The main "problem" that this PR creates is that for the central guides, we don't create a separate test database anymore. This means that for every guide one has to now pay a bit attention that one works with meaningful data; the `acid.md` accidentally had written the same file as the `track.md` guide and this made it look like a problem.

Since things will only evolve very slowly given the overall structure has long converged, I'd say this price is OK. The benefit is that every guide is prettier / less daunting due to the simpler `init` statement.

falexwolf:
I'm fixing this now in another PR... sorry again.

Koncopd:
ok, clear, thank you.

falexwolf:
Fixed:

- https://github.com/laminlabs/lamindb/pull/3865

PR #3889, opened 2026-09-07
:children_crossing: Allow scoping `schema.itype` to a `Feature` type

ishitamjain:
This PR is to enabling scoping of features in the schema
Fixes: https://github.com/pfizer-collab/laminlabs/issues/1176

### DIFF
--- a/lamindb/curators/core.py
+++ b/lamindb/curators/core.py
@@ -676,9 +676,19 @@ class ComponentCurator(Curator):
         feature_ids: set[int] = set()
 
         if schema.flexible:
-            features += (
-                Feature.connect(using).filter(name__in=self._dataset.keys()).to_list()
-            )
+            _itype = schema.itype
+            if _itype and _itype.startswith("Feature["):
+                # Scoped itype: "Feature[<uid>]" — only fetch features of that type.
+                _type_uid = _itype[len("Feature["):-1]
+                features += (
+                    Feature.connect(using)
+                    .filter(name__in=self._dataset.keys(), type__uid=_type_uid)
+                    .to_list()
+                )
+            else:
+                features += (
+                    Feature.connect(using).filter(name__in=self._dataset.keys()).to_list()
+                )
             feature_ids = {feature.id for feature in features}
 
         if schema.n_members and schema.n_members > 0:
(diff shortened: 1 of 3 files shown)


lamindb/curators/core.py
```diff
@@ -676,9 +676,19 @@ def __init__(
         feature_ids: set[int] = set()
 
         if schema.flexible:
-            features += (
-                Feature.connect(using).filter(name__in=self._dataset.keys()).to_list()
-            )
+            _itype = schema.itype
```
falexwolf: This would be a little prettier if you did:

```python
qs = Feature.connect(using).filter(name__in=self._dataset.keys())
itype = schema.itype
if itype and itype.startswith("Feature["):
    type_uid =  itype[8:-1]  # len(Feature[) = 8
    qs = qs.filter(type__uid=type_uid)
features = qs.to_list()
```

falexwolf:
This looks great!

Just a small comment.

Also, do want `schema.itype.startswith("Feature")` (meaning features in the schema are of type `Feature` and not e.g. `bionty.Gene`) or `schema.itype.startswith("Feature[")` (meaning features in the schema are restricted to a feature type?

ishitamjain:
For the transpose guard: startswith("Feature"), because both "Feature" and "Feature[uid]" mean feature names are in the columns, not gene IDs in the row index, so neither should be transposed.

For the flexible lookup: startswith("Feature["), because only the scoped case should narrow the query by type__uid. Generic "Feature" means any feature, no type restriction.

lamindb/curators/core.py
```diff
@@ -711,9 +676,32 @@ def __init__(
...
+                root_uid = itype[8:-1]  # len("Feature[") == 8
+                type_uids: set[str] = {root_uid}
+                queue = [root_uid]
+                while queue:
+                    uid = queue.pop()
+                    for child_uid in (
+                        Feature.connect(using)
+                        .filter(type__uid=uid, is_type=True)
+                        .values_list("uid", flat=True)
+                    ):
```
falexwolf, line 699: This is waaaaay too complicated and connection-intensive. Whenever you make connections inside a Python loop, your code is likely wrong.

Instead, please just use `feature.query_features()`: https://docs.lamin.ai/lamindb.feature#lamindb.Feature.query_features

This should do everything efficiently for you. @Koncopd solved this problem a few years ago with that implementation. ☺️

falexwolf:
Everything looks great now! Test failure is unrelated and either due to updates of cellxgene or @sunnyosun's bionty updates. That failure should be resolved in another PR to not mix concerns.

PR #3915, opened 2026-09-16
:children_crossing: Enable transferring records with their features

ishitamjain:
This PR fixes the transfer of records from one instance, along with the features.


lamindb/models/artifact.py
```diff
@@ -3345,16 +3345,17 @@ def delete(
     def save(
         self,
         upload: bool | None = None,
-        transfer: Literal["record", "annotations"] = "record",
+        transfer: Literal["sqlrecord", "notes", "annotations", "record"] = "sqlrecord",
```
falexwolf: Please remove the "record" option from the type annotation, since it is now deprecated and should no longer be part of the user-facing API docs.

lamindb/models/artifact.py
```diff
@@ -3345,16 +3345,17 @@ def delete(
...
         **kwargs,
     ) -> Artifact:
         """Save to database & storage.
 
         Args:
             upload: Trigger upload to cloud storage in instances with hybrid storage mode.
             transfer: In case artifact was queried on a different instance, dictates behavior of sync.
-                If "record", only the artifact record is synced to the current instance.
-                If "annotations", also the annotations linked in the source instance are synced.
+                ``sqlrecord`` (default; alias ``record`` until v3): only the SQL row and its foreign keys.
```
falexwolf: Please use quotes " and not double backticks (we never use double backticks, but single backticks, here though it's a Literal so it should be in quotes).

The way you're writing the docstring also likely doesn't render nicely on the docs. Please take a screenshot to make sure it's pretty and add that screenshot to the PR description.

lamindb/models/artifact.py
```diff
@@ -3537,12 +3538,9 @@ def save(
             if not _handle_suffix_change_on_save(self):
                 return None
 
-        if transfer not in {"record", "annotations"}:
-            raise ValueError(
-                f"transfer should be either 'record' or 'annotations', not {transfer}"
-            )
-        else:
-            kwargs["transfer"] = transfer
+        from .sqlrecord import normalize_transfer_config
```
falexwolf: That import can be top-level at the file. There is no danger of a cycle or cold start.

lamindb/models/artifact.py
```diff
@@ -3537,12 +3538,9 @@ def save(
...
+
+        kwargs["transfer"] = normalize_transfer_config(transfer)
```
falexwolf, line 3542: Nice!

lamindb/models/record.py
```diff
@@ -1230,6 +1230,14 @@ def __init__(
         super().__init__(**super_kwargs)
 
     def save(self, *args, **kwargs) -> Record:
+        """Save.
+
+        Args:
+            transfer: If this record was queried on another instance:
+                ``sqlrecord`` (default; alias ``record`` until v3) copies the row
```
falexwolf: Please use quotes here.

lamindb/models/sqlrecord.py
```diff
@@ -1389,6 +1389,11 @@ def save(self: T, *args, **kwargs) -> T:
 
         Args:
             using: Optional database slug for a target database that differs from the default database.
+            transfer: If this object was queried on another instance:
+                ``sqlrecord`` (default; alias ``record`` until v3) copies the row
```
falexwolf: Please quotes.

falexwolf:
Thanks, Ishita! This looks really nice already. A lot of good choices.

I made some small comments re docs and typing cosmetics.

The big problem is that you have 38 uncovered lines in this:

image

Please click on the `sqlrecord.py` link and then add tests for every line that's not covered. Ideally you don't add 100 tests but you parametrize and broaden the unit test you added.

Please don't use Codex 5.3 anymore, but a better model. Either modern Grok with Cursor (4.7 or 4.8) or powerful Codex ones.

docs/transfer.md
```diff
@@ -83,9 +131,32 @@ Every `SQLRecord` object has an attribute `._state.db` which can take the follow
...
-# test the last 3 cells here
 assert artifact.transform.description == "Transfer from `laminlabs/lamindata`"
 assert artifact.transform.key == "__lamindb_transfer__/4XIuR0tvaiXM"
 assert artifact.transform.uid == "4XIuR0tvaiXM0000"
 assert artifact.run.initiated_by_run.transform.description.startswith("Transfer & sync")
+assert artifact.features.slots
+for schema in artifact.features.slots.values():
+    _ = schema.index
+
+source = db.Record.get("mNDJgWFrkWQVW3ox")
```
falexwolf, line 142: @ishitamjain, can you manually verify that these asserts capture everything that's there in the source? You could also run this on a manual SQLite to understand how related entities are generated.

falexwolf:
@ishitamjain, there are also still a few uncovered lines.

image

For example, here in `sqlrecord.py`:

image

Please either test them or remove them: https://laminlabs.slack.com/archives/C07DB677JF6/p1789673524112759?thread_ts=1789630082.022259&cid=C07DB677JF6

docs/transfer.md
ishitamjain: The asserts are comprehensive enough for that record, the key set has to match the source, the listed cells are compared, and owner remaps. I’ll also run it once on a local SQLite to see how the related entities land

falexwolf:
For example, this here is a case, @ishitamjain, that doesn't make sense:

image

If the parsing doesn't succeed we need to hard error.

falexwolf:
Great, @ishitamjain! 

This can be removed:

image

All models have `ablocks`.

These ones below still deserve attention. You can easily construct such a case by e.g. adding bionty to one of the databases in the `transfers/` group. It will probably reveal that we should catch a more specific than just `Exception`.

image

falexwolf:
Almost, almost there!

This exception is still not raised in the tests. Or, the test is not included into our coverage report:

image

tests/transfer/test_transfer.py
```diff
@@ -203,6 +205,10 @@ def _source_on_testdb1() -> tuple[str, str]:
     ln.models.RecordBlock(record=source, content=README, kind="readme").save()
     empty = ln.Record(name=EMPTY_NAME, type=sheet).save()
     assert source.features.get_values()[FEAT_VERSION] == "2.10.0"
+    # Stored as str so get_values() stays valid; dtype is flipped so transfer
+    # hits parse_dtype ValidationError on an instance without bionty.
+    skip = ln.Feature(name=FEAT_SKIP, dtype=str).save()
```
falexwolf: Why is this testing the `raise ValueError`? This seems to be testing skipping a feature transfer. But the code shows that an exception would be raised: https://github.com/laminlabs/lamindb/pull/3915#issuecomment-5774333079

ishitamjain: This is the missing-module case, not a skip. The old name was leftover. Transferring cat[bionty.Organism] onto testdb2 (no bionty) must raise ValueError. Correcting it

falexwolf:
I'm merging this PR in the current state now. I will follow-up with refactoring and performance optimization.

What this PR achieves is correctness and more sensible behavior during transfer (e.g. maintaining user identity and space identity).

PR #3920, opened 2026-09-17
✨ Validate zarr format constraints for `SpatialData` schemas

namsaraeva:
Adds validation of storage conventions for `SpatialData`, so that a schema constrains not just the *content* of a dataset but also how it is stored.

This is how the format constraints are declared:

```python
schema.formats.zarr = {
    "zarr_format": 3,
    "chunk_shape": {"y": [256, 512], "x": [256, 512]},
    "multiscales": {"scale": 2, "axes": ["y", "x"], "coarsest_single_chunk": True},
    "layers": {"csc": "csc_matrix"},
}
```


namsaraeva:
@falexwolf this is the first prototype of the additional checks. What do you think of the design? 

I will also add a test later, when you are happy with how things look.

falexwolf:
Thank you, Altana! I think this looks great! The only thing that misses are a few tests.

I'd add a public API on schema that defines how one adds these additional zarr constraints:

```
schema.formats.zarr = zarr_constraints_dict
```

Allowed formats for this schema are managed through a `FormatConstraints` class and essentially write a JSON under a given key. I could do that tomorrow if this sounds good! It's very simple but would make it easy to document and discover in the API. It would also publicly bake the `zarr` key.

namsaraeva:
@falexwolf this sounds good! Thank you!

falexwolf:
Great! I'll add the public API on `Schema`. 

Can you add a few tests that cover different zarr configurations? Ideally in a new file? You could run them in the `/curators/` test suite.

namsaraeva:
@falexwolf added some tests 😃 the cli test is failing right now though.

falexwolf:
I'll fix the CLI test, no worries! Your tests are passing.

I'll take it from here and this should land in the 2.11 release that's scheduled for very very soon.

namsaraeva:
Thank you, @falexwolf!

